PR #342, opened 2021-09-17
chore: fix contentmd5middleware to render properly

wooj2:
ContentMD5Middleware was not properly conforming to `MiddlewareRenderable` (well, it was but.. since there was a default impl  of `{}`, it was possible to call render on this middleware and render nothing) -- therefore in real code generated client, the ContentMD5Middleware was never getting properly rendered

This is a temporary fix before we revamp the interface.


kneekey23 (approved):
LGTM

PR #349, opened 2021-09-20
fix: encoding of lists of maps of lists which fixes ssm service

kneekey23:
## Issue \#

It fixes 2/3 of https://github.com/awslabs/aws-sdk-swift/issues/391
## Description of changes

This fix accounts for correct encoding of lists of maps of lists I.e. `public let targetMaps: [[String : [String]]]?`.

## Scope


smithy-swift-codegen/src/test/kotlin/mocks/MockHttpAWSJson11ProtocolGenerator.kt
```diff
@@ -0,0 +1,109 @@
...
+        decodeGenerator.render()
+    }
+
+    override fun addProtocolSpecificMiddleware(ctx: ProtocolGenerator.GenerationContext, operation: OperationShape) {
+        // Intentionally empty
+    }
+
+    override fun getProtocolHttpBindingResolver(ctx: ProtocolGenerator.GenerationContext):
+        HttpBindingResolver = MockJsonHttpBindingResolver(ctx)
+}
```
wooj2, line 109: You may need to rebase -- this function i believe needs two parameters.

kneekey23: ive rebased like 3x and for some reason still not getting this change but I see what you mean

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -164,8 +164,8 @@ abstract class MemberShapeEncodeGenerator(
         when (targetShape) {
             is CollectionShape -> {
                 val topLevelContainerName = "${memberName}Container"
-                writer.write("var \$L = $containerName.nestedContainer(keyedBy: \$N.self)", topLevelContainerName, ClientRuntimeTypes.Serde.Key)
-                renderEncodeMap(ctx, memberName, topLevelContainerName, targetShape, level)
+                writer.write("var \$L = $containerName.nestedUnkeyedContainer(forKey: \$N($dictKey${level - 1}))", topLevelContainerName, ClientRuntimeTypes.Serde.Key)
```
wooj2, line 167: is it possible for level to be zero?  if so, then we would generate `Key(dictKey-1)` ?

kneekey23: not for it to go here in the when statement. level would have to be >= 1 if the target shape were a collection shape

kneekey23: but I did think of that and check that that wasn't gonna happen :)

wooj2: I think it would make the code more readable if we removed the default value of ` = 0` for this parameter -- and then in StructEncodeGenerator, pass in "0" so that i the call to this method looks like:
```
                                renderEncodeMapMember(target, memberName, containerName, 0)
```

wooj2 (approved):
lgtm!

PR #350, opened 2021-09-20
chore: simplify MiddlewareRenderable

wooj2:
Simplifying protocol before adding complexity

corresponding pr:
https://github.com/awslabs/aws-sdk-swift/pull/401

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/LoggingMiddleware.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/LoggingMiddleware.kt
@@ -29,14 +29,10 @@ class LoggingMiddleware : MiddlewareRenderable {
         op: OperationShape,
         operationStackName: String
     ) {
-        writer.write("$operationStackName.${middlewareStep.stringValue()}.intercept(position: ${position.stringValue()}, middleware: \$N(${middlewareParamsString(model, symbolProvider, op)}))", ClientRuntimeTypes.Middleware.LoggerMiddleware)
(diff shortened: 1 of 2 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/LoggingMiddleware.kt
```diff
@@ -29,14 +29,10 @@ class LoggingMiddleware : MiddlewareRenderable {
...
-        writer.write("$operationStackName.${middlewareStep.stringValue()}.intercept(position: ${position.stringValue()}, middleware: \$N(${middlewareParamsString(model, symbolProvider, op)}))", ClientRuntimeTypes.Middleware.LoggerMiddleware)
+        writer.write("$operationStackName.${middlewareStep.stringValue()}.intercept(position: ${position.stringValue()}, middleware: \$N(${middlewareParamsString()}))", ClientRuntimeTypes.Middleware.LoggerMiddleware)
     }
 
-    override fun middlewareParamsString(
-        model: Model,
-        symbolProvider: SymbolProvider,
-        op: OperationShape
-    ): String {
+    private fun middlewareParamsString(): String {
```
kneekey23, line 35: if we are removing this, might as well jsut remove the whole need for another function to just return a string. can you just inline this into line 32?

kneekey23 (approved):
LGTM

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/LoggingMiddleware.kt
wooj2: spoke offline. keeping this as-is

PR #352, opened 2021-09-21
feat: Expose sigv4 config for presigner

wooj2:
This is one proposed way of allowing customers to override the sigv4 config for presigners.

Corresponding:
https://github.com/awslabs/aws-sdk-swift/pull/402

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/middleware/MiddlewareRenderable.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/middleware/MiddlewareRenderable.kt
@@ -10,6 +10,11 @@ import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.swift.codegen.SwiftWriter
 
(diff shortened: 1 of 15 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/middleware/MiddlewareRenderable.kt
```diff
@@ -22,5 +27,5 @@ interface MiddlewareRenderable {
 
     val position: MiddlewarePosition
 
-    fun render(model: Model, symbolProvider: SymbolProvider, writer: SwiftWriter, op: OperationShape, operationStackName: String)
+    fun render(model: Model, symbolProvider: SymbolProvider, writer: SwiftWriter, op: OperationShape, operationStackName: String, executionContext: MiddlewareRenderableExecutionContext)
```
kneekey23, line 30: does it make sense to default this to CLIENT since in most situations its CLIENT? then you can avoid passing it everywhere?

kneekey23 (approved):
I actually like this idea, ship it! just had one question tho

PR #355, opened 2021-09-24
feat: make struct members var

kneekey23:
## Description of changes

This adds an extension used by the glacier customization to check if a string is nil or empty in swift.
Corresponding PR: https://github.com/awslabs/aws-sdk-swift/pull/408

## Scope

- [ x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
+++ b/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
@@ -113,10 +113,3 @@ extension String {
         return ByteBuffer(data: self.data(using: .utf8) ?? Data())
     }
 }
-
-public extension Optional where Wrapped: Collection {
-  var isNilOrEmpty: Bool {
-    guard let value = self else { return true }
-    return value.isEmpty
-  }
-}
--- a/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
+++ b/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
@@ -113,3 +113,10 @@ extension String {
         return ByteBuffer(data: self.data(using: .utf8) ?? Data())
     }
 }
+
+public extension Optional where Wrapped: Collection {
+  var isEmptyOrNil: Bool {
+    guard let value = self else { return true }
+    return value.isEmpty
+  }
+}
--- a/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
+++ b/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
@@ -115,7 +115,7 @@ extension String {
 }
 
 public extension Optional where Wrapped: Collection {
-  var isEmptyOrNil: Bool {
+  var isNilOrEmpty: Bool {
     guard let value = self else { return true }
     return value.isEmpty
   }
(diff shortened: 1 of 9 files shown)


Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
```diff
@@ -113,3 +113,10 @@ extension String {
         return ByteBuffer(data: self.data(using: .utf8) ?? Data())
     }
 }
+
+public extension Optional where Wrapped: Collection {
+  var isNilOrEmpty: Bool {
+    guard let value = self else { return true }
+    return value.isEmpty
+  }
+}
```
wooj2, line 122: 1.  Does this need to be public?  I guess it does since a package like AWSGlacier wants to be able to use it?
2. It looks like this extension is applied to all optional collections?  If this is true, is this what we want?  If this is what we want, seems like this should be in a separate file (as it is currently in a file called `String+Extensions`
3. I don't understand how `where Wrapped` works -- can you explain?

Packages/ClientRuntime/Tests/ClientRuntimeTests/PrimitiveTypeExtensionsTests/StringExtensionsTests.swift
```diff
@@ -96,4 +96,14 @@ class StringExtensionsTests: XCTestCase {
...
+    
+    func testNilOrEmptyExtension() {
+        let emptyString: String? = ""
+        let nilString: String? = nil
+        let stringWithValue: String? = "something"
+        
+        XCTAssertFalse(stringWithValue.isNilOrEmpty)
+        XCTAssertTrue(nilString.isNilOrEmpty)
+        XCTAssertTrue(emptyString.isNilOrEmpty)
+    }
```
wooj2, line 108: It would be better to split this out in to 3 different functions to reduce the scope of each case.

Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
kneekey23: removing per changes in other PR

Packages/ClientRuntime/Tests/ClientRuntimeTests/PrimitiveTypeExtensionsTests/StringExtensionsTests.swift
kneekey23: removing as no longer needed

smithy-swift-codegen/src/test/kotlin/StructEncodeGenerationIsolatedTests.kt
```diff
@@ -7,7 +7,7 @@ import io.kotest.matchers.string.shouldContainOnlyOnce
 import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.Test
 
-class StructEncodeGenerationIsolatedTests {
+class varStructEncodeGenerationIsolatedTests {
```
wooj2, line 10: Class name should probably start with a capital letter

wooj2 (approved):
assuming we change the class name to have a capital letter, lgtm!

smithy-swift-codegen/src/test/kotlin/StructEncodeGenerationIsolatedTests.kt
kneekey23: ha corrected

PR #361, opened 2021-10-06
fix: runtime bug fix

kneekey23:
## Issue \#

This closes https://github.com/awslabs/aws-sdk-swift/issues/423

## Scope

### DIFF
--- a/smithy-swift-codegen/src/test/kotlin/HttpUrlPathMiddlewareTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/HttpUrlPathMiddlewareTests.kt
@@ -28,9 +28,8 @@ class HttpUrlPathMiddlewareTests {
                     guard let label1 = input.label1 else {
                         return .failure(.client(ClientRuntime.ClientError.pathCreationFailed(("label1 is nil and needs a value for the path of this operation"))))
                     }
-                    let hostCustomPath = URL(string: "http://\(context.getHost())")?.path
                     var urlPath = "/smoketest/\(label1.urlPercentEncoding())/foo"
-                    if let hostCustomPath = hostCustomPath, !hostCustomPath.isEmpty {
+                    if let host = context.getHost(), let hostCustomPath = URL(string: "http://\(host)")?.path, !hostCustomPath.isEmpty {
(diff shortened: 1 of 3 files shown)


smithy-swift-codegen/src/test/kotlin/HttpUrlPathMiddlewareTests.kt
```diff
@@ -28,9 +28,8 @@ class HttpUrlPathMiddlewareTests {
                     guard let label1 = input.label1 else {
                         return .failure(.client(ClientRuntime.ClientError.pathCreationFailed(("label1 is nil and needs a value for the path of this operation"))))
                     }
-                    let hostCustomPath = URL(string: "http://\(context.getHost())")?.path
                     var urlPath = "/smoketest/\(label1.urlPercentEncoding())/foo"
-                    if let hostCustomPath = hostCustomPath, !hostCustomPath.isEmpty {
+                    if let host = context.getHost(), let hostCustomPath = URL(string: "http://\(host)")?.path, !hostCustomPath.isEmpty {
```
wooj2, line 32: 1.  This doesn't seem to be something you introduced in this PR, but why are we defaulting to http?
2.  This seems awkward - because if `context.getHost()` is not nil, then urlPath can  have the protocol and host, but if `context.getHost()` is nil, then we would be returning `/smoketest/label1.percentencoding/foo` -- seems like `urlPath` should consistently contain similar url components, no?

wooj2: Ah, thanks for the reminder -- I was confused because i would expect getHost() would return only the host name (not the host and the path), therefore, if we execute the code:
```
if let host = context.getHost(), let hostCustomPath = URL(string: \"http://\\(host)\")?.path
```
I would expect that hostCustomPath would only ever be `/`.

But, I think I'm starting to remember, there was one protocol codegen test which set host to `host.com/something`, correct?  And the behavior was suppose to prepend the url path?  Oof, i think from a coding perspective, merging this in to unblock people not setting host is fine, but it makes me think we should update codegen to allow users to add another field called 'prepend path' rather than overloading the host field.

PR #363, opened 2021-10-11
fix: Expose Client Log Mode

kneekey23:
## Description of changes

Closes: https://github.com/awslabs/aws-sdk-swift/issues/425
Corresponding PR: https://github.com/awslabs/aws-sdk-swift/pull/438

## Scope

### DIFF
--- a/smithy-swift-codegen/src/test/kotlin/HttpProtocolClientGeneratorTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/HttpProtocolClientGeneratorTests.kt
@@ -16,6 +16,7 @@ class HttpProtocolClientGeneratorTests {
         contents.shouldContainOnlyOnce(
             """
             public class RestJsonProtocolClient {
+                public static let clientName = "RestJsonProtocolClient"
                 let client: ClientRuntime.SdkHttpClient
                 let config: ClientRuntime.SDKRuntimeConfiguration
                 let serviceName = "Rest Json Protocol"


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
```diff
@@ -88,8 +88,10 @@ object ClientRuntimeTypes {
     }
 }
 
-private fun runtimeSymbol(name: String): Symbol = buildSymbol {
+private fun runtimeSymbol(name: String, defaultValue: String? = null, nullable: Boolean = true): Symbol = buildSymbol {
     this.name = name
     this.namespace = SwiftDependency.CLIENT_RUNTIME.target
+    this.defaultValue = defaultValue
+    this.nullable = nullable
```
wooj2, line 95: 1.  Where is this new field `nullable` being used?
2. Probably will get cleared up once we answer #1, but, can't everything be nullable?

kneekey23: 1. https://github.com/awslabs/smithy-swift/blob/main/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/SymbolBuilder.kt#L72
2. technically yes but per spec no. actually we might not need this. I might have messed up. checking one thing.

wooj2 (approved):
lgtm!

PR #368, opened 2021-10-15
chore: add support for resolvedHost for request tests

wooj2:

### DIFF
--- a/Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
+++ b/Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
@@ -32,9 +32,13 @@ open class HttpRequestTestBase: XCTestCase {
                                          forbiddenQueryParams: [String]? = nil,
                                          requiredQueryParams: [String]? = nil,
                                          body: String?,
-                                         host: String) -> ExpectedSdkHttpRequest {
+                                         host: String,
+                                         resolvedHost: String?) -> ExpectedSdkHttpRequest {
         let builder = ExpectedSdkHttpRequestBuilder()
-        
+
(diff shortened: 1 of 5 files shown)


Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
```diff
@@ -87,6 +91,19 @@ open class HttpRequestTestBase: XCTestCase {
...
+        var deconflictedHost: String? = nil
+        if !host.isEmpty,
+           let urlFromHost = ClientRuntime.URL(string: "http://\(host)"),
+           let parsedHost = urlFromHost.host {
+            deconflictedHost = parsedHost
+        }
+        if let resolvedHost = resolvedHost {
+            deconflictedHost = resolvedHost
+        }
+        return deconflictedHost
```
kneekey23, line 105: do you know at this point that we are definitely going to return a String? like does it need to be optional here, your return value?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -30,14 +30,16 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
     }
 
     private fun renderExpectedBlock(test: HttpRequestTestCase) {
+        var resolvedHostValue = test.resolvedHost?.let { it } ?: run { "nil" }
```
kneekey23, line 33: nice!

kneekey23 (approved):
one question, non-blocking. LGTM. Looks like CI is failing tho also

PR #369, opened 2021-10-18
chore: use host to get prefix and inject into middleware

wooj2:
Decoupling "host" parameter from request test specification ( https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html#smithy-test-httprequesttests-trait )  from production middleware.

### DIFF
--- a/Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
+++ b/Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
@@ -105,9 +105,9 @@ open class HttpRequestTestBase: XCTestCase {
         return deconflictedHost
     }
 
-    public func urlPrefixFromHost(host: String) -> String {
+    public func urlPrefixFromHost(host: String) -> String? {
         guard !host.isEmpty, let hostCustomPath = URL(string: "http://\(host)")?.path else {
-            return ""
+            return nil
         }
         return hostCustomPath
     }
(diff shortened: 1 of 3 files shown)


Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
```diff
@@ -104,6 +104,13 @@ open class HttpRequestTestBase: XCTestCase {
         }
         return deconflictedHost
     }
+
+    public func urlPrefixFromHost(host: String) -> String {
+        guard !host.isEmpty, let hostCustomPath = URL(string: "http://\(host)")?.path else {
+            return ""
```
kneekey23, line 110: why not just return nil here and check for nil in the middleware?

smithy-swift-codegen/src/test/kotlin/HttpProtocolUnitTestRequestGeneratorTests.kt
```diff
@@ -81,12 +81,11 @@ class HttpProtocolUnitTestRequestGeneratorTests {
         encoder.nonConformingFloatEncodingStrategy = .convertToString(positiveInfinity: "Infinity", negativeInfinity: "-Infinity", nan: "NaN")
         let context = HttpContextBuilder()
                       .withEncoder(value: encoder)
-                      .withHost(value: host)
                       .build()
         var operationStack = OperationStack<SmokeTestInput, SmokeTestOutputResponse, SmokeTestOutputError>(id: "SmokeTest")
-        operationStack.initializeStep.intercept(position: .after, middleware: SmokeTestInputURLPathMiddleware())
+        operationStack.initializeStep.intercept(position: .after, middleware: SmokeTestInputURLPathMiddleware(urlPrefix: urlPrefix))
```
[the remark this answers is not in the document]
wooj2: Yep, it defaults to empty.

Packages/SmithyTestUtil/Sources/RequestTestUtil/HttpRequestTestBase.swift
wooj2: completed!

PR #375, opened 2021-10-22
chore: unify host and host prefix middleware

wooj2:
Introduced URLHostMiddleware which is responsible for setting up the context's host and host prefix.

HostPrefix can be generated directly from reading in the spec, while host can be injected in to the constructor because of how request tests work.

corresponding:
https://github.com/awslabs/aws-sdk-swift/pull/458

### DIFF
--- a/smithy-swift-codegen/src/test/kotlin/HttpProtocolClientGeneratorTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/HttpProtocolClientGeneratorTests.kt
@@ -111,7 +111,7 @@ class HttpProtocolClientGeneratorTests {
     @Test
     fun `it renders host prefix middleware with label correctly`() {
         val context = setupTests("host-prefix-operation.smithy", "com.test#Example")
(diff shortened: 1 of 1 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/RequestTestEndpointResolverMiddleware.kt
```diff
@@ -26,6 +26,8 @@ class RequestTestEndpointResolverMiddleware(private val model: Model, private va
             ClientRuntimeTypes.Core.SdkError
         ) {
             writer.write("input.withPath(context.getPath())")
+            writer.write("let host = \"\\(context.getHostPrefix() ?? \"\")\\(context.getHost() ?? \"\")\"")
+            writer.write("input.withHost(host)")
```
wooj2, line 30: To keep things simple, i've added these two lines -- in the future, we can work on unifying the endpoint resolver to the request tests

PR #382, opened 2021-10-29
fix: tcp connection timeout

kneekey23:
## Description of changes

This fixes the tcp timeout for iOS/watchOS per the new SEP requirements. Corresponding PR: https://github.com/awslabs/aws-crt-swift/pull/56

## Scope

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
@@ -37,7 +37,10 @@ public class CRTClientEngine: HttpClientEngine {
             logger.error("Server name was not able to be set in TLS Connection Options. TLS Negotiation will fail.")
             logger.error("Error: \(err.localizedDescription)")
         }
-        let socketOptions = SocketOptions(socketType: .stream)
+        var socketOptions = SocketOptions(socketType: .stream)
+#if os(iOS) || os(watchOS)
+        socketOptions.connectTimeoutMs = 30_000
+#endif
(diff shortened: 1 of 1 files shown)


Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -37,7 +37,10 @@ public class CRTClientEngine: HttpClientEngine {
             logger.error("Server name was not able to be set in TLS Connection Options. TLS Negotiation will fail.")
             logger.error("Error: \(err.localizedDescription)")
         }
-        let socketOptions = SocketOptions(socketType: .stream)
+        var socketOptions = SocketOptions(socketType: .stream)
+#if os(iOS) || os(watchOS)
+        socketOptions.connectTimeoutMs = 30_000
+#endif
```
wooj2, line 43: I'm not quite sure if i understand the need to do this if we are doing it in the crt:
https://github.com/awslabs/aws-crt-swift/pull/56/files

wooj2: chatted offline-- other than changing the commit title to something like "tcp connection timeout", lgtm!

wooj2 (approved):
lgtm!

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
kneekey23: answered offline but basically this is a different timeout. this timeout is tcp and the other is tls negotiation timeout.

PR #386, opened 2021-11-04
chore: fix compiler warnings in clientruntime

wooj2:

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
@@ -37,7 +37,7 @@ public class CRTClientEngine: HttpClientEngine {
             logger.error("Server name was not able to be set in TLS Connection Options. TLS Negotiation will fail.")
             logger.error("Error: \(err.localizedDescription)")
         }
-        var socketOptions = SocketOptions(socketType: .stream)
+        let socketOptions = SocketOptions(socketType: .stream)
 #if os(iOS) || os(watchOS)
         socketOptions.connectTimeoutMs = 30_000
 #endif
--- a/Packages/ClientRuntime/Sources/Networking/Http/Middlewares/ContentMD5Middleware.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/Middlewares/ContentMD5Middleware.swift
@@ -28,7 +28,7 @@ public struct ContentMD5Middleware<OperationStackOutput: HttpResponseBinding,
                 return next.handle(context: context, input: input)
             }
             input.headers.update(name: "Content-MD5", value: base64Encoded)
-        case .stream(let stream):
+        case .stream:
             guard let logger = context.getLogger() else {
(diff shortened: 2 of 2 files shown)


Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -37,7 +37,7 @@ public class CRTClientEngine: HttpClientEngine {
             logger.error("Server name was not able to be set in TLS Connection Options. TLS Negotiation will fail.")
             logger.error("Error: \(err.localizedDescription)")
         }
-        var socketOptions = SocketOptions(socketType: .stream)
+        let socketOptions = SocketOptions(socketType: .stream)
```
kneekey23, line 40: You can still adjust connect timeout on iOS even if this is let?

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -37,7 +37,7 @@ public class CRTClientEngine: HttpClientEngine {
...
 #if os(iOS) || os(watchOS)
```
kneekey23, line 41: Can you add tvOS ti this line while you're in here please?

kneekey23 (approved):
2 questions but LGTM

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
wooj2: Seems that way :)

wooj2: Let me triple check that we can do that on iOS

wooj2:
```
        let socketOptions = SocketOptions(socketType: .stream)
#if os(iOS) || os(watchOS)
        socketOptions.connectTimeoutMs = 30_000
#endif
```

Seems to compile with iOS, so i think we are good.  Here is how i tested it.
1.  Compiled the following for mac:
```
        let socketOptions = SocketOptions(socketType: .stream)
#if os(iOS) || os(watchOS)
alkdjfklasjflksjdf
        socketOptions.connectTimeoutMs = 30_000
#endif
```
Result: build successful

1.  Compiled the following for iOS:
```
        let socketOptions = SocketOptions(socketType: .stream)
#if os(iOS) || os(watchOS)
alkdjfklasjflksjdf
        socketOptions.connectTimeoutMs = 30_000
#endif
```
result: build failure

1.  Compiled the following for iOS (removed junk):
```
        let socketOptions = SocketOptions(socketType: .stream)
#if os(iOS) || os(watchOS)
        socketOptions.connectTimeoutMs = 30_000
#endif
```
result: build succeeded

kneekey23: nice, lets merge this then!

PR #389, opened 2021-11-15
feat: refactor everything to be async

kneekey23:
## Description of changes

Refactor entire sdk to be async

Corresponding:
https://github.com/awslabs/aws-sdk-swift/pull/497

## Scope

### DIFF
--- a/Packages/ClientRuntime/Sources/Middleware/MiddlewareStep.swift
+++ b/Packages/ClientRuntime/Sources/Middleware/MiddlewareStep.swift
@@ -4,16 +4,14 @@
 /// An instance of MiddlewareStep will be contained in the operation stack, and recognized as a single
 /// step (initialize, build, etc..) that contains an ordered list of middlewares. This class is
 /// responsible for ordering these middlewares so that they are executed in the correct order.
-public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output, MError: Error>: Middleware {
+public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output>: Middleware {
     public typealias Context = StepContext
     public typealias MInput = Input
     public typealias MOutput = Output
-    public typealias MError = MError
     
     var orderedMiddleware: OrderedGroup<MInput,
                                         MOutput,
-                                        Context,
-                                        MError> = OrderedGroup<MInput, MOutput, Context, MError>()
+                                        Context> = OrderedGroup<MInput, MOutput, Context>()
     
     public let id: String
     
@@ -21,21 +19,21 @@ public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output, MErr
         self.id = id
     }
     
-    func get(id: String) -> AnyMiddleware<MInput, MOutput, Context, MError>? {
+    func get(id: String) -> AnyMiddleware<MInput, MOutput, Context>? {
         return orderedMiddleware.get(id: id)
     }
     
     /// This execute will execute the stack and use your next as the last closure in the chain
     public func handle<H: Handler>(context: Context,
                                    input: MInput,
-                                   next: H) -> Result<MOutput, MError>
-    where H.Input == MInput, H.Output == MOutput, H.Context == Context, H.MiddlewareError == MError {
+                                   next: H) async throws -> MOutput
+    where H.Input == MInput, H.Output == MOutput, H.Context == Context{
         
         var handler = next.eraseToAnyHandler()
         let order = orderedMiddleware.orderedItems
         
         guard !order.isEmpty else {
-            return handler.handle(context: context, input: input)
+            return try await handler.handle(context: context, input: input)
         }
         let numberOfMiddlewares = order.count
         let reversedCollection = (0...(numberOfMiddlewares-1)).reversed()
@@ -44,12 +42,11 @@ public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output, MErr
             handler = composedHandler.eraseToAnyHandler()
         }
         
-        let result = handler.handle(context: context, input: input)
-        return result
+        return try await handler.handle(context: context, input: input)
     }
     
     public mutating func intercept<M: Middleware>(position: Position, middleware: M)
-    where M.MInput == MInput, M.MOutput == MOutput, M.Context == Context, M.MError == MError {
+    where M.MInput == MInput, M.MOutput == MOutput, M.Context == Context {
         orderedMiddleware.add(middleware: middleware.eraseToAnyMiddleware(), position: position)
     }
     
@@ -61,7 +58,7 @@ public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output, MErr
     ///
     public mutating func intercept(position: Position,
                                    id: String,
-                                   middleware: @escaping MiddlewareFunction<MInput, MOutput, Context, MError>) {
+                                   middleware: @escaping MiddlewareFunction<MInput, MOutput, Context>) {
         let middleware = WrappedMiddleware(middleware, id: id)
         orderedMiddleware.add(middleware: middleware.eraseToAnyMiddleware(), position: position)
     }
--- a/Packages/ClientRuntime/Sources/Middleware/MiddlewareStep.swift
+++ b/Packages/ClientRuntime/Sources/Middleware/MiddlewareStep.swift
@@ -27,7 +27,7 @@ public struct MiddlewareStep<StepContext: MiddlewareContext, Input, Output>: Mid
     public func handle<H: Handler>(context: Context,
                                    input: MInput,
                                    next: H) async throws -> MOutput
-    where H.Input == MInput, H.Output == MOutput, H.Context == Context{
+    where H.Input == MInput, H.Output == MOutput, H.Context == Context {
         
         var handler = next.eraseToAnyHandler()
         let order = orderedMiddleware.orderedItems
--- a/Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
+++ b/Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
@@ -5,11 +5,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 	
-public struct NoopHandler<Output: HttpResponseBinding, OutputError: HttpResponseBinding>: Handler {
-    public init() {
-    }
-    public func handle(context: HttpContext, input: SdkHttpRequest) -> Result<OperationOutput<Output>, SdkError<OutputError>> {
+public struct NoopHandler<Output: HttpResponseBinding>: Handler {
+    public init() {}
+    
+    public func handle(context: HttpContext, input: SdkHttpRequest) async throws -> OperationOutput<Output> {
         let output = OperationOutput<Output>(httpResponse: HttpResponse())
-        return .success(output)
+        return output
     }
 }
--- a/Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
+++ b/Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
@@ -9,7 +9,6 @@ public struct NoopHandler<Output: HttpResponseBinding>: Handler {
     public init() {}
     
     public func handle(context: HttpContext, input: SdkHttpRequest) async throws -> OperationOutput<Output> {
-        let output = OperationOutput<Output>(httpResponse: HttpResponse())
-        return output
+        return OperationOutput<Output>(httpResponse: HttpResponse())
     }
 }
--- a/Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
+++ b/Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
@@ -32,16 +32,19 @@ public struct OperationStack<OperationStackInput,
     where H.Input == SdkHttpRequest,
           H.Output == OperationOutput<OperationStackOutput>,
           H.Context == HttpContext {
-        
-        let deserialize = compose(next: DeserializeStepHandler(handler: next), with: deserializeStep)
-        let finalize = compose(next: FinalizeStepHandler(handler: deserialize), with: finalizeStep)
-        let build = compose(next: BuildStepHandler(handler: finalize), with: buildStep)
-        let serialize = compose(next: SerializeStepHandler(handler: build), with: serializeStep)
-        let initialize = compose(next: InitializeStepHandler(handler: serialize), with: initializeStep)
-        
-        let result = try await initialize.handle(context: context, input: input)
-        return result.output!
-    }
+              
+              let deserialize = compose(next: DeserializeStepHandler(handler: next), with: deserializeStep)
+              let finalize = compose(next: FinalizeStepHandler(handler: deserialize), with: finalizeStep)
+              let build = compose(next: BuildStepHandler(handler: finalize), with: buildStep)
+              let serialize = compose(next: SerializeStepHandler(handler: build), with: serializeStep)
+              let initialize = compose(next: InitializeStepHandler(handler: serialize), with: initializeStep)
+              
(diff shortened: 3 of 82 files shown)


Packages/ClientRuntime/Sources/Middleware/MiddlewareStep.swift
```diff
@@ -4,38 +4,36 @@
...
         return orderedMiddleware.get(id: id)
     }
     
     /// This execute will execute the stack and use your next as the last closure in the chain
     public func handle<H: Handler>(context: Context,
                                    input: MInput,
-                                   next: H) -> Result<MOutput, MError>
-    where H.Input == MInput, H.Output == MOutput, H.Context == Context, H.MiddlewareError == MError {
+                                   next: H) async throws -> MOutput
+    where H.Input == MInput, H.Output == MOutput, H.Context == Context{
```
wooj2, line 30: again with parenthesis and space

Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
```diff
@@ -7,76 +7,65 @@ public struct OperationStack<OperationStackInput: Encodable & Reflection,
...
         let build = compose(next: BuildStepHandler(handler: finalize), with: buildStep)
         let serialize = compose(next: SerializeStepHandler(handler: build), with: serializeStep)
         let initialize = compose(next: InitializeStepHandler(handler: serialize), with: initializeStep)
         
-        let result = initialize.handle(context: context, input: input)
-        return result.flatMap { operationOutput in
-            return .success(operationOutput.output!)
-        }
+        let result = try await initialize.handle(context: context, input: input)
+        return result.output!
```
wooj2, line 43: Now that we have changed the function signature to include  `throw`, should we do something like:

```
let result = try await initialize.handle(...)
guard let output = result.output else {
   throw ....
}
return output
```
??

Packages/ClientRuntime/Sources/Middleware/PresignerShimHandler.swift
```diff
@@ -24,19 +24,18 @@ struct PresignerShim<OperationStackOutput: HttpResponseBinding,
...
+          Self.MInput == H.Input,
+          Self.MOutput == H.Output {
+              handler(input)
+              let httpResponse = HttpResponse(body: .none, statusCode: .ok)
+              do {
+                  let output: OperationStackOutput? = try OperationStackOutput(httpResponse: httpResponse, decoder: nil)
+                  return .init(httpResponse: httpResponse, output: output)
+              } catch {
+                  throw SdkError<OperationStackError>.unknown(ClientError.unknownError("PresignerShimHandler: This code should not execute"))
+              }
```
wooj2, line 39: Relooking at this code, it seems like we could get rid of `OperationStackError`, and just change tihs code to something like:
```
                  let output: OperationStackOutput? = try! OperationStackOutput(httpResponse: httpResponse, decoder: nil)
                  return .init(httpResponse: httpResponse, output: output)
```
lets just keep this as you have it for now, and we can update this in the future.

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -68,38 +68,16 @@ public class CRTClientEngine: HttpClientEngine {
...
-                            response.statusCode = HttpStatusCode(rawValue: statusCode) ?? HttpStatusCode.notFound
-                            return response
-                        }
-                    }
-                }
-                
-                return future
-            }
-        return httpResponseFuture
+        let connection = try await connectionMgr.acquireConnection()
```
wooj2, line 73: In reading the implementation of the `acquireConnection()` function, it seems that the future will be called with either `fulfill` or `fail` -- therefore, it seems possible for this to fail, but it seems that we are always logging "Connection as acquired to: ...` is this the expected behavior?

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -109,43 +87,43 @@ public class CRTClientEngine: HttpClientEngine {
...
-        
+        let crtRequest = request.toHttpRequest(bufferSize: windowSize)
         let streamReader: StreamReader = DataStreamReader()
         
         let requestOptions = HttpRequestOptions(request: crtRequest) { [self] (stream, _, httpHeaders) in
             logger.debug("headers were received")
-            response.statusCode = HttpStatusCode(rawValue: Int(stream.getResponseStatusCode()))
-                ?? HttpStatusCode.notFound
+            response.statusCode = HttpStatusCode(rawValue: Int(stream.statusCode))
+            ?? HttpStatusCode.notFound
```
wooj2, line 98: Can we put this on the same line?

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -109,43 +87,43 @@ public class CRTClientEngine: HttpClientEngine {
...
             response.headers.addAll(httpHeaders: httpHeaders)
         } onIncomingHeadersBlockDone: { [self] (stream, _) in
             logger.debug("header block is done")
-            response.statusCode = HttpStatusCode(rawValue: Int(stream.getResponseStatusCode()))
-                ?? HttpStatusCode.notFound
+            response.statusCode = HttpStatusCode(rawValue: Int(stream.statusCode))
+            ?? HttpStatusCode.notFound
```
wooj2, line 103: Same line plz.

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -109,43 +87,43 @@ public class CRTClientEngine: HttpClientEngine {
...
-                    future.fail(error)
-                    return
+                    continuation.resume(throwing: error)
                 }
             }
-
+            
             response.body = .stream(.reader(streamReader))
-            future.fulfill(response)
+            self.logger.debug("response came back with success")
```
wooj2, line 121: Putting this log here seems like it would create a ton of noise.  I suspect you meant to delete this?

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -109,43 +87,43 @@ public class CRTClientEngine: HttpClientEngine {
...
+        } onStreamComplete: { [self] (stream, error) in
             logger.debug("stream completed")
             streamReader.hasFinishedWriting = true
             if case let CRTError.crtError(unwrappedError) = error {
                 if unwrappedError.errorCode != 0 {
                     logger.error("Response encountered an error: \(error)")
                     streamReader.onError(error: ClientError.crtError(error))
-                    future.fail(error)
-                    return
+                    continuation.resume(throwing: error)
```
wooj2, line 116: why do we no longer return?

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -109,43 +87,43 @@ public class CRTClientEngine: HttpClientEngine {
...
             }
-
+            
             response.body = .stream(.reader(streamReader))
-            future.fulfill(response)
+            self.logger.debug("response came back with success")
+            if let stream = stream {
+                response.statusCode = HttpStatusCode(rawValue: Int(stream.statusCode)) ?? HttpStatusCode.notFound
+            }
+            continuation.resume(returning: response)
```
wooj2, line 125: it's unclear what the success cases and failure cases are -- seems like there ought to be a better way to write this so that we can clearly see what cases we should bail, and what cases we should succeed.

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/LoggerMiddleware.swift
```diff
@@ -35,18 +34,12 @@ public struct LoggerMiddleware<Output: HttpResponseBinding,
...
-            } else if clientLogMode == .requestAndResponseWithBody || clientLogMode == .responseWithBody {
-                logger.debug("Response: \(output.httpResponse.debugDescriptionWithBody)")
-            }
-            
-        } catch {
-            return response
+        if clientLogMode == .response || clientLogMode == .requestAndResponse {
+            logger.debug("Response: \(response.httpResponse.debugDescription)")
+        } else if clientLogMode == .requestAndResponseWithBody || clientLogMode == .responseWithBody {
+            logger.debug("Response: \(response.httpResponse.debugDescriptionWithBody)")
```
wooj2, line 42: We seem to be changing the behavior here in that before we were ignoring any exceptions being thrown and returning the response, but now, if getting the response throws an exception, we no longer return a response.  This might be exactly what we want to do, but not clear on if this is the expected behavior.

Packages/ClientRuntime/Tests/MiddlewareTests/MiddlewareStackTests.swift
```diff
@@ -45,40 +41,38 @@ class MiddlewareStackTests: XCTestCase {
...
+                                                return output
                                             }))
         
-        switch result {
-        case .success(let output):
-            XCTAssert(output.value == 200)
-        case .failure(let error):
-            XCTFail(error.localizedDescription)
-        }
+        //wait(for: [deserializeMiddleware], timeout: 0.3)
```
wooj2, line 71: remove the comment?

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/LoggerMiddleware.swift
kneekey23: don't see how we are changing the behavior? we just removed the do catch now that this is a throws function. if line 37 fails it will just throw and it will never get to line 39 exactly as the code was before?

Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
kneekey23: updated

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
kneekey23: not sure why I removed that, putting it back

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
kneekey23: okay happy to explore options with you...

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
kneekey23: no that would be impossible. if acquireConnection failed it would throw and the function would throw from this line and never get to the next line.

Packages/ClientRuntime/Tests/MiddlewareTests/MiddlewareStackTests.swift
kneekey23: removed, thanks

Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
```diff
@@ -5,11 +5,11 @@
...
-    public init() {
-    }
-    public func handle(context: HttpContext, input: SdkHttpRequest) -> Result<OperationOutput<Output>, SdkError<OutputError>> {
+public struct NoopHandler<Output: HttpResponseBinding>: Handler {
+    public init() {}
+    
+    public func handle(context: HttpContext, input: SdkHttpRequest) async throws -> OperationOutput<Output> {
         let output = OperationOutput<Output>(httpResponse: HttpResponse())
-        return .success(output)
+        return output
```
wooj2, line 13:
```
return OperationOutput<Output>(httpResponse: HttpResponse())
```
is probably more appropriate

Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
```diff
@@ -7,76 +7,68 @@ public struct OperationStack<OperationStackInput: Encodable & Reflection,
...
+              
+              let deserialize = compose(next: DeserializeStepHandler(handler: next), with: deserializeStep)
+              let finalize = compose(next: FinalizeStepHandler(handler: deserialize), with: finalizeStep)
+              let build = compose(next: BuildStepHandler(handler: finalize), with: buildStep)
+              let serialize = compose(next: SerializeStepHandler(handler: build), with: serializeStep)
+              let initialize = compose(next: InitializeStepHandler(handler: serialize), with: initializeStep)
+              
+              let result = try await initialize.handle(context: context, input: input)
+              guard let output = result.output else {
+                  throw ClientError.unknownError("Something went terribly wrong where the output was not set on the response. Please open a ticket with us at https://github.com/awslabs/aws-sdk-swift")
```
wooj2, line 44: Isn't it possible for requests not to have a response -- e.g. a http 204 response would indicate no body?  i don't think we should assume this is a client error, and seems like we should do some more thinking about how to handle this.

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
wooj2: Oup, nevermind -- i was reviewing an older version of `acquireConnection()`. I didn't noticed it changed from `acquireConnection() -> Future<HttpClientConnection>` to `acquireConnection() async throws`

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/LoggerMiddleware.swift
wooj2: my bad, I didn't see the try on line 41 of the old code.

Packages/ClientRuntime/Sources/Middleware/NoopHandler.swift
kneekey23: updated

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
wooj2: I re-read this code, and i think the only thing that is confusing is that we were not setting the `response.statusCode` in the `onIncomingBody` block nor the `onStreamComplete` body.
1.  Why is it now needed to set `response.statusCode` within the `onStreamComplete` block?
2. if it is needed, is there an else clause that needs to be filled out?  For example:
```
if let stream = stream {
    response.statusCode = ...
} else {
   response.statusCode = //does this need to be set?
}
```

kneekey23: investigated this, first of all looks like it was just a miss for why we weren't originally setting it additionally in onIncomingBody (looks like we set it in all other callbacks), secondly, there was no need for the stream to be optional in the crt code in onStreamComplete. I have updated both now. Thanks for the catch!

Packages/ClientRuntime/Sources/Middleware/OperationStack.swift
wooj2: I reviewed the protocol codegen tests for this -- and this is not a problem.  The only thing that we may need to update (in a separate PR) is that the generated test in AWSJson1_1/EmptyOperationResponseTest/testhandles_empty_output_shape is not doing any asserts.  We'll need to update our codegen to make sure it is comparing actual and expected.

wooj2 (approved):
LGTM!

aajtodd (approved):

PR #393, opened 2021-12-30
fix: Fixes an issue with creation of crt Logger

kneekey23:
## Issue \#

This corrects an issue @royjit found where one thread was creating the logger and another was creating the event loop in the crt and trying to call the logger before it had been created (race condition). the same thread should be creating both resources so the logger should be moved into the shared io file to be created there with all other crt types.

## Scope


wooj2 (approved):
lgtm!

kneekey23:
This PR closes awslabs/aws-sdk-swift#514

PR #395, opened 2022-01-13
fix: fixes s3 decode compiler errors

kneekey23:
## Description of changes

This PR fixes S3 compiler errors with decoding non optional associated values on enums. This additionally changes our decoding of enums with associated values from needing return statemetns everywhere to one switch statement based on the key in the container.

```swift
init(from decoder: Decoder) throws {
    let containerValues = try decoder.container(keyedBy: CodingKeys.self)
    let key = containerValues.allKeys.first
    
    switch key {
    case .empty:
        self = .empty
    case .domain:
        let domain = try container.decode(
            String.self,
            forKey: .editing
        )
        self = .domain(domain)
    case .port:
        let port = try container.decode(
            Int.self,
            forKey: .port
        )
        self = .port(port)
    default:
        self = .sdkUnknown("")
    }
}

```

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
@@ -39,28 +39,10 @@ abstract class MemberShapeDecodeXMLGenerator(
     abstract fun renderAssigningDecodedMember(memberName: String, decodedMemberName: String, isBoxed: Boolean = false)
     abstract fun renderAssigningSymbol(memberName: String, symbol: String)
     abstract fun renderAssigningNil(memberName: String)
+    abstract fun renderListMember(member: MemberShape, memberTarget: CollectionShape, containerName: String)
+    abstract fun renderMapMember(member: MemberShape, memberTarget: MapShape, containerName: String)
 
-    fun renderListMember(
-        member: MemberShape,
-        memberTarget: CollectionShape,
-        containerName: String,
-        isUnion: Boolean = false
-    ) {
-        val memberName = ctx.symbolProvider.toMemberName(member).removeSurrounding("`", "`")
-
-        if (isUnion) {
-            renderListMember(memberName, containerName, member, memberTarget)
-        } else {
-            writer.openBlock("if $containerName.contains(.$memberName) {", "} else {") {
-                renderListMember(memberName, containerName, member, memberTarget)
-            }
-            writer.indent()
-            renderAssigningNil(memberName)
-            writer.dedent().write("}")
-        }
-    }
-
-    private fun renderListMember(memberName: String, containerName: String, member: MemberShape, memberTarget: CollectionShape) {
+    fun renderListMember(memberName: String, containerName: String, member: MemberShape, memberTarget: CollectionShape) {
         val memberIsFlattened = member.hasTrait(XmlFlattenedTrait::class.java)
         var currContainerName = containerName
         var currContainerKey = ".$memberName"
@@ -166,21 +148,7 @@ abstract class MemberShapeDecodeXMLGenerator(
         }
     }
 
-    fun renderMapMember(member: MemberShape, memberTarget: MapShape, containerName: String, isUnion: Boolean = false) {
-        val memberName = ctx.symbolProvider.toMemberName(member)
-        if (isUnion) {
-            renderMapMember(member, memberTarget, containerName, memberName)
-        } else {
-            writer.openBlock("if $containerName.contains(.$memberName) {", "} else {") {
-                renderMapMember(member, memberTarget, containerName, memberName)
-            }
-            writer.indent()
-            renderAssigningNil(memberName)
-            writer.dedent().write("}")
-        }
-    }
-
-    private fun renderMapMember(member: MemberShape, memberTarget: MapShape, containerName: String, memberName: String) {
+    fun renderMapMember(member: MemberShape, memberTarget: MapShape, containerName: String, memberName: String) {
         val memberTargetValue = ctx.symbolProvider.toSymbol(memberTarget.value)
         val symbolOptional = if (ctx.symbolProvider.toSymbol(memberTarget).isBoxed()) "?" else ""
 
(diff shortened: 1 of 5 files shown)


smithy-swift-codegen/src/test/kotlin/serde/xml/UnionEncodeXMLGenerationTests.kt
```diff
@@ -65,93 +65,78 @@ class UnionEncodeXMLGenerationTests {
...
-                                self = .mapvalue(mapvalue)
-                                return
+                        case .unionvalue:
+                            let unionvalueDecoded = try containerValues.decode(Box<RestXmlProtocolClientTypes.XmlUnionShape>.self, forKey: .unionvalue)
+                            self = .unionvalue(unionvalueDecoded.value)
+                        case .structvalue:
+                            let structvalueDecoded = try containerValues.decode(RestXmlProtocolClientTypes.XmlNestedUnionStruct.self, forKey: .structvalue)
+                            self = .structvalue(structvalueDecoded)
+                        case .mapvalue:
+                            if containerValues.contains(.mapvalue) {
```
kneekey23, line 91: not sure if we need these if statements anymore but didn't want to remove just yet until I checked with you @wooj2

smithy-swift-codegen/src/test/kotlin/serde/xml/UnionEncodeXMLGenerationTests.kt
```diff
@@ -65,93 +65,78 @@ class UnionEncodeXMLGenerationTests {
...
+                                    self = .stringlist([])
                                 }
+                            } else {
+                                //No-op
                             }
-                            if let stringlist = stringlistBuffer {
-                                self = .stringlist(stringlist)
-                                return
+                        case .timestampvalue:
+                            let timestampvalueDecoded = try containerValues.decodeIfPresent(Swift.String.self, forKey: .timestampvalue)
```
kneekey23, line 131: per smithy team, union associated values should never be null so if we get to this piont in the switch statement I think we should use decode not decodeIfPresent and skip the null check. so I think we can simplify this code further actually. less code  == smaller binary :) @wooj2

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -293,14 +293,14 @@ abstract class MemberShapeDecodeXMLGenerator(
         renderAssigningDecodedMember(memberName, "$value")
     }
 
-    fun renderScalarMember(member: MemberShape, memberTarget: Shape, containerName: String, unkeyed: Boolean = false) {
+    fun renderScalarMember(member: MemberShape, memberTarget: Shape, containerName: String, unkeyed: Boolean = false, isUnion: Boolean = false) {
```
wooj2, line 296: 1.  I'm surprised that protocol codegen tests still pass with or without this change. I think we should add at least one protocol codegen test.  Clearly, we'll need one that addresses the decode, but I suspect we'll want to add one for encode as well.

kneekey23: can you elaborate further on what that protocol codegen test might be testing for? in this case, it seems we got the spec wrong and when its a union, its value can never be null. so we should never be using decodeIfPresent if we are decoding the associated value of an enum?

kneekey23: I think a unit test might suffice in this case

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -43,51 +43,60 @@ abstract class MemberShapeDecodeXMLGenerator(
     fun renderListMember(
         member: MemberShape,
         memberTarget: CollectionShape,
-        containerName: String
+        containerName: String,
+        isUnion: Boolean = false
```
wooj2, line 47: 1.  Rather than pass in isUnion to each of the functions, it seems more appropriate to add the following to MemberShapeDecodeXMLGenerator:
```
    abstract val isUnion: Boolean
```
2. The public functions that were modified to accept `isUnion` (e.g: `renderListMember`, `renderMapMember`, etc.) seem to be completely different between a union and a struct.  For that reason, it seems to make sense to put unique functions into UnionDecodeXMLGenerator and StructDecodeXMLGenerator rather than having a big if/else in the abstract class.  Doing so, would allow us to avoid creating an `abstract val isUnion`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -43,51 +43,60 @@ abstract class MemberShapeDecodeXMLGenerator(
...
+                renderListMemberItems(memberTarget, memberContainerName, memberBuffer)
             }
-            writer.indent()
-            renderAssigningSymbol(memberName, "[]")
-            writer.dedent().write("}")
+            renderAssigningDecodedMember(memberName, memberBuffer)
         }
         writer.indent()
-        renderAssigningNil(memberName)
+        renderAssigningSymbol(memberName, "[]")
```
wooj2, line 99: What was the need to assign [] vs nil? I'm surprised this didn't break a bunch of stuff?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
wooj2: I think i misunderstood the changes, yep, lgtm!

smithy-swift-codegen/src/test/kotlin/serde/xml/UnionEncodeXMLGenerationTests.kt
wooj2: handled offline, resolving.

smithy-swift-codegen/src/test/kotlin/serde/xml/UnionEncodeXMLGenerationTests.kt
wooj2: resolved offline., resolving.

wooj2 (approved):
Chatted offline - approving assuming that changes are made and unit tests do not need to be updated.

Thanks so much for handling this!

PR #397, opened 2022-01-19
chore: Adding extension to strip first prefix from string

wooj2:
Dependent on:
https://github.com/awslabs/aws-sdk-swift/pull/521

### DIFF
--- a/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
+++ b/Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
@@ -113,3 +113,14 @@ extension String {
         return ByteBuffer(data: self.data(using: .utf8) ?? Data())
     }
 }
+
+public extension String {
+    func stripFirstMatching(prefixes: [String]) -> String {
+        for prefixToStrip in prefixes {
+            if self.hasPrefix(prefixToStrip) {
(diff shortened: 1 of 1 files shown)


Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
```diff
@@ -113,3 +113,14 @@ extension String {
         return ByteBuffer(data: self.data(using: .utf8) ?? Data())
     }
 }
+
+public extension String {
+    func stripFirstMatching(prefixes: [String]) -> String {
+        for prefixToStrip in prefixes {
```
kneekey23, line 119:
```suggestion
        for prefixToStrip in prefixes where self.hasPrefix(prefixToStrip) {
```
could be shortened to the above :)

kneekey23 (approved):
LGTM

Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
wooj2: I didn't think of using `where` there, but semantically using `where` seems like it would imply returning an array (since we are starting with an array), and I'm only expecting one to case to hit.  I prefer the `if` statement over the `where`.  I'm open to hear other suggestions, but for now i'm going to leave it.

PR #398, opened 2022-01-20
feat: modularize the runtime layer

kneekey23:
## Description of changes

In an effort to decrease binary size, this PR modularizes the runtime layer into three targets, runtime, xmlruntime and jsonruntime. this way xml pieces never find their way into json sdks and vice versa.
Corresponding PR: https://github.com/awslabs/aws-sdk-swift/pull/528

## Scope


Packages/Package.swift
```diff
@@ -29,23 +31,45 @@ let package = Package(
             name: "ClientRuntime",
             dependencies: [
                 .product(name: "AwsCommonRuntimeKit", package: "AwsCrt"),
-                .product(name: "Logging", package: "swift-log"),
-                .product(name: "XMLCoder", package: "XMLCoder")
+                .product(name: "Logging", package: "swift-log")
             ],
-            path: "./ClientRuntime/Sources",
+            path: "./Runtime/Sources",
```
wooj2, line 36: Why not keep everything in a folder called `ClientRuntime` ?

Packages/Runtime/Tests/MiddlewareTests/MiddlewareStackTests.swift
```diff
@@ -54,7 +55,7 @@ class MiddlewareStackTests: XCTestCase {
         }
 
         stack.buildStep.intercept(position: .before, id: "add a header") { (context, input, next) -> Result<OperationOutput<MockOutput>, SdkError<MockMiddlewareError>> in
-            input.headers.add(name: "TestHeaderName2", value: "TestHeaderValue2")
+            input.withHeader(name: "TestHeaderName2", value: "TestHeaderValue2")
```
wooj2, line 58: Did we change an interface or something?  Not sure why this is needed?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftDependency.kt
```diff
@@ -27,6 +27,22 @@ enum class SwiftDependency(
         Resources.computeAbsolutePath("smithy-swift/Packages", "Packages", "SMITHY_SWIFT_CI_DIR") + "/Packages",
         "ClientRuntime"
     ),
+    JSON_RUNTIME(
+        "JSONRuntime",
+        "main",
+        "0.1.0",
```
wooj2, line 33: CLIENT_RUNTIME, JSON_RUNTIME and XML_RUNTIME will be versioned the same --- seems like it would make sense to create a constant private string ?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
```diff
@@ -101,8 +101,13 @@ class SwiftWriter(private val fullPackageName: String) : CodeWriter() {
...
-    fun addImport(packageName: String, isTestable: Boolean = false) {
-        imports.addImport(packageName, isTestable)
+    fun addImport(target: String, isTestable: Boolean = false) {
+        imports.addImport(target, isTestable)
+    }
+
+    fun addImport(dep: SymbolDependencyContainer, target: String) {
+        dependencies.addAll(dep.dependencies)
+        imports.addImport(target)
     }
```
wooj2, line 111: This function reads weird.  Seems like if we need access to `dependencies`, we should create a function called:
```
fun addDependencies(dependencies: List<SymbolDependency>) {
}
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolClientGenerator.kt
```diff
@@ -41,6 +41,11 @@ open class HttpProtocolClientGenerator(
         writer.addImport(SwiftDependency.CLIENT_RUNTIME.target)
         writer.addImport(SwiftDependency.SWIFT_LOG.target)
         writer.addFoundationImport()
+        if (ctx.protocol.name.lowercase().contains("json")) {
+            writer.addImport(SwiftDependency.JSON_RUNTIME.target)
+        } else {
+            writer.addImport(SwiftDependency.XML_RUNTIME.target)
+        }
```
wooj2, line 48: 1.  This code seems to be pretty repetitive. Can we add something to SwiftWriter like:
```
fun addSerdeDependency(protocol: ShapeId) {
    if (json) {

    } else if xml {
        ...
    }
```
I believe this can be used in cases we have either:
a: where we have common code for both json/xml like:
```
      if (ctx.protocol.name.lowercase().contains("json")) {
                    writer.addImport(SwiftDependency.JSON_RUNTIME.target)
                } else {
                    writer.addImport(SwiftDependency.XML_RUNTIME.target)
                }
```
b. where we have specific code for just xml like:
```
writer.addImport(SwiftDependency.XML_RUNTIME.target)
```
That way, we can have a common function to inject the correct import and not have to make changes across the codebase.



2. Note the `else if` above vs `else`.  I think it would be better to be explicit about which serde dependency we need.

jbelkins:
Closing due to abandonment.

PR #399, opened 2022-01-26
feat: add initialize/cleanup interface for users of clientruntime

wooj2:
Seems like we need an interface here, to avoid people calling directly into the CRT.... thoughts on naming?  If we are good with naming, i'll finish up documentation.

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/ClientRuntime.swift
+++ b/Packages/ClientRuntime/Sources/Networking/ClientRuntime.swift
@@ -18,7 +18,7 @@ public class ClientRuntime {
      this function once and only once, but it should be safe to call this multiple times during
      start up
      */
-    static func initialize() {
+    public static func initialize() {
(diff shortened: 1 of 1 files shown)


Packages/ClientRuntime/Sources/Networking/ClientRuntime.swift
```diff
@@ -0,0 +1,35 @@
...
+    public static func initialize() {
+        AwsCommonRuntimeKit.initialize()
+    }
+
+    /**
+    TODO:
+
+     - Note: It is expected that calling this function is idempotent. Do your best to call
+     this function once and only once, but it should be safe to call this multiple times during
+     start up
```
kneekey23, line 30: startup -> clean up

kneekey23 (approved):
Good with naming feel free to finish documentation :)

jbelkins:
Closing due to abandonment.

PR #410, opened 2022-02-26
feat: pagination

kneekey23:
## Description of changes

This PR implements the paginator trait in Smithy to provide paginators for services.

It will generate:
```swift
   /// Paginate over `[ListFunctionsOutputResponse]` results.
   ///
   /// When this operation is called, an `AsyncSequence` is created. AsyncSequences are lazy so no service
   /// calls are made until the sequence is iterated over. This also means there is no guarantee that the request is valid
   /// until then. If there are errors in your request, you will see the failures only after you start iterating.
   /// - Parameters:
   ///     - input: A `[ListFunctionsInput]` to start pagination
   /// - Returns: An `AsyncSequence` that can iterate over `ListFunctionsOutputResponse`
   extension LambdaClient {
       func listFunctionsPaginated(input: ListFunctionsInput) -> ClientRuntime.PaginatorSequence<ListFunctionsInput, ListFunctionsOutputResponse> {
           return ClientRuntime.PaginatorSequence<ListFunctionsInput, ListFunctionsOutputResponse>(input: input, inputKey: \ListFunctionsInput.marker, outputKey: \ListFunctionsOutputResponse.nextMarker, paginationFunction: self.listFunctions(input:))
       }
   }

   extension ListFunctionsInput: ClientRuntime.PaginateToken {
       public func usingPaginationToken(_ token: Swift.String) -> ListFunctionsInput {
           return ListFunctionsInput(
               functionVersion: self.functionVersion,
               marker: token,
               masterRegion: self.masterRegion,
               maxItems: self.maxItems
           )}
   }
```
Compilled all SDKs locally to prevent any compiler regressions. 
Corresponding PR with bug fix after compiling all sdks: https://github.com/awslabs/aws-sdk-swift/pull/553
This closes https://github.com/awslabs/aws-sdk-swift/issues/244
## Scope

- [ x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
+++ b/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
@@ -30,14 +30,16 @@ public struct PaginatorSequence<Input: PaginateToken, Output: HttpResponseBindin
         
         public mutating func next() async throws -> Output? {
             while token != nil || isFirstPage {
-                if let token = token {
-                    if (token is String && !(token as! String).isEmpty) || (token is [String: Any] && !(token as! [String: Any]).isEmpty) {
-                        self.input = input.usingPaginationToken(token)
-                    }
+
+                if let token = token, (token is String && !(token as! String).isEmpty) || (token is [String: Any] && !(token as! [String: Any]).isEmpty) {
+                    self.input = input.usingPaginationToken(token)
                 }
                 let output = try await sequence.paginationFunction(input)
                 isFirstPage = false
                 token = output[keyPath: sequence.outputKey]
+                if token == input[keyPath: sequence.inputKey!] {
+                    break
+                }
                 return output
             }
             return nil
--- a/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
+++ b/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
@@ -13,7 +13,7 @@ public struct PaginatorSequence<Input: PaginateToken, Output: HttpResponseBindin
     let paginationFunction: (Input) async throws -> Output
     
     public init(input: Input,
-                inputKey: KeyPath<Input, Input.Token?>?,
+                inputKey: KeyPath<Input, Input.Token?>? = nil,
                 outputKey: KeyPath<Output, Input.Token?>,
                 paginationFunction: @escaping (Input) async throws -> Output) {
         self.input = input
@@ -23,7 +23,6 @@ public struct PaginatorSequence<Input: PaginateToken, Output: HttpResponseBindin
     }
     
     public struct PaginationIterator: AsyncIteratorProtocol {
-        public typealias Element = Output
         var input: Input
         let sequence: PaginatorSequence
         var token: Input.Token? = nil
@@ -32,7 +31,9 @@ public struct PaginatorSequence<Input: PaginateToken, Output: HttpResponseBindin
         public mutating func next() async throws -> Output? {
             while token != nil || isFirstPage {
                 if let token = token {
-                    self.input = input.usingPaginationToken(token)
+                    if (token is String && !(token as! String).isEmpty) || (token is [String: Any] && !(token as! [String: Any]).isEmpty) {
+                        self.input = input.usingPaginationToken(token)
+                    }
                 }
                 let output = try await sequence.paginationFunction(input)
                 isFirstPage = false
--- a/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
+++ b/Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
@@ -5,7 +5,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-public struct PaginatorSequence<Input: PaginateToken, Output: HttpResponseBinding>: AsyncSequence where Input.Token: Equatable {
+public struct PaginatorSequence<Input: PaginateToken,
+                                Output: HttpResponseBinding>: AsyncSequence
+where Input.Token: Equatable {
     public typealias Element = Output
     let input: Input
     let inputKey: KeyPath<Input, Input.Token?>?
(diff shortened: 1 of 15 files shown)


Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
```diff
@@ -0,0 +1,49 @@
...
+        public typealias Element = Output
+        var input: Input
+        let sequence: PaginatorSequence
+        var token: Input.Token? = nil
+        var isFirstPage: Bool = true
+        
+        public mutating func next() async throws -> Output? {
+            while token != nil || isFirstPage {
+                if let token = token {
+                    self.input = input.usingPaginationToken(token)
```
rcoh, line 35: I'm haven't dug into this yet, but you need to make sure that you handle both the case where the pagination token is totally unset as well as the case where it's set to `""`

Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/AsyncSequence+Extension.swift
```diff
@@ -0,0 +1,22 @@
...
+        var values = [T]()
+
+        for try await element in self {
+            if let element = transform(element) {
+                values.append(contentsOf: element)
+            }
+        }
+
+        return values
+    }
```
rcoh, line 21: I guess this is a transformer to collect an async iterator into a list? a doc comment would probably be useful if customers are going to use it

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PaginatorGenerator.kt
```diff
@@ -0,0 +1,217 @@
...
+        - Returns: `${itemDesc.itemSymbol.fullName}`
+        """.trimIndent()
+
+        writer.writeSingleLineDocs {
+            this.write(docBody)
+        }
+
+        writer.openBlock("extension PaginatorSequence where Input == \$N, Output == \$N {", "}", inputSymbol, outputSymbol) {
+            writer.openBlock("func \$L() async throws -> \$N {", "}", itemDesc.itemLiteral, itemDesc.itemSymbol) {
+                writer.write("return try await self.asyncCompactMap { item in item.\$L }", itemDesc.itemPathLiteral)
```
rcoh, line 188: so does this always read to the end? (as in, it forces loading the entire stream into memory, no way to read it lazily)

rcoh (approved):
LGTM—I would strongly consider testing IRL or adding some more test cases, we found a number of issues in the Rust paginators.

EC2 paginator tests: https://github.com/awslabs/smithy-rs/blob/21ffa905dbc048dde6dff654bf953ee01df478b5/aws/sdk/integration-tests/ec2/tests/paginators.rs#L21

Dynamo Paginators test: https://github.com/awslabs/smithy-rs/blob/2e50dab45d5e3a541404f542ad14e0e8382cd325/aws/sdk/integration-tests/dynamodb/tests/paginators.rs#L19

Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
kneekey23: ah yes I forgot !token.isEmpty here cuz I was running into some compiler issues. let me add it back. good call out. and hm on the second one, ill add a check for that as well

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PaginatorGenerator.kt
kneekey23: ya I couldnt figure out how to return another `AsyncSequence` of that nested type. the higher level order functions like .map, or .forEach don't seem to cover this use case right now since this type is new to the language. I was thinking this is a temp situation until we can properly transform it and turn it into an `AsyncSequence` of the nested type so it's still lazy. unless you have a better suggestion? I tried like a million different thigns and couldn't get any of them to compile.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PaginatorGenerator.kt
```diff
@@ -0,0 +1,217 @@
...
+        }
+        val markerLiteral = paginationInfo.inputTokenMember.camelCaseName()
+        val docBody = """
+            Paginate over `[${outputSymbol.name}]` results.
+            
+            When this operation is called, an `AsyncSequence` is created. AsyncSequences are lazy so no service
+            calls are made until the sequence is iterated over. This also means there is no guarantee that the request is valid
+            until then. If there are errors in your request, you will see the failures only after you start iterating.
+            - Parameters: 
+                - input: A `[${inputSymbol.name}]` to start pagination
```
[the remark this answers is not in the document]
kneekey23: will add to these docs

Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
kneekey23: is token always a string? I think I remember why I didnt do this cuz I coudlnt figure out if token was always a string to make this check.

smithy-swift-codegen/src/test/resources/pagination.smithy
```diff
@@ -0,0 +1,72 @@
+namespace com.test
+
+use aws.protocols#restJson1
+
+service Lambda {
+    operations: [ListFunctions, ListFunctions2]
+}
+
+@paginated(
```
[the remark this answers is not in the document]
rcoh: add a model that uses `map` as the token

Packages/ClientRuntime/Sources/Pagination/PaginatorSequence.swift
kneekey23: handled both cases I think now

smithy-swift-codegen/src/test/resources/pagination.smithy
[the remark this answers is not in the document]
kneekey23: addressed

PR #413, opened 2022-03-15
fix: add Swift 5.6 case in swiftVersion property to resolve compiler error

atierian:
## Description of changes

smithy-swift (and thereby the AWSSwiftSDK) throw a compiler error when building with Swift 5.6
This is a quick fix that resolves the compiler error and doesn't break anything else based on manual testing.

Next step would be supporting Swift 5.6 language features, e.g. `any P`. That change should come in a subsequent feat PR so we can fix the existing issue quickly.

Side note: Swift 5.7 has been confirmed as the next major version. However If we want to keep this language version checking strategy going forward, it may make sense to switch to `> existing version` rather than `>= potential next version`. 

## Scope

- [ ] ~New feature (non-breaking change which adds functionality)~

### DIFF
--- a/Packages/ClientRuntime/Sources/Util/SwiftVersion.swift
+++ b/Packages/ClientRuntime/Sources/Util/SwiftVersion.swift
@@ -6,8 +6,10 @@
 //
 
 public var swiftVersion: String {
-    #if swift(>=5.6)
+    #if swift(>=5.7)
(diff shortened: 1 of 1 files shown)


kneekey23:
This is purely for grabbing the version number at runtime for the user agent String. No need for any follow on PR with language changes with this change. Also don't change to > because it's for the user agent String so that wouldn't make sense.

aajtodd (approved):

Issue #423, opened 2022-07-16
CRTClientEngine has a data race where connectionPools can be mutated from more than 1 thread


brennanMKE:
## Warning ##
* You should only be opening a bug report in this repository if it pertains to **non-AWS specific bugs pertaining to code generation of an SDK.**
* **If you have bug pertaining to the AWS SDK for Swift**, please open it in:
    * https://github.com/awslabs/aws-sdk-swift/blob/main/.github/ISSUE_TEMPLATE/bug_report.md

## Description

data race

## Environment

iOS & macOS

## Reproduction Steps

Run an app which uses many HTTP connections.

## Observed Behavior

Runtime exceptions due to bad access. These errors with calling `count` resulted in `unrecognized selector` with various types which were not the expected dictionary type.

```
-[__NSCFNumber count]: unrecognized selector sent to instance 0x8000000000000000
-[__NSTaggedDate count]: unrecognized selector sent to instance 0x8000000000000000
-[NSTaggedPointerString count]: unrecognized selector sent to instance 0x8000000000000000
```

## Possible Solution

Move the stored property and functions into an `actor` and call those functions so the state is protected from multiple threads.

Running Address Sanitizer and Thread Sanitizer many times helped to generate reports which pointed to this data race and others.

PR #424, opened 2022-07-16
fixes data race in CRTClientEngine

brennanMKE:
* creates nested actor named ConnectionPoolContainer in CRTClientEngine
* public functions were directed to the actor to manage connection pools

## Issue \#

#423

## Description of changes

Moves state management into an `actor` to make it thread-safe.

## Scope

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
@@ -11,9 +11,67 @@ import Darwin
 #endif
 
 public class CRTClientEngine: HttpClientEngine {
+    actor ConnectionPoolContainer {
+        private var logger: LogAgent
+
+        private let windowSize: Int
+        private let maxConnectionsPerEndpoint: Int
+        private var connectionPools: [Endpoint: HttpClientConnectionManager] = [:]
+
+        init(config: CRTClientEngineConfig) {
+            self.windowSize = config.windowSize
+            self.maxConnectionsPerEndpoint = config.maxConnectionsPerEndpoint
+            self.logger = SwiftLogger(label: "ConnectionPoolContainer")
+        }
+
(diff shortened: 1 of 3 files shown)


Packages/ClientRuntime/Sources/Networking/Http/SdkHttpClient.swift
```diff
@@ -24,7 +24,9 @@ public class SdkHttpClient {
     }
     
     public func close() {
-        engine.close()
+        Task.detached {
+            await self.engine.close()
+        }
```
ganeshnj, line 29: Wonder why `Task.detached`?

What if you just use `Task`
```
        Task {
            await self.engine.close()
        }
```

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -11,9 +11,67 @@ import Darwin
 #endif
 
 public class CRTClientEngine: HttpClientEngine {
+    actor ConnectionPoolContainer {
```
ganeshnj, line 14:
```suggestion
    actor ConnectionPoolManager {
```

It should be `ConnectonPoolManager` and there is no reason it to be nested class. Although, keep it internal.

ganeshnj (changes requested):

Packages/ClientRuntime/Sources/Networking/Http/CRT/CRTClientEngine.swift
brennanMKE: It is internal since it should only be used with the paren type. I also considered calling it `SerialExecutor` as the nested type because that is what it is.

Packages/ClientRuntime/Sources/Networking/Http/SdkHttpClient.swift
brennanMKE: done

PR #425, opened 2022-08-02
chore: update smithy version to 1.22.0

ganeshnj:
### Motivation

https://github.com/awslabs/aws-sdk-swift/pull/597 is failing due to unresolved shape `smithy.api#Unit` which was introduced in https://github.com/awslabs/smithy/pull/980/

### Changes

1. Updated smithyVersion to 1.22.0.
2. Update smithyGradleVersion to 0.6.0
3. Fix issues with header list parsing. The parsing was always correct but the test setup was not, where expected result always assumed that the list is a date list. To support this, I took kotlin implementation of string list parsing and translated to Swift with test cases.
4. `JsonName` trait now has no effect in AwsJson1_0 and AwsJson1_1, this change is covered at https://github.com/awslabs/aws-sdk-swift/pull/600
5. serde change: in case where the top container is sparse map which contains a dense list as value, the dense list was considered as sparse. Therefore, added a way to track the member parent to fix the nullability.
6. Sensitive trait fixes because now it can't target member directly.

### Result

AWS SDK handles `smithy.api#Unit` for operation correctly.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpHeaderProvider.kt
```diff
@@ -114,6 +115,8 @@ class HttpHeaderProvider(
                 writer.openBlock("if $memberName != ${member.defaultValue(ctx.symbolProvider)} {", "}") {
                     writer.write("items.add(Header(name: \"$paramName\", value: \$N($memberNameWithExtension)))", SwiftTypes.String)
                 }
+            } else if (inCollection && ctx.model.expectShape(member.target) !is TimestampShape) {
+                writer.write("items.add(Header(name: \"$paramName\", value: quoteHeaderValue(\$N($memberNameWithExtension))))", SwiftTypes.String)
             } else {
                 writer.write("items.add(Header(name: \"$paramName\", value: \$N($memberNameWithExtension)))", SwiftTypes.String)
             }
```
ianbotsf, line 122: **Question**: Why shouldn't timestamps get their value quoted like everything else?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -153,15 +154,15 @@ abstract class MemberShapeDecodeGenerator(
...
                 writer.write("\$L = \$N()", decodedMemberName, originalSymbol)
-                renderDecodeListTarget(nestedTarget, decodedMemberName, listContainerName, insertMethod, topLevelMember, shape.isSetShape, level)
+                renderDecodeListTarget(nestedTarget, decodedMemberName, listContainerName, insertMethod, topLevelMember, shape, level)
             }
             renderAssigningDecodedMember(topLevelMember, decodedMemberName)
         } else {
             writer.openBlock("if let \$L = \$L {", "}", memberName, memberName) {
                 val previousDecodedMemberName = "${memberName.removeSurroundingBackticks()}Decoded${level - 1}"
-                val symbolName = determineSymbolForShape(shape, isSparse)
+                val symbolName = determineSymbolForShape(shape, false)
```
ianbotsf, line 163: **Question**: Is `isSparse` used anywhere else in this method?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpHeaderProvider.kt
ganeshnj: Initially I was actually doing the quoting for all values (timestamp and string) and protocol tests were failing.

Then I looked into [kotlin implementation  ](https://github.com/awslabs/smithy-kotlin/blob/f6eeecb2cda7c345bef6f510dc30eac2c0697725/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializer.kt#L93-L120) which also quote only for plain string.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
ganeshnj: good catch, no reference for `isSparse`, cleaned.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpHeaderProvider.kt
ianbotsf: It seems like this code will quote everything that isn't a timestamp (i.e., `!is TimestampShape`). Have you verified this works with lists of booleans, numbers, etc.?

ganeshnj: Verified from protocol tests defined at https://github.com/awslabs/smithy[/blob/9af3fab79eeee645e28ca92e0b711cfa17472a5d/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy](https://github.com/awslabs/smithy/blob/9af3fab79eeee645e28ca92e0b711cfa17472a5d/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy

- "X-StringSet": "a, b, c"
- "X-IntegerList": "1, 2, 3",
- "X-BooleanList": "true, false, true"
- "X-TimestampList": "Mon, 16 Dec 2019 23:48:18 GMT, Mon, 16 Dec 2019 23:48:18 GMT"
- "X-EnumList": "Foo, Bar, Baz"

Also, [quoteHeaderValue](https://github.com/awslabs/smithy-swift/blob/jangirg/chore/bump-smithy-version/Packages/ClientRuntime/Sources/Serialization/SerializationUtils/HeaderUtils.swift#L77-L91) doesn't always quote. The string must contains quotable chars which is also translation of [kotlin](https://github.com/awslabs/smithy-kotlin/blob/9e5c5acc1922821b15ceff93385ea6230dce57ff/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/HeaderLists.kt#L122-L134) version.

ianbotsf: Makes sense, thanks!

ianbotsf (approved):
LGTM

Packages/ClientRuntime/Sources/PrimitiveTypeExtensions/String+Extensions.swift
```diff
@@ -124,3 +124,30 @@ public extension String {
...
+        prefix(range.upperBound.advanced(by: 1))
+    }
+    
+    subscript(_ range: PartialRangeUpTo<Int>) -> SubSequence {
+        prefix(range.upperBound)
+    }
+    
+    subscript(_ range: PartialRangeFrom<Int>) -> SubSequence {
+        suffix(Swift.max(0, count-range.lowerBound))
+    }
```
hectormatos2011, line 152: Aren't these native to Swift?

ganeshnj: Good call

There are [subscript implementation]( https://developer.apple.com/documentation/swift/string/subscript(_:)-lc0v) for string but they are `Index` based which are almost impossible to use in real life.

Packages/ClientRuntime/Tests/ClientRuntimeTests/PrimitiveTypeExtensionsTests/StringExtensionsTests.swift
```diff
@@ -96,4 +96,40 @@ class StringExtensionsTests: XCTestCase {
...
+        let string = "FooBar"
+        let range = string[..<3]
+        XCTAssertEqual("Foo", range)
+    }
+
+    func testPartialRangeFrom() {
+        let string = "FooBar"
+        let range = string[3...]
+        XCTAssertEqual("Bar", range)
+    }
```
hectormatos2011, line 134: Since emojis are valid identifiers for Swift, you should add tests for emojis here too. I think your subscript extensions won't work for any characters that aren't UTF-8

ganeshnj: AFIK, HTTP headers are only ASCII but some services may end up doing some crazy things. 

That said, now it is all `Index` based.

sugmanue (approved):

hectormatos2011 (approved):

PR #432, opened 2022-08-16
Updgrade 1.23

sugmanue:
Upgrade to Smithy version 1.23

## Description of changes

* Upgrade smithyVersion to 1.23.0 that adds support for IDL-2.0
* Avoid emitting the available attribute for shapes deprecated in the prelude, this will keep the output unchanged.
- Add the logic to support for enums shapes that now are their own type instead of being a trait over string shapes
- Fix the recursive logic to annotate shapes to avoid infinite recursion

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithHttpPayload.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithHttpPayload.kt
@@ -54,6 +54,13 @@ class HttpResponseTraitWithHttpPayload(
                     writer.indent()
                     writer.write("self.\$L = nil", memberName).closeBlock("}")
                 }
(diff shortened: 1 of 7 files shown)


ganeshnj:
Change looks good to me, please use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) message when you merge your PR.

i.e. `chore: update smithy version to 1.23.0`

ganeshnj (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithHttpPayload.kt
```diff
@@ -54,6 +54,13 @@ class HttpResponseTraitWithHttpPayload(
                     writer.indent()
                     writer.write("self.\$L = nil", memberName).closeBlock("}")
                 }
+                ShapeType.ENUM -> {
```
ganeshnj, line 57: It would have been nicer if you can add a unit test to exercise this code path.

PR #433, opened 2022-08-17
feat(endpoints)!: integrate endpoints 2.0

ganeshnj:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/624

## Description of changes

Counterpart: https://github.com/awslabs/aws-sdk-swift/pull/607

- Introduces endpoint string config at SDK config level
- Add supports for headers and attribute to `Endpoint` to allow returning headers and metadata (auth schemes) as part of endpoint resolution.
- Codegen update to start creating service specific configuration protocol.

Ex.
Before
```swift
protocol ExampleServiceProtocol  {
}
```
After
```swift
protocol ExampleServiceProtocol  {
}

protocol ExampleServiceConfigurationProtocol {
}
```

## Scope


Packages/ClientRuntime/Sources/Config/SDKRuntimeConfiguration.swift
```diff
@@ -17,6 +17,7 @@ public protocol SDKRuntimeConfiguration {
     var logger: LogAgent {get}
     var clientLogMode: ClientLogMode {get}
     var retryer: SDKRetryer {get}
+    var endpoint: String? {get set}
```
ianbotsf, line 20: **Question**: Why is a simple string sufficient for configuration? Won't you need an endpoint _resolver_ at the client level which returns an `Endpoint` struct (not a simple string)?

Packages/ClientRuntime/Sources/Networking/Endpoint.swift
```diff
@@ -12,16 +12,49 @@ public struct Endpoint: Hashable {
...
+    
+    public init(urlString: String,
+                headers: Headers? = nil,
+                properties: [String: AnyHashable] = [:]) throws {
+        guard let url = URL(string: urlString) else {
+            throw ClientError.unknownError("invalid url \(urlString)")
+        }
+        
+        try self.init(url: url, headers: headers, properties: properties)
+    }
```
ianbotsf, line 26: **Correctness**: After some protracted discussion with the EP2.0 authors, it sounds like we should not allow users to set `properties` given that it has no defined keys/values or stable usage semantics. The `Endpoint` type still needs this field but it should only be settable by other code within the SDK (not from without).

smithy-swift-codegen/build.gradle.kts
```diff
@@ -17,6 +17,10 @@ extra["moduleName"] = "software.amazon.smithy.swift.codegen"
 group = "software.amazon.smithy"
 version = "0.1.0"
 
+configurations.implementation {
+    exclude(group = "brazil")
+}
```
ianbotsf, line 22: **Question**: Why is this necessary?

gradle.properties
```diff
@@ -3,7 +3,7 @@ kotlin.code.style=official
 # config
 
 # codegen
-smithyVersion=1.23.0
+smithyVersion=1.23.1
```
ianbotsf, line 6: **Suggestion**: Consider using the latest version, **1.24.0**.

smithy-swift-codegen/build.gradle.kts
```diff
@@ -34,6 +38,8 @@ dependencies {
     implementation("software.amazon.smithy:smithy-aws-traits:$smithyVersion")
     testImplementation("org.junit.jupiter:junit-jupiter:$junitVersion")
     testImplementation("io.kotest:kotest-assertions-core-jvm:$kotestVersion")
+    implementation("software.amazon.smithy:smithy-rules-engine:$smithyVersion")
+    implementation("software.amazon.smithy:reterminus-core:0.2.0")
```
ianbotsf, line 42: **Correctness**: I cannot find **reterminus-core** [listed in Maven Central](https://repo1.maven.org/maven2/software/amazon/smithy/). Is this replaced by **smithy-rules-engine**?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
```diff
@@ -43,6 +43,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Void>() {
     private val integrations: List<SwiftIntegration>
     private val protocolGenerator: ProtocolGenerator?
     private val baseGenerationContext: GenerationContext
+    private var protocolContext: ProtocolGenerator.GenerationContext?
```
ianbotsf, line 46: **Nit**: Should be `val`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/ServiceConfig.kt
```diff
@@ -8,25 +8,27 @@ package software.amazon.smithy.swift.codegen.integration
...
 /**
  * Represents a config field on a client config struct.
  */
-data class ConfigField(val memberName: String?, val type: Symbol, val formatter: String = "\$N", private val documentation: String? = null)
+data class ConfigField(val memberName: String?, val type: Symbol, val propFormatter: String = "\$N", private val documentation: String? = null, val paramFormatter: String = "\$D")
 
 /**
  * ServiceConfig abstract class that allows configuration customizations to be configured for the protocol client generator
  */
 abstract class ServiceConfig(val writer: SwiftWriter, val serviceName: String) {
```
ianbotsf, line 21: **Question**: Why is this `abstract`? The only implementation I can see in this PR is `DefaultServiceConfig`. Are there (or will there be) different implementations?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -107,7 +113,7 @@ class ServiceGenerator(
      * We will generate the following:
      * ```
      * public protocol ExampleServiceProtocol {
-     *      func getFoo(input: GetFooInput, completion: @escaping (SdkResult<GetFooOutput, GetFooError>) -> Void)
+     *      func getFoo(input: GetFooInput) async throws -> GetFooResponse)
```
ianbotsf, line 116: **Nit**: Erroneous trailing `)`.

Packages/ClientRuntime/Sources/Config/SDKRuntimeConfiguration.swift
ganeshnj: This is to allow customers configure their endpoint by themselves. This is not particularly useful for AWS SDK, but required for generic SDKs, because `EndpointResolver` is not part of the base configuration anymore.

When the endpoint resolution happens, this string will be considered and prioritized.

Packages/ClientRuntime/Sources/Networking/Endpoint.swift
ganeshnj: This is good intention vs what can be implemented. `Endpoint` struct is part of the `ClientRuntime` (Shared code for Generic SDK as well as AWS SDK). Without making it public interface, it is impossible to use across boundaries.

smithy-swift-codegen/build.gradle.kts
ganeshnj: Required just for time being until dependencies are pushed to maven.

gradle.properties
ganeshnj: Wasn't aware of that, updated to latest.

smithy-swift-codegen/build.gradle.kts
ganeshnj: Correct, it will be part of the smithy-rules-engine.

There is already a PR https://github.com/awslabs/smithy/pull/1356 for it.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
ganeshnj: nice catch, fixed.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/ServiceConfig.kt
ganeshnj: Yes, there are two implementations of it. `DefaultServiceConfig` for generic codegen and `AWSServiceConfig` for AWS codegen.

https://github.com/awslabs/aws-sdk-swift/blob/5496c694c80fcfd0b3d825af632a55a677251f8a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceConfig.kt#L40

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
ganeshnj: Nice catch, fixed.

Packages/ClientRuntime/Sources/Config/SDKRuntimeConfiguration.swift
ianbotsf: This seems far less flexible than allowing a custom endpoint resolver. It only allows for a single URL regardless of operation, which may not be sufficient. It also doesn't allow for custom headers.

Packages/ClientRuntime/Sources/Networking/Endpoint.swift
ianbotsf: The Kotlin SDK uses an annotation called `@InternalApi` to indicate when something is `public` by necessity but shouldn't be used by callers unless they're part of the SDK (in which case, they must opt-in at the callsite with another annotation).

Does Swift SDK have any similar capabilities? If not, how will users know that they should not set this field?

ganeshnj: There isn't a swift provided attribute. Also, saying it internal is incorrect because if a customer wants to implement their own EndpointResolver, they have to provide `properties` with exact same structure which we expect else it will not work.

We have discussed this many times to not use dictionary but that doesn't satisfy endpoints 2.0 requirements.

jbelkins (approved):
Reviewed for awareness, not so much for correctness.

Packages/ClientRuntime/Sources/Config/SDKRuntimeConfiguration.swift
ganeshnj: Fair call, I have asked this question with endpoints 2.0 team. I agree, this is limiting and ideally we should allow overriding `Endpoint` type than just a `String` `endpoint`.

However, this would have ripple effect to other places if we want resolve endpoint at a common place which is ideal but we can short circuit if endpoint exists to reduce the scope of changes.

PR #434, opened 2022-08-23
fix: skip the deprecated trait only for shapes in the prelude

sugmanue:
## Description of changes
Some SDK inherit the deprecated traits from members, we should only skip them for deprecated shapes in the prelude. 
This change aligns this logic with the same logic used in [smithy-typescript](https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java#L277-L281).

## Scope

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ kotlin.code.style=official
 # config
 
 # codegen
-smithyVersion=1.23.0
+smithyVersion=[1.24.0,1.25.0[
 smithyGradleVersion=0.6.0
 
 # kotlin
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -3,7 +3,7 @@ kotlin.code.style=official
 # config
 
 # codegen
-smithyVersion=1.23.0
+smithyVersion=[1.23.0,1.24.0[
```
ganeshnj, line 6: I would rather prefer exact version than range. 

Any specific reason for range?

ganeshnj (changes requested):

gradle.properties
sugmanue: Yes, that way we can support any patch-level within the same major-minor, e.g., `1.23.1`. Having said that we are in the process to release `1.24.0` so we might need a follow up change for that.

jbelkins:
Closing due to staleness, after discussing with author we're not sure what problem this was intended to remedy.

PR #436, opened 2022-09-02
fix: use default value when deserializing non-optional members

ganeshnj:
## Issue \#
fixes https://github.com/awslabs/aws-sdk-swift/issues/564
fixes https://github.com/awslabs/aws-sdk-swift/issues/611

## Description of changes

Codegen uses default values to generate the init for non-optional values eg. https://github.com/awslabs/aws-sdk-swift/blob/0.2.6/release/AWSSecretsManager/models/Models.swift#L4148 but while doing the deserialization, the member is considered as always present which is not true in service like AWSSecretsManager.

This change uses the the same default value is available to set the member when the response doesn't contain the field. To generate less code, codegen skips assigning `nil`.

Before
```swift
let rotationEnabledDecoded = try containerValues.decode(Swift.Bool.self, forKey: .rotationEnabled)
rotationEnabled = rotationEnabledDecoded
```
After
```swift
let rotationEnabledDecoded = try containerValues.decodeIfPresent(Swift.Bool.self, forKey: .rotationEnabled) ?? false
rotationEnabled = rotationEnabledDecoded
```

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
@@ -20,6 +20,7 @@ import software.amazon.smithy.swift.codegen.SwiftWriter
 import software.amazon.smithy.swift.codegen.customtraits.SwiftBoxTrait
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
 import software.amazon.smithy.swift.codegen.integration.serde.MemberShapeDecodeGeneratable
+import software.amazon.smithy.swift.codegen.model.defaultValue
 import software.amazon.smithy.swift.codegen.model.hasTrait
 import software.amazon.smithy.swift.codegen.model.isBoxed
 import software.amazon.smithy.swift.codegen.model.recursiveSymbol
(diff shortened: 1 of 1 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -63,9 +64,14 @@ abstract class MemberShapeDecodeGenerator(
...
-        val decodeVerb = if (symbol.isBoxed()) "decodeIfPresent" else "decode"
+        val defaultValue = symbol.defaultValue()
+        val decodeVerb = if (symbol.isBoxed() || !defaultValue.isNullOrEmpty()) "decodeIfPresent" else "decode"
         val decodedMemberName = "${memberName}Decoded"
-        writer.write("let \$L = try \$L.$decodeVerb(\$N.self, forKey: .\$L)", decodedMemberName, containerName, symbol, memberName)
+
+        // no need to assign nil to a member that is optional
+        val defaultValueLiteral = if (defaultValue != null && defaultValue != "nil") "?? $defaultValue" else ""
+
+        writer.write("let \$L = try \$L.$decodeVerb(\$N.self, forKey: .\$L) $defaultValueLiteral", decodedMemberName, containerName, symbol, memberName)
```
jbelkins, line 74: nit: when there is no default value, this will result in trailing space.  Do we care about such things?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -63,9 +64,14 @@ abstract class MemberShapeDecodeGenerator(
...
             symbol = symbol.recursiveSymbol()
         }
-        val decodeVerb = if (symbol.isBoxed()) "decodeIfPresent" else "decode"
+        val defaultValue = symbol.defaultValue()
+        val decodeVerb = if (symbol.isBoxed() || !defaultValue.isNullOrEmpty()) "decodeIfPresent" else "decode"
         val decodedMemberName = "${memberName}Decoded"
-        writer.write("let \$L = try \$L.$decodeVerb(\$N.self, forKey: .\$L)", decodedMemberName, containerName, symbol, memberName)
+
+        // no need to assign nil to a member that is optional
+        val defaultValueLiteral = if (defaultValue != null && defaultValue != "nil") "?? $defaultValue" else ""
```
jbelkins, line 72: Kotlin Q: is there a "helper function" that can clean up `defaultValue != null && defaultValue != "nil"`, akin to `isNullOrEmpty()`?

jbelkins (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
ganeshnj: there is `isNullOrEmpty` but here we are checking the actual string which has Swift code (default value).

PR #437, opened 2022-09-02
chore: Project cleanup

### LINKED ISSUE #421: Fix Xcode support by removing Packages.swift from Packages folder
When working in Xcode the contents of Packages are hidden because it contains `Package.swift`. The purpose of it is not known yet it does cause difficulties when working with this package.

jbelkins:
## Issue \#

Fixes #421 by removing an unused `Package.swift` file contained within this project.

## Description of changes
- Fix formatting in `Package.swift`.
- Delete several files which are unneeded or, in the case of a nested and unused `Package.swift` file, were causing undesired side-effects with IDE features. (#421)
- Correct a couple Swift compile warnings caused by `try`-ing actions that don't throw
- Ignore Visual Studio Code preference files in `.vscode`
- Fix several Swift linter errors

## Scope


Package.swift
```diff
@@ -1,50 +1,42 @@
 // swift-tools-version:5.4
+
 import PackageDescription
-let excludes = ["README.md"]
 
 let package = Package(
-    name: "ClientRuntime",
```
ganeshnj, line 6: We must not rename the packages yet. This is something we have to confirm with Smithy team.

Historically we have chosen `ClientRuntime` suffix which I don't necessarily agree with comparing with other SDKs. let's keep them as they are as of now so we only break customers once.

ganeshnj: @mtdowling is the person to go and confirm the names here and shorter is better for less symbol size.

Few suggestions
Runtime suffix
- SmithyRuntime
- AWSRuntime

Core suffix
- SmithyCore
- AWSCore

ganeshnj (changes requested):

ganeshnj:
You would need to create similar change in https://github.com/awslabs/aws-sdk-swift in `jbe/swift_package_reorg`

CI / downstream job exercise the SDK and looks for the same branch name with main branch as fallback.

ganeshnj:
Also, we shouldn't move the Runtime code to top level else it will mixup with codegen part. Let's keep the Packages dir (will rename it based on feedback from Smithy team) and only delete the nested Package.swift file and modify the root Package.swift to include `Packages/blah` things

Package.swift
jbelkins: I am rolling this name change back in this PR, but I'm also filing an issue on it because I think this should be explored prior to the 1.0.

jbelkins: https://github.com/awslabs/smithy-swift/issues/438

jbelkins:
@ganeshnj I rolled back the changes to public names and project structure.  I filed GH issues for each of those changes, though, because I think they are both worth pursuing prior to 1.0 of this project.  Each should be done once we can address any concerns that the Smithy team may have.

jbelkins:
The `CI/downstream` build task is failing because the `Project.swift` file referenced by `aws-sdk-swift` has been deleted in this PR.  Merging https://github.com/awslabs/aws-sdk-swift/pull/616 should correct the build failure.

ganeshnj:
> The `CI/downstream` build task is failing because the `Project.swift` file referenced by `aws-sdk-swift` has been deleted in this PR. Merging [awslabs/aws-sdk-swift#616](https://github.com/awslabs/aws-sdk-swift/pull/616) should correct the build failure.

You might want to create the [awslabs/aws-sdk-swift#616](https://github.com/awslabs/aws-sdk-swift/pull/616) again  with the same branch name as this PR branch name.

The way the downstream check works, it checks the branch name same as this PR i.e. `jbe/swift_package_reorg` and if exists it uses it. This way, we can make sure any change done in either smithy-swift or aws-sdk-swift, both repos are up to date. We must* never merge a change where downstream is failing.

jbelkins:
@ganeshnj There were two more `Project.swift` files buried in the AWS Swift SDK project that had to be modified in https://github.com/awslabs/aws-sdk-swift/pull/616 for this to build successfully.

Please continue review, I think this PR is good to merge now.

ganeshnj (approved):

PR #440, opened 2022-09-12
fix: properly check if a member can be nullable

sugmanue:
## Description of changes

Change the way we compute the nullability of members. Before we were only looking at the box trait in the target type for the `box` trait and ignoring the trait on the member itself. See for instance the test `can read box trait from member` which has the following model:

```
        namespace com.test
        structure MyStruct {
           @box
           foo: MyFoo
        }
        long MyFoo
```
The code was incorrectly assuming that the `foo` member was not nullable. This change accounts for this by using the `NullableIndex` which correctly computes whether or not the member or the target type are nullable. This change in turn was also required in several places that were making the same mistake and assuming that computing the `Symbol` for the target and testing for `isBoxed()` was enough to know if the member is itself boxed.

After this change several models from the `aws-sdk-swift` generate different code, I checked and all the changes seems legit.

**NOTE** this is a draft change as the smithy release `1.25.0` is still pending.

## Scope

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ kotlin.code.style=official
 # config
 
 # codegen
-smithyVersion=1.23.0
+smithyVersion=[1.25.0,1.26.0[
 smithyGradleVersion=0.6.0
 
 # kotlin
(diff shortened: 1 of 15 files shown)


ganeshnj:
I just realized that this PR is made from the fork. @sugmanue can you push the PR to https://github.com/awslabs/smithy-swift and create a PR again?

downstream CI doesn't work with forks.

sugmanue:
@ganeshnj I will cancel this one and create a new PR.

PR #442, opened 2022-09-15
fix: Escape special characters in URL path prior to constructing URL

jbelkins:
## Issue \#
Fixes:
- https://github.com/awslabs/aws-sdk-swift/issues/613
- https://github.com/awslabs/aws-sdk-swift/issues/615

## Description of changes
There have been several instances where the AWS request signature on a HTTP request was invalid because the URL path contained special characters, usually resulting from the path including some sort of user input (i.e. a S3 bucket key that includes special characters.)

While I am making this change specifically to address AWS code signing, it should provide correct & safe special character handling for all Smithy applications.

Also includes a unit test to ensure proper percent encoding of a few special characters that have caused trouble for users, and for compliance with [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.3).

## Scope
- [X] Bug fix (non-breaking change which fixes an issue)


.swiftpm/xcode/xcshareddata/xcschemes/ClientRuntime-Package.xcscheme
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ganeshnj, line 1: this shouldn't be required. XCode already supports Swift packages and for Linux is it of no use.

Packages/ClientRuntime/Sources/Networking/Http/SdkHttpRequest.swift
```diff
@@ -29,12 +29,22 @@ public class SdkHttpRequest {
...
 }
 
+// Create a `CharacterSet` of the characters that need not be percent encoded in the
+// resulting URL.  This set consists of alphanumerics plus underscore, dash, tilde, and
+// period.  Any other character should be percent-encoded when used in a path segment.
+// Forward-slash is added as well because the segments have already been joined into a path.
+//
+// See, for URL-allowed characters:
+// https://www.rfc-editor.org/rfc/rfc3986#section-2.3
+private let allowed = CharacterSet.alphanumerics.union(CharacterSet(charactersIn: "/_-.~"))
```
ganeshnj, line 39: does S3 follow the same spec?

It would be nice to add a reference to the S3 documentation.

.swiftpm/xcode/xcshareddata/xcschemes/ClientRuntime.xcscheme
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ganeshnj, line 1: same as above.

.swiftpm/xcode/xcshareddata/xcschemes/SmithyTestUtil.xcscheme
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ganeshnj, line 1: same as above.

ganeshnj (changes requested):

.swiftpm/xcode/xcshareddata/xcschemes/ClientRuntime-Package.xcscheme
jbelkins: I will remove this and the other `.xcscheme`s from this PR.  The auto-generated schemes will not run tests because the package definition includes multiple test targets and Xcode cannot auto-match them, hence the manually created schemes.

I think this can be worked around by creating non-shared schemes that live on the developer's machine only, so long as out `.gitignore` file is properly configured.

jbelkins: Removed.

.swiftpm/xcode/xcshareddata/xcschemes/ClientRuntime.xcscheme
jbelkins: Removed.

.swiftpm/xcode/xcshareddata/xcschemes/SmithyTestUtil.xcscheme
jbelkins: Removed.

Packages/ClientRuntime/Sources/Networking/Http/SdkHttpRequest.swift
jbelkins: As far as I can tell, S3 docs never explicitly say how to encode they key for use in the URI.  But it makes sense that it must be at least encoded per the web standards to form a valid URI path.

S3 has [this doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html) that discusses characters that are & aren't allowed in keys, but it doesn't really talk about how to encode them in a URL.

Last, we must remember that this is in Smithy and that this code is not specific to AWS or S3.  So this must behave for any web service.  Encoding all but the explicitly allowed characters should be the most safe option.

ganeshnj (approved):

PR #450, opened 2022-10-12
feat: Add waiter client runtime components

### LINKED ISSUE #447: Create smithy-swift ClientRuntime additions for Waiters
Create reusable types for implementing waiters.  The types are described in detail in [the Waiters design doc for Swift SDK](https://quip-amazon.com/dLUUAwMak4ka/Waiters-Design#temp:C:MbW21ccff1020fb4df88c77bde15).

Requirements:
- Implement all `ClientRuntime` types described in the design doc
- Cover these types with unit tests
- Perform a wait operation by manually configuring these types to wait on a resource

When complete, this work should merge to a Waiters project branch while awaiting the completion of the rest of the Waiters project.

jbelkins:
## Issue \#
Closes #447

## Description of changes
Adds Swift types to the `ClientRuntime` package that implement waiters.
Also adds unit tests to verify & document their proper operation.

This code may be bench-tested in a CLI tool or other simple app using AWS Swift SDK's S3 HeadBucket (assumes you have AWS credentials set up):
```
import ArgumentParser
import ClientRuntime

let config = try WaiterConfig<HeadBucketInput, HeadBucketOutputResponse>(minDelay: 2.0, maxDelay: 10.0, acceptors: [
    WaiterConfig.Acceptor(state: .success, matcher: { _, result in
        if case .success = result { return true } else { return false }
    }),
    WaiterConfig.Acceptor(state: .retry, matcher: { _, _ in return true })
])
let client = try S3Client(region: "us-east-1")
let waiter = Waiter(config: config, operation: client.headBucket(input:))
let options = WaiterOptions(maximumWaitTime: 60.0)
let input = HeadBucketInput(bucket: "<myBucketName>")
let result = try await waiter.waitUntil(options: options, input: input)
print("Wait succeeded after \(result.attempts) attempts")
```
While this code runs, either:
- Create the bucket with name specified in `<myBucketName>` to allow waiter to succeed
- Do nothing to allow the waiter to timeout
- Something else to cause an error while waiting and "fail fast"

Also: enables CI on all PRs (not just ones against `main`).

## Scope


Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
```diff
@@ -0,0 +1,27 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// `Acceptor` is a Swift-native equivalent of Smithy acceptors:
+/// https://awslabs.github.io/smithy/2.0/additional-specs/waiters.html#acceptor-structure
+public struct Acceptor<Input, Output> {
```
epau, line 12: Nit: It'd be helpful to provide a description as to what an _Acceptor_ is or what is does

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
```diff
@@ -0,0 +1,27 @@
...
+        case failure
+    }
+
+    public let state: State
+    public let matcher: Matcher
+
+    public init(state: State, matcher: Matcher) {
+        self.state = state
+        self.matcher = matcher
+    }
```
epau, line 26: Nit: Doc-comments for each of these would be very helpful

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension Acceptor {
+
+    /// `Acceptor.Matcher` is a Swift-native equivalent of Smithy matchers:
+    /// https://awslabs.github.io/smithy/2.0/additional-specs/waiters.html#matcher-union
+    public enum Matcher {
```
epau, line 14: Nit: Similar to a comment above, it'd be helpful to provide a description as to what a _Matcher_ is or what is does

epau, line 14: Nit: Feel free to not-change, I leave that to your best judgement but a struct might be more appropriate here rather than an enum. Switching over an enum from within the same enum is usually a smell that the construct might be better modeled as a struct. 

Some rough code to show an example: 
```swift
public struct Matcher {
    private let evaluateMatch: (Input, Result<Output, Error>) -> Bool
    private init(_ evaluateMatch: () -> Bool) {
        self.evaluateMatch = evaluateMatch
    }

    public func hasAMatch(input: Input, result: Result<Output, Error>) -> Bool {
        evaluateMatch(input, result)
    }

    // MARK - Factory

    public static func output(_ f: @escaping (Output) -> Bool) -> Self {
        Matcher { input, result in
            switch result {
            case .succes(let output):
                result f(output)
            case .failure:
                return false
            }
        }
    }

    public static func inputOutput(_ f: @escaping (Input, Output) -> Bool) -> Self {
        // ..
    }

    // ..
}
```

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
```diff
@@ -0,0 +1,45 @@
...
+        case output((Output) -> Bool)
+        case inputOutput((Input, Output) -> Bool)
+        case success(Bool)
+        case errorType((Error) -> Bool)
+
+        func isAMatch(input: Input, output: Output?, error: Error?) -> Bool {
```
epau, line 20: Is it acceptable that this implementation allows for more flexibility than the [spec defines](https://awslabs.github.io/smithy/2.0/additional-specs/waiters.html#matcher-union)
For example, this implementation allows for the caller to define "has a match" as any logic they want, where as the spec has tight controls and restricts it to a set number (currently 4) comparators. etc.

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
```diff
@@ -0,0 +1,45 @@
...
+            switch self {
+            case .output(let testClosure):
+                guard let output = output else { return false }
+                return testClosure(output)
+            case .inputOutput(let testClosure):
+                guard let output = output else { return false }
+                return testClosure(input, output)
+            case .success(let success):
+                let hasOutput = output != nil
+                return success ? hasOutput : !hasOutput
```
epau, line 30: Should we also check if error is nil here? 
(This might not be necessary if the previous comment is addressed)

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+public class Waiter<Input, Output> {
+    public let input: Input
+    public let acceptors: [Acceptor<Input, Output>]
+    let minDelay: TimeInterval
+    let maxDelay: TimeInterval
+    let maximumWaitTime: TimeInterval
+    public let operation: (Input) async throws -> Output
+
+    public init(input: Input, acceptors: [Acceptor<Input, Output>], minDelay: TimeInterval, maxDelay: TimeInterval,
+                maximumWaitTime: TimeInterval, operation: @escaping (Input) async throws -> Output) {
```
epau, line 21: Nit: To provide a consistent format/style when using line breaks, can we put each argument on its own line. This follows the style that is used within the Swift standard library. (We should have our linter catch this as well)

```suggestion
    public init(
        input: Input, 
        acceptors: [Acceptor<Input, Output>],
        minDelay: TimeInterval, 
        maxDelay: TimeInterval,
        maximumWaitTime: TimeInterval, 
        operation: @escaping (Input) async throws -> Output
    ) {
```

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Makes a waiter operation, as configured by the caller, into an async sequence of waiter status events.
+/// Intended to be a generic type for use when waiting on any Smithy `operation`.
```
epau, line 11: Nit: 

>Makes a waiter operation, as configured by the caller, into an async sequence of waiter status events.

I think this might be a more appropriate comment for `public func asyncSequence() -> WaiterSequence<Input, Output>` below

>/// Intended to be a generic type for use when waiting on any Smithy `operation`.

I'm struggling a bit to understand the meaning and purpose of this comment. Is it necessary?

Lastly, perhaps we can provide a detailed comment here taking from the Smithy docs

```suggestion
/// A Waiter is a client-side abstraction used to poll a resource until a desired state is reached, or until it is determined that the resource will never enter into the desired state.
```

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+public class Waiter<Input, Output> {
+    public let input: Input
+    public let acceptors: [Acceptor<Input, Output>]
```
epau, line 14: Nit: Missing doc-comments

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+    let minDelay: TimeInterval
+    let maxDelay: TimeInterval
+    let maximumWaitTime: TimeInterval
+    public let operation: (Input) async throws -> Output
```
epau, line 18: Nit: Missing doc-comments

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    public init(input: Input, acceptors: [Acceptor<Input, Output>], minDelay: TimeInterval, maxDelay: TimeInterval,
```
epau, line 20: Nit: Missing doc-comments

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,33 @@
...
+                maximumWaitTime: TimeInterval, operation: @escaping (Input) async throws -> Output) {
+        self.input = input
+        self.acceptors = acceptors
+        self.minDelay = minDelay
+        self.maxDelay = maxDelay
+        self.maximumWaitTime = maximumWaitTime
+        self.operation = operation
+    }
+
+    public func asyncSequence() -> WaiterSequence<Input, Output> {
```
epau, line 30: Nit: Missing doc-comments

Packages/ClientRuntime/Sources/Waiters/WaiterIterator.swift
```diff
@@ -0,0 +1,86 @@
...
+/// ends iteration.
+public class WaiterIterator<Input, Output>: AsyncIteratorProtocol {
+    let input: Input
+    let acceptors: [Acceptor<Input, Output>]
+    let scheduler: WaiterScheduler
+    let operation: (Input) async throws -> Output
+
+    private var finished = false
+
+    init(input: Input, acceptors: [Acceptor<Input, Output>], minDelay: TimeInterval, maxDelay: TimeInterval, maximumWaitTime: TimeInterval, operation: @escaping (Input) async throws -> Output) {
```
epau, line 22: Nit: Can we put arguments on new lines when we have more than ~3 or 4? (Perhaps we should add a linter rule for this

```suggestion
    init(
        input: Input, 
        acceptors: [Acceptor<Input, Output>], 
        minDelay: TimeInterval, 
        maxDelay: TimeInterval, 
        maximumWaitTime: TimeInterval, 
        operation: @escaping (Input) async throws -> Output
    ) {
```

Packages/ClientRuntime/Sources/Waiters/WaiterIterator.swift
```diff
@@ -0,0 +1,86 @@
...
+        self.input = input
+        self.acceptors = acceptors
+        self.operation = operation
+        self.scheduler = WaiterScheduler(minDelay: minDelay, maxDelay: maxDelay, maximumWaitTime: maximumWaitTime)
+    }
+
+    // MARK: - AsyncIteratorProtocol
+
+    public func next() async throws -> WaiterStatus<Output>? {
```
epau, line 31: Nit: Missing doc-comment

Packages/ClientRuntime/Sources/Waiters/WaiterIterator.swift
```diff
@@ -0,0 +1,86 @@
...
+        // When success is reached, the finished flag is set.
+        // Returning nil from this function causes the async iterator
+        // to end.
+        // The iterator ends automatically after an error is thrown, so
+        // no need to set finished (thereby nil-terminating the
+        // AsyncSequence) after a failure event.
```
epau, line 37: This and the other comments in this function are super helpful! Thank you!!!

Packages/ClientRuntime/Sources/Waiters/WaiterIterator.swift
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+
+        // If no acceptor matched the response and an error was thrown, fail with the error.
+        // Otherwise, initiate a retry.
+        if let error = error {
+            throw error
+        } else {
+            return .retry(scheduler.updateAfterRetry())
+        }
```
epau, line 84: Nit: It might improve readability to use `first(where:)` here instead of a for-loop. This reduces levels of indentation and maintains a consistent function structure where we return early for all edge cases and handle the main case at the bottom. 

Some rough code showing an example:

```swift
guard let matchingAcceptor = acceptors.first { acceptor in
    acceptor.matcher.isAMatchinput: input, output: output, error: error)
} else {
    // If no acceptor matched the response and an error was thrown, fail with the error.
    // Otherwise, initiate a retry.
    if let error = error {
        throw error
    } else {
        return .retry(scheduler.updateAfterRetry())
    }
}

switch matchingAcceptor.state {
case .success:
    finished = true
    guard let output = output else {
        throw ClientError.unknownError("`success` Acceptor succeeded with nil output")
    }
    return .success(output)
case .retry:
    return .retry(scheduler.updateAfterRetry())
case .failure:
    throw error ?? ClientError.unknownError("Failure condition encountered while waiting")
}
```

Packages/ClientRuntime/Sources/Waiters/WaiterSequence.swift
```diff
@@ -0,0 +1,22 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Enables a wait operation to be expressed as a sequence of asynchronous events.
+public struct WaiterSequence<Input, Output>: AsyncSequence {
```
epau, line 11: Do we need to define a public init so that SDK consumers can create instance of `WaiterSequence` if need be?

Packages/ClientRuntime/Sources/Waiters/WaiterSequence.swift
```diff
@@ -0,0 +1,22 @@
...
+    public typealias AsyncIterator = WaiterIterator<Input, Output>
+    public typealias Element = WaiterStatus<Output>
+
+    public let waiter: Waiter<Input, Output>
```
epau, line 15: Missing doc-comment

Packages/ClientRuntime/Sources/Waiters/WaiterStatus.swift
```diff
@@ -0,0 +1,24 @@
...
+    case success(Output)
+    case retry(WaiterRetryInfo)
+}
+
+/// `WaiterRetryInfo` is provided to a caller when a wait operation retries,
+/// to keep the caller informed about the progress and status of waiting.
+public struct WaiterRetryInfo {
+    public let attempt: Int
+    public let timeUntilNextAttempt: TimeInterval
+    public let timeUntilTimeout: TimeInterval
```
epau, line 23: Missing doc-comments

epau (commented):
This is looking really good!! I have a few questions and mostly nit comments around styling and missing comments.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
jbelkins: Added

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
jbelkins: I've overhauled Matcher, it's now closer to your suggestion, thanks

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
```diff
@@ -0,0 +1,45 @@
...
+            case .errorType(let testClosure):
+                guard let error = error else { return false }
+                return testClosure(error)
+            }
+        }
+    }
+
+    /// `Acceptor.Result` encapsulates the action required by an `Acceptor` that matches the
+    /// operation's response.
+    public enum Result {
```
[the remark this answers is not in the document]
jbelkins: Removed.  In any case I shouldn't have named something to class with a commonly used Swift or Foundation type.

Packages/ClientRuntime/Sources/Waiters/WaiterIterator.swift
jbelkins: The updated code uses `compactMap` rather than loop.

Packages/ClientRuntime/Sources/Waiters/Matcher.swift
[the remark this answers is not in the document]
jbelkins: Done.  As you'll see, it works well with switch

jbelkins: Unless someone really goes deep-diving, matchers should always be code-generated, so I'm not too worried about a friendly interface at this point

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
```diff
@@ -0,0 +1,65 @@
...
+
+import Foundation
+
+extension WaiterConfig {
+    /// `Acceptor` is a Swift-native equivalent of Smithy acceptors:
+    /// https://awslabs.github.io/smithy/2.0/additional-specs/waiters.html#acceptor-structure
+    public struct Acceptor {
+
+        public typealias Matcher = (Input, Result<Output, Error>) -> Bool
+
```
jbelkins, line 16: lol this is a dumb linter rule.

Speaking of dumb linter rules, we don't seem to have any `.swiftlint` file in this project.  Wonder what rules we are using?

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
jbelkins: I've reformatted my method signatures this way, at least the ones with more than one param.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
ganeshnj: https://github.com/awslabs/smithy-swift/blob/main/Packages/.swiftlint.yml

either update the rule or fix the issue.

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    /// The configuration this waiter was created with.
+    public let config: WaiterConfig<Input, Output>
+
+    /// The operation that this waiter will call one or more times while waiting on the success condition.
+    public let operation: (Input) async throws -> Output
+
+    /// Allows for advancing of time during waiter unit tests.
+    /// The default hook does nothing.
+    var postSchedulerUpdateHook: (WaiterScheduler) -> Void = { _ in }
```
ganeshnj, line 23: is this for unit tests only? It is bad approach  if something is needed just for testing. It should be refactored.

Packages/ClientRuntime/Sources/Waiters/WaiterConfig.swift
```diff
@@ -0,0 +1,41 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// `WaiterConfig` contains the parameters for a predefined waiter on a given operation.  `WaiterConfig` will
+/// usually be code-generated from a Smithy definition for a Smithy waiter defined on an operation.
+public struct WaiterConfig<Input, Output> {
```
ganeshnj, line 12: trivial: `WaiterConfiguration`

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,103 @@
...
+/// An object used to wait on an operation.
+/// Intended to be a generic type for use when waiting on any Smithy `operation`.
+/// May be reused for multiple waits, including concurrent operations.
+public class Waiter<Input, Output> {
+
+    /// The configuration this waiter was created with.
+    public let config: WaiterConfig<Input, Output>
+
+    /// The operation that this waiter will call one or more times while waiting on the success condition.
+    public let operation: (Input) async throws -> Output
```
ganeshnj, line 19: why is operation tied to the Waiter itself? This means customer (even AWS SDK) have to use separate waiter for each operation.

Also saving a method signature seems like a quick solution but what would happen if add an extra optional param in the operation?

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import Foundation
+
+/// Calculates the time for the next retry, and whether waiting should continue.
+/// Upon creation, the scheduler will allow an immediate request.
+/// After the first request, further requests are scheduled per the retry strategy formula
+/// published in the Smithy specification:
+/// https://awslabs.github.io/smithy/2.0/additional-specs/waiters.html#waiter-retries
+class WaiterScheduler {
```
ganeshnj, line 15: This not how we should schedule. What is happening now

We have a standard retry strategy applied. That is going to try 3 times (I think that's we have configured) withe exponential backoff with jitter. On top that this scheduler works.

Rather we want to use the config provided as part of the waiter and apply them to the original retry strategy.

ganeshnj (changes requested):

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
jbelkins: I can refactor tests to use dependency injection of a mocked-out `WaiterScheduler`, but it will add at least as much complexity as this does.  I normally prefer not to add properties like this that are for testing, but it seemed the best option here.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
jbelkins: Rule removed.  This is better left to the dev case-by-case IMO

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,110 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// An object used to wait on an operation.
```
epau, line 10: Nit:
```suggestion
/// An object used to wait on an operation until a desired state is reached, or until it is determined that the desired state will never be reached.
```

Packages/ClientRuntime/Sources/Waiters/WaiterConfig.swift
jbelkins: Changed

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
jbelkins: Per our discussion, keeping waiter retry as its own mechanism that "wraps" the client & operation.

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
jbelkins: I refactored the logic inside the `while` loop to a new type that can be tested separately.  It is dependency-injected by method parameter to allow a mock to be used for testing.

ganeshnj (changes requested):

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
epau, line 15: Can we mark this as `final`?

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
```diff
@@ -0,0 +1,85 @@
...
+        // Update attempt number; the first request is 1.
+        attempt += 1
+
+        // The first time this method is called, set startDate
+        // which is used to determine total elapsed time.
+        if attempt == 1 {
+            startDate = now()
+        }
+
+        // Calculate & set the delay using the formula in the Smithy retry strategy:
```
epau, line 68: Can we add a link to the smithy retry strategy here?

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
```diff
@@ -0,0 +1,85 @@
...
+    let maxWaitTime: TimeInterval
+
+    /// Returns the current Date.  For testing, this closure may be replaced to provide simulated time.
+    var now: () -> Date = { Date() }
+
+    /// The Date when the next request should take place after a retry.
+    var nextRequestDate = Date.distantPast
+    
+    /// The number of requests ("attempts") that have been made by this waiter.
+    private(set) var attempt = 0
```
epau, line 30: Nit: Can we name this `attempts` to stay consistent with the naming elsewhere in this PR?

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
```diff
@@ -0,0 +1,112 @@
...
+        input: Input,
+        config: WaiterConfiguration<Input, Output>,
+        operation: (Input) async throws -> Output
+    ) async throws -> WaiterOutcome<Output>? {
+        try await closure(scheduler, input, config, operation)
+    }
+}
+
+/// A type which tries the waited operation after a delay, then returns or throws the result.
+class WaiterRetryer<Input, Output>: WaiterRetryerInterface {
```
epau, line 56: Does this need to be a class?

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
```diff
@@ -0,0 +1,112 @@
...
+        scheduler: WaiterScheduler,
+        input: RetryerInput,
+        config: WaiterConfiguration<RetryerInput, RetryerOutput>,
+        operation: (RetryerInput) async throws -> RetryerOutput
+    ) async throws -> WaiterOutcome<RetryerOutput>?
+}
+
+/// Container used to type-erase the retryer which implements `WaiterRetryerInterface`
+/// Used to permit injection of a `WaiterRetryerMock` when testing `Waiter`.
+class WaiterRetryerContainer<Input, Output>: WaiterRetryerInterface {
```
epau, line 29: Can we remove this class in favor of reworking `WaiterRetryer` to be more like this container aka more composable? This would prevent our testability things from leaking into our implementation. Additionally, it'll make things a bit simpler by no longer needing two constructs with similar names and purposes (WaiterRetryer + WaiterRetryerContainer) and we no longer need the protocol either.

```swift
struct WaiterRetryer<Input, Output> {
    typealias Operation = (Input) async throws -> Output
    typealias WaitThenRequest = (WaiterScheduler, Input, WaiterConfiguration<Input, Output>, Operation) async throws -> WaiterOutcome<Output>?
    
    let waitThenRequest: WaitThenRequest
    
    // Allow for folks to inject in their own implementation for `waitThenRequest` but default to our "live" implementation is nothing is provided
    init(waitThenRequest: @escaping WaitThenRequest = Self.smithy(scheduler:input:config:operation:)) {
        self.waitThenRequest = waitThenRequest
    }
    
    private static func smithy(
        scheduler: WaiterScheduler,
        input: Input,
        config: WaiterConfiguration<Input, Output>,
        operation: Operation
    ) async throws -> WaiterOutcome<Output>? {
        // Find the required delay from the scheduler, and wait if required.
        if scheduler.currentDelay > 0.0 {
            try await Task.sleep(nanoseconds: UInt64(scheduler.currentDelay * 1_000_000_000.0))
        }

        // Try the operation, collect the result & update the scheduler (for test use only) for this attempt.
        let result = await Result<Output, Error>(catching: { try await operation(input) })
        scheduler.updateAfterRetry()

        // Test the acceptors, to see if one matches.  Take the first match if there is one.
        // If a match is found, take action as required.
        if let match = config.acceptors.compactMap({ $0.evaluate(input: input, result: result) }).first {
            switch match {
            case .success(let lastResult):
                return WaiterOutcome(attempts: scheduler.attempt, result: lastResult)
            case .failure(let lastResult):
                throw WaiterFailureError<Output>(attempts: scheduler.attempt, failedOnMatch: true, result: lastResult)
            case .retry:
                return nil  // Returning nil causes retry, if time remains
            }
        // If no matching acceptor is found, fail if the result was an error.
        } else if case .failure(let error) = result {
            throw WaiterFailureError<Output>(
                attempts: scheduler.attempt,
                failedOnMatch: false,
                result: .failure(error)
            )
        }
        // Since no acceptor matched and the response was not an unhandled error,
        // return nil to cause retry, if time remains
        return nil
    }
}
```

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,79 @@
...
+    public let operation: (Input) async throws -> Output
+
+    /// Creates a `waiter` object with the supplied config and operation.
+    /// - Parameters:
+    ///   - config: An instance of `WaiterConfiguration` that defines the default behavior of this waiter.
+    ///   - operation: A closure that is called one or more times to perform the waiting operation;
+    ///   takes an `Input` as its sole param & returns an `Output` asynchronously.
+    ///   The `operation` closure throws an error if the operation cannot be performed or the
+    ///   operation completes with an error.
+    public init(
```
epau, line 29: For better DI, the Waiter shouldn't know how to create schedulers nor retryers, can we inject those details in? This would remove the need for duplicate `waitUntil` functions and would make `Waiter` a bit easier to test. If we don't want to expose this publicly, we can keep it internal.

```swift 
public init(
    config: WaiterConfiguration<Input, Output>,
    operation: @escaping (Input) async throws -> Output
) {
    self.config = config
    self.operation = operation
    self.init(
        config: config,
        operation: operation,
        scheduler: { options in
            let minDelay = options.minDelay ?? config.minDelay
            let maxDelay = options.maxDelay ?? config.maxDelay
            let maxWaitTime = options.maxWaitTime
            let scheduler = WaiterScheduler(minDelay: minDelay, maxDelay: maxDelay, maxWaitTime: maxWaitTime)
        },
        retryer: { WaiterRetryer<Input, Output>() }
    )
}

init(
    config: WaiterConfiguration<Input, Output>,
    operation: @escaping (Input) async throws -> Output,
    scheduler: (WaiterOptions) -> WaiterScheduler, 
    retryer: () -> WaiterRetryer<Input, Output>
) {
    self.config = config
    self.operation = operation
    self.scheduler = scheduler
    self.retryer = retryer
}
```

Packages/ClientRuntime/Sources/Waiters/WaiterConfiguration.swift
```diff
@@ -0,0 +1,43 @@
...
+        acceptors: [Acceptor]
+    ) throws {
+        guard acceptors.filter({ $0.state == .success }).count >= 1 else {
+            let localizedDescription = "There must be at least one Acceptor with a success state"
+            throw WaiterConfigurationError(localizedDescription: localizedDescription)
+        }
+        self.minDelay = minDelay ?? 2.0
+        self.maxDelay = maxDelay ?? 120.0
+        self.acceptors = acceptors
+    }
```
epau, line 42: Can we move the defaults to the initializer signature? This makes the initializer signature more accurately reflect the behavior since minDelay and maxDelay can never be nil. 

```suggestion
    public init(
        acceptors: [Acceptor],
        minDelay: TimeInterval = 2.0,
        maxDelay: TimeInterval = 120.0
    ) throws {
        guard acceptors.filter({ $0.state == .success }).count >= 1 else {
            let localizedDescription = "There must be at least one Acceptor with a success state"
            throw WaiterConfigurationError(localizedDescription: localizedDescription)
        }
        self.minDelay = minDelay
        self.maxDelay = maxDelay 
        self.acceptors = acceptors
    }
```

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
epau: Does this class need to be thread safe?

epau (changes requested):
Looking good!!! A few questions and suggestions

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
jbelkins: This violates my own conventions too... but I used `attempt` because that is the exact name of the variable in the pseudocode Smithy gives for the retry logic.  I'll change it to `attempts` because I don't think using a different name from the Smithy sample pseudocode will cause any confusion.

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
jbelkins: It's in the doc comment for the class, but I copied it here as well.

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
jbelkins: Going to look at using factory pattern to create new schedulers & retryers, since a new retry & a new scheduler should be made for every request.

Packages/ClientRuntime/Sources/Waiters/WaiterConfiguration.swift
jbelkins: Changed per your suggestion

Packages/ClientRuntime/Sources/Waiters/WaiterScheduler.swift
epau: ahh i see, yeah i agree

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
epau: Factory pattern works too, although it can be a bit heavy handed. This is a pseudo factory pattern tbh. Using closures should allow for new instances of scheduler and retryer to be created for each request

Packages/ClientRuntime/Sources/Waiters/WaiterErrors.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    /// The result of the final (and necessarily successful) operation performed while waiting.
+    public let result: Result<Output, Error>
+}
+
+/// Thrown when a waiter timeouts.
+public struct WaiterTimeoutError: Error {
+    /// The number of attempts at the operation that were made while waiting.
+    public let attempts: Int
+}
```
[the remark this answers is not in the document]
jbelkins: @ganeshnj What do you mean by "keep tidy"?  I typically disfavor enums as errors for a couple reasons:
1) adding a new error case to an enum is a breaking change if the error type is public, but I can add a new struct anytime
2) enum associated values aren't named, while struct properties are, which makes it clearer what the values associated with the error are & what they mean.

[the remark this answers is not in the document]
jbelkins:
```
import Foundation

enum A: Error {
    case a(Int)
}

struct B: Error {
    let attempts: Int
}

do {
//    throw A.a(1)
// or
    throw B(attempts: 1)
} catch A.a(let attempts) {
    print("attempts: \(attempts)")
} catch let error as B {
    print("attempts: \(error.attempts)")
}
```
If you mean the error handling code is more tidy, I think these are both acceptable, with the added bonus that when you catch B the value is labeled as "attempts".

ganeshnj (approved):

Packages/.swiftlint.yml
```diff
@@ -20,6 +20,8 @@ disabled_rules:
   - file_length
   - syntactic_sugar
   - unused_capture_list
+  - nesting
+  - large_tuple
```
epau, line 24: Can we add these to our other .swiftlint.yml configs as well for consistency? 
(we can add some automation around this in the future)

Packages/ClientRuntime/Sources/Waiters/WaiterErrors.swift
```diff
@@ -0,0 +1,30 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Thrown when a `WaiterConfiguration` is constructed with invalid parameters.
+struct WaiterConfigurationError: Error {
```
epau, line 9: Should we make this public since the WaiterConfiguration initializer, that throws this error, is public? That way consumers could catch on the specific error if they wish

Packages/ClientRuntime/Sources/Waiters/WaiterOptions.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    /// Creates a new set of `WaiterOptions`
+    /// - Parameters:
+    ///   - minDelay: The minimum delay between retries while waiting.  If `nil`, the default for the `Waiter` is used.
+    ///   - maxDelay: The maximum delay between retries while waiting.  If `nil`, the default for the `Waiter` is used.
+    ///   - maxWaitTime: The maximum time to spend waiting for the retry to succeed, before a timeout error is thrown.
+    public init(
+        minDelay: TimeInterval? = nil,
+        maxDelay: TimeInterval? = nil,
+        maxWaitTime: TimeInterval
```
epau, line 31: Nit: Its preferable to have parameters with defaults be defined after parameters without defaults

```suggestion
        maxWaitTime: TimeInterval,
        minDelay: TimeInterval? = nil,
        maxDelay: TimeInterval? = nil
```

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
```diff
@@ -0,0 +1,88 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A type which tries the waited operation after a delay, then returns or throws the result.
+/// Called by the `Waiter` to make the initial request, then called again for an additional request when retry
+/// is needed.
+/// A custom closure may be injected at initialization to allow `WaiterRetryer` to be used as a mock.
+class WaiterRetryer<Input, Output> {
```
epau, line 14: Same as a comment I had left previously, does this need to be a class?

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
```diff
@@ -0,0 +1,88 @@
...
+    typealias WaitThenRequest =
+        (WaiterScheduler, Input, WaiterConfiguration<Input, Output>, (Input) async throws -> Output)
+            async throws -> WaiterOutcome<Output>?
+
+    private let closure: WaitThenRequest
```
epau, line 19: Nit: prefer a descriptive name over `closure`. 
For example, we wouldn't have name a property `string` or `bool` etc

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
```diff
@@ -0,0 +1,90 @@
...
+    /// `WaiterTimeoutError` if the waiter times out.
+    @discardableResult
+    public func waitUntil(
+        options: WaiterOptions,
+        input: Input
+    ) async throws -> WaiterOutcome<Output> {
+        let minDelay = options.minDelay ?? config.minDelay
+        let maxDelay = options.maxDelay ?? config.maxDelay
+        let maxWaitTime = options.maxWaitTime
+        let scheduler = WaiterScheduler(minDelay: minDelay, maxDelay: maxDelay, maxWaitTime: maxWaitTime)
```
epau, line 77: Same as a comment I had left previously, can we inject the scheduler for better separation and to make this more testable? If want a new instance of scheduler for every call to `waitUntil` then we can inject a closure.

epau (changes requested):
Overall looking great!!!

I think we're very close to the finish line, just a few comments that remain unaddressed

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
epau: Resolving this, left a new comment for this.

Packages/ClientRuntime/Sources/Waiters/WaiterErrors.swift
jbelkins: Yes, this should definitely be public

Packages/.swiftlint.yml
jbelkins: I will open PRs for these.

As an aside, I wonder if there is a way to separate the rules part of SwiftLint config (which we could share across projects) from the project-specific part of config (ignored files, etc.)

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
jbelkins: I'm not too concerned because this is not a public interface, but it is renamed to `waitThenRequest` like the function it wraps.

Packages/ClientRuntime/Sources/Waiters/WaiterRetryer.swift
jbelkins: It does not anymore.  Changed to `struct`.

Packages/ClientRuntime/Sources/Waiters/WaiterOptions.swift
jbelkins: Accepted

Packages/ClientRuntime/Sources/Waiters/Waiter.swift
jbelkins: Done.  A closure now creates `WaiterScheduler` for `Waiter` when needed.

epau (approved):

PR #451, opened 2022-10-19
fix: add explicit conformance for Equatable and Hashable on Header, Comparable on URLQueryItem

ganeshnj:
## Description of changes

Change required for https://github.com/awslabs/aws-sdk-swift/pull/651

## Scope

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/Http/Headers.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/Headers.swift
@@ -162,6 +162,11 @@ public struct Headers: Hashable {
 }
 
 extension Headers: Equatable {
+    /// Returns a boolean value indicating whether two values are equal irrespective of order.
+    /// - Parameters:
+    ///   - lhs: The first `Headers` to compare.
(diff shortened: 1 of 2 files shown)


Packages/ClientRuntime/Sources/Networking/Http/Headers.swift
```diff
@@ -161,6 +161,12 @@ public struct Headers: Equatable, Hashable {
     }
 }
 
+extension Headers: Equatable {
+    public static func == (lhs: Headers, rhs: Headers) -> Bool {
```
jbelkins, line 165: This and the other public operator overloads should have doc comments to explain their behavior (i.e. insensitive to the order of the headers array)

jbelkins (changes requested):
Code is good, I just think some documentation comments are warranted

Packages/ClientRuntime/Sources/Networking/Http/Headers.swift
ganeshnj: for sure. fixed.

jbelkins (approved):

Issue #453, opened 2022-10-25
Allow for optional millisecond precision when SerDe timestamps of date-time and http-date formats


epau:
### Describe the feature

According to the [Smithy docs](https://awslabs.github.io/smithy/2.0/spec/protocol-traits.html#timestampformat-trait) both the `date-time` and the `http-date` formats must support optional millisecond precision. 

### Use Case

To adhere to the Smithy spec

### Proposed Solution

The deserialization path is straightforward. We can first attempt to deserialize without millisecond precision, if that fails then attempt to deserialize with millisecond precision. 

The serialization case is not as straightforward as we'll need to discuss the best route. 
Either there is some way for a service to specify which variant they require or we expose a lever for consumers to specify the variant.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

Issue #452, opened 2022-10-25
Generating SerDe code for some timestamps doesn't use the correct timestamp format


epau:
### Describe the bug

The generated SerDe code for some timestamps unexpectedly fails because it uses the incorrect timestamp format

### Expected Behavior

The generated SerDe code runs successfully

### Current Behavior

The generated SerDe code fails throwing an exception such as `Expected to decode Double but found a string/data instead` when attempting to encode or decode a timestamp. https://github.com/awslabs/aws-sdk-swift/issues/632

### Reproduction Steps

N/A

### Possible Solution

It appears that there are some code paths for generating SerDe code that do not take into account the member's target. Therefore, the generated code uses the incorrect timestamp format.

The fix is to update these codepaths to the following logic:
1. Check the traits of the member for a timestamp format
2. Check the traits of the member's target for a timestamp format
3. Fallback to a default timestamp format

### AWS SWIFT SDK version used

0.3.1

### Compiler and Version used

14.0.1

### Operating System and version

MacOS 12.6

PR #454, opened 2022-10-28
fix: Timestamp encoding and decoding

epau:
## Issue \#

#453 
#452 
https://github.com/awslabs/aws-sdk-swift/issues/632

## Description of changes
After [this issue](https://github.com/awslabs/aws-sdk-swift/issues/632) was filed for a data decoding issue when calling `AWSLocation.batchUpdateDevicePosition`, it was found that there were more general issues with our decoding and encoding logic of timestamps.
- Timestamps didn't support fractional seconds
- Timestamps weren't using the correct format in some cases (such as when they inherit the format from a target)

This PR fixes both of those issues and does the following:
- Includes fractional seconds when encoding timestamps for all formats (if the date value has non-zero values for fractional seconds).
- Allows services to send timestamps with optional fractional seconds, previously the timestamp would've failed decoding if a service sent timestamps with fractional seconds.
- Checks the member's target for a timestamp format if the member doesn't define one.

## Scope


Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
```diff
@@ -0,0 +1,236 @@
...
+    /// Date time as defined by the date-time production in RFC3339 section 5.6 with no UTC offset and optional fractional precision (for example, 1985-04-12T23:20:50.52Z).
+    case dateTime
+    
+    /// An HTTP date as defined by the IMF-fixdate production in RFC 7231#section-7.1.1.1 (for example, Tue, 29 Apr 2014 18:30:38 GMT) with optional fractional seconds (for example, Sun, 02 Jan 2000 20:34:56.000 GMT).
+    case httpDate
+}
+
+// MARK: - Encoding Helpers
+
+struct TimestampEncodable: Encodable {
```
epau, line 23: Will add tests and docs for this. Wanted to get this up quickly to verify it fixes issues running swift 5.5.3 (confirmed locally)

epau: tests and docs have been added

ganeshnj:
Use `./gradlew ktlintFormat` to fix Kotlin linter issues.

Packages/ClientRuntime/Tests/SerializationTests/SerializationUtilsTests/TimestampSerdeUtilsTests.swift
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }
+        
+        let subjects: [(TimestampFormat, String, Date)] = [
+            (.epochSeconds, "{\"timestamp\":\(testDateWithFractionalSeconds.timeIntervalSince1970)}", testDateWithFractionalSeconds),
+            (.epochSeconds, "{\"timestamp\":\(testDateWithoutFractionalSeconds.timeIntervalSince1970)}", testDateWithoutFractionalSeconds),
+            (.dateTime, "{\"timestamp\":\"1991-05-04T10:12:10.123Z\"}", testDateWithFractionalSeconds),
+            (.dateTime, "{\"timestamp\":\"1991-05-04T10:12:10Z\"}", testDateWithoutFractionalSeconds),
+            (.httpDate, "{\"timestamp\":\"Sat, 04 May 1991 10:12:10.123 GMT\"}", testDateWithFractionalSeconds),
+            (.httpDate, "{\"timestamp\":\"Sat, 04 May 1991 10:12:10 GMT\"}", testDateWithoutFractionalSeconds)
```
ganeshnj, line 114: should there be a test for XML Encode and Decoder?

smithy-swift-codegen/src/test/kotlin/StructDecodeGenerationTests.kt
```diff
@@ -183,36 +184,24 @@ extension TimestampInputOutputResponseBody: Swift.Decodable {
...
 
     public init (from decoder: Swift.Decoder) throws {
         let containerValues = try decoder.container(keyedBy: CodingKeys.self)
-        let normalDateString = try containerValues.decodeIfPresent(Swift.String.self, forKey: .normal)
-        var normalDecoded: ClientRuntime.Date? = nil
-        if let normalDateString = normalDateString {
-            let normalFormatter = ClientRuntime.DateFormatter.iso8601DateFormatterWithoutFractionalSeconds
-            normalDecoded = normalFormatter.date(from: normalDateString)
-        }
+        let normalDecoded = try containerValues.decodeTimestampIfPresent(.dateTime, forKey: .normal)
```
ganeshnj, line 195: very nice.

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
```diff
@@ -0,0 +1,274 @@
...
+        case .epochSeconds:
+            try container.encode(date.timeIntervalSince1970)
+        default:
+            let dateString = TimestampFormatter(format: format).string(from: date)
+            try container.encode(dateString)
+        }
+    }
+}
+
+extension KeyedEncodingContainer {
```
ganeshnj, line 113: how does  the encoding decoding works for the list types used in header and query params?

ganeshnj (changes requested):

ganeshnj:
Overall this PR is on right track and simplify codegen to a good degree.

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
```diff
@@ -0,0 +1,274 @@
...
+    case dateTime
+    
+    /// An HTTP date as defined by the IMF-fixdate production in RFC 7231#section-7.1.1.1 (for example, Tue, 29 Apr 2014 18:30:38 GMT) with optional fractional seconds (for example, Sun, 02 Jan 2000 20:34:56.000 GMT).
+    case httpDate
+}
+
+/// A formatter that converts between dates and their smithy timestamp string representations.
+public struct TimestampFormatter {
+    /// The timestamp serialization format
+    public var format: TimestampFormat
```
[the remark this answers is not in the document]
epau: With the current implementation it doesn't **need** to be but imo a var is preferable to provide a better and more convenient API.  Since it's a struct, mutating the value doesn't have any downsides. Additionally, it stays consistent with the [DateFormatter](https://developer.apple.com/documentation/foundation/dateformatter) APIs such as `dateFormat` and `timeStyle`.

All that being said, I'm not opposed to making it a constant, so lmk.

Packages/ClientRuntime/Tests/SerializationTests/SerializationUtilsTests/TimestampSerdeUtilsTests.swift
epau: This test should be decoder agnostic. The purpose of this test is to test the `decodeTimestamp(,forKey:)` function. Unfortunately, we need to use some decoder to do that, in this case its the JSON decoder. But really, the specific decoder shouldn't matter here.

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
epau: I will investigate this today to ensure the only change here is that it might contain fractional seconds (if the date includes non-zero fractional seconds).

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/DateFormatters.swift
```diff
@@ -10,56 +10,66 @@ import struct Foundation.Locale
...
+    /// For more information, see the _Working With Fixed Format Date Representations_ section in:
+    ///  https://developer.apple.com/documentation/foundation/dateformatter
+    private convenience init(fixedFormat dateFormat: String) {
+        self.init()
+        self.dateFormat = dateFormat
+        self.locale = Locale(identifier: "en_US_POSIX")
+        self.timeZone = TimeZone(secondsFromGMT: 0)!
+    }
+    
+    private static func fixedFormatFormatter(dateFormat: String) -> DateFormatter {
```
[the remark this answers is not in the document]
epau: actually, this function isn't used and can be removed

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
epau: These cases are covered by protocol tests and are passing

Packages/ClientRuntime/Sources/Serialization/SerializationUtils/TimestampSerdeUtils.swift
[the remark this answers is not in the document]
ganeshnj: I always prefer immutability over mutability as default. `DateFormatter` that you linked doesn't seem Swifty enough rather look copy-paste from obj-c.

You can always go to `var` in future but can't to `let`. I would still prefer it a let.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseHeaders.kt
```diff
@@ -77,20 +78,7 @@ class HttpResponseHeaders(
...
-                        writer.dedent()
-                        writer.write("} else {")
-                        writer.indent()
-                        writer.write("throw \$N.deserializationFailed(HeaderDeserializationError.invalidTimestampHeader(value: \$LHeaderValue))", ClientRuntimeTypes.Core.ClientError, memberName)
-                        writer.dedent()
-                        writer.write("}")
-                    } else {
-                        writer.write("self.\$L = $memberValue", memberName)
-                    }
+                    writer.write("self.\$L = $memberValue", memberName)
```
ganeshnj, line 81:
```suggestion
                    writer.write("self.\$L =  \$L", memberName, $memberValue)
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/formurl/MemberShapeEncodeFormURLGenerator.kt
```diff
@@ -88,9 +89,14 @@ abstract class MemberShapeEncodeFormURLGenerator(
...
-                    val encodeValue = "TimestampWrapper($nestedMemberTargetName, format: .$format), forKey: ${ClientRuntimeTypes.Serde.Key}(\"${nestedMemberResolvedName}\")"
-                    writer.write("try $containerName.encode($encodeValue)")
+                    val codingKey = "${ClientRuntimeTypes.Serde.Key}(\"${nestedMemberResolvedName}\")"
+                    val code = TimestampEncodeGenerator(
+                        containerName,
+                        nestedMemberTargetName,
+                        codingKey,
+                        TimestampHelpers.getTimestampFormat(nestedMember, nestedMemberTarget, defaultTimestampFormat)
+                    ).generate()
+                    writer.write(code)
```
ganeshnj, line 99: this is fine here but we should prefer using formatter over generating some string and dumping to writer.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -37,25 +38,16 @@ abstract class MemberShapeDecodeGenerator(
...
-        }
+        val timestampFormat = TimestampHelpers.getTimestampFormat(member, target, defaultTimestampFormat)
+        val decodingCode = TimestampDecodeGenerator(
+            containerName,
+            ".$memberName",
+            timestampFormat,
+            true
+        ).generate()
+        val decodedMemberName = "${memberName}Decoded"
+        writer.write("let $decodedMemberName = $decodingCode")
```
ganeshnj, line 49:
```suggestion
        val decodedMemberName = "${memberName}Decoded"
        writer.write("let \$L = \$L, decodedMemberName, decodingCode")
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -203,12 +173,9 @@ abstract class MemberShapeDecodeGenerator(
...
                         }
                     } else { // decode date as a string manually
-                        val formatterName = "${iteratorName}Formatter"
-                        writeDateFormatter(formatterName, tsFormat, writer)
                         val dateName = "date$level"
-                        writer.openBlock("guard let $dateName = $formatterName.date(from: $iteratorName) else {", "}") {
-                            renderDecodingDateError(topLevelMember)
-                        }
+                        val swiftTimestampName = TimestampHelpers.generateTimestampFormatEnumValue(timestampFormat)
+                        writer.write("let $dateName = try containerValues.timestampStringAsDate($iteratorName, format: .$swiftTimestampName, forKey: .${topLevelMember.memberName})")
```
ganeshnj, line 178:
```suggestion
                        writer.write("let \$L = try containerValues.timestampStringAsDate(\$L, format: .\$L, forKey: \$L)", dateName, iteratorName, swiftTimestampName, topLevelMember.memberName)
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeDecodeGenerator.kt
```diff
@@ -318,12 +282,9 @@ abstract class MemberShapeDecodeGenerator(
...
                         }
                     } else { // decode date as a string manually
-                        val formatterName = "${mapName}Formatter"
-                        writeDateFormatter(formatterName, tsFormat, writer)
                         val dateName = "date$level"
-                        writer.openBlock("guard let $dateName = $formatterName.date(from: $valueIterator) else {", "}") {
-                            renderDecodingDateError(topLevelMember)
-                        }
+                        val swiftTimestampName = TimestampHelpers.generateTimestampFormatEnumValue(timestampFormat)
+                        writer.write("let $dateName = try containerValues.timestampStringAsDate($valueIterator, format: .$swiftTimestampName, forKey: .${topLevelMember.memberName})")
```
ganeshnj, line 287:
```suggestion
                        writer.write("let \$L = try containerValues.timestampStringAsDate(\$L, format: \$L, forKey: \$L)", dateName, valueIterator, swiftTimestampName, topLevelMember.memberName)
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -118,12 +110,43 @@ abstract class MemberShapeEncodeGenerator(
...
+            else -> {
+                val extension = getShapeExtension(memberShape, memberName, isBoxed)
+                if (codingKey != null) {
+                    "try $containerName.encode($extension, forKey: $codingKey)"
+                } else {
+                    "try $containerName.encode($extension)"
+                }
+            }
+        }
+        writer.write(code)
```
ganeshnj, line 147: prefer code writer
```suggestion
                    writer.write("try \$L.encode(\$L, forKey: \$L)", containerName, extension, codingKey)
                } else {
                    writer.write("try \$L.encode(\$L)", containerName, extension)
                }
            }
        }
        
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -224,30 +229,29 @@ abstract class MemberShapeDecodeXMLGenerator(
...
-        writer.write("let $decodedMemberName = try $containerName.$decodeVerb(\$N.self, forKey: .$memberName)", SwiftTypes.String)
-
-        val memberBuffer = "${memberName}Buffer"
-        val format = determineTimestampFormat(member, memberTarget, defaultTimestampFormat)
-        writer.write("var $memberBuffer:\$T = nil", memberTargetSymbol)
-        writer.openBlock("if let $decodedMemberName = $decodedMemberName {", "}") {
-            writer.write("$memberBuffer = try \$N.parseDateStringValue($decodedMemberName, format: .$format)", ClientRuntimeTypes.Serde.TimestampWrapperDecoder)
-        }
-        renderAssigningDecodedMember(memberName, memberBuffer)
+        writer.write("let $decodedMemberName = $decodingCode")
```
ganeshnj, line 253:
```suggestion
        writer.write("let \$LDecoded = \$L", memberName, decodingCode)
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeEncodeXMLGenerator.kt
```diff
@@ -81,9 +82,14 @@ abstract class MemberShapeEncodeXMLGenerator(
...
-                    val encodeValue = "${ClientRuntimeTypes.Serde.TimestampWrapper}($nestedMemberTargetName, format: .$format), forKey: ${ClientRuntimeTypes.Serde.Key}(\"${nestedMemberResolvedName}\")"
-                    writer.write("try $containerName.encode($encodeValue)")
+                    val codingKey = "${ClientRuntimeTypes.Serde.Key}(\"${nestedMemberResolvedName}\")"
+                    val code = TimestampEncodeGenerator(
+                        containerName,
+                        nestedMemberTargetName,
+                        codingKey,
+                        TimestampHelpers.getTimestampFormat(nestedMember, nestedMemberTarget, defaultTimestampFormat)
+                    ).generate()
+                    writer.write(code)
```
ganeshnj, line 92: could you make use of formatters here?

also applies to rest of the similar code snippets, potentially the generate method takes the writer as param and uses formatter inside it.

ganeshnj:
We must leverage CodeWriter features to make our life easier.

https://awslabs.github.io/smithy/javadoc/1.6.1/software/amazon/smithy/utils/CodeWriter.html

Smithy has some prebuilt formatters which are great and for our convenience we have added our [own](https://github.com/awslabs/smithy-swift/blob/1380e277e4bd8c594db365d7d37388fff5b82752/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt#L81) too.

ganeshnj (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/ProtocolGenerator.kt
```diff
@@ -42,21 +42,11 @@ interface ProtocolGenerator {
...
-            return when (tsFormat) {
-                TimestampFormatTrait.Format.EPOCH_SECONDS -> "${memberName}$terminator.timeIntervalSince1970$epochTerminator"
-                // FIXME return to this to figure out when to use fractional seconds precision in more general sense after we switch
-                // to custom date type
-                TimestampFormatTrait.Format.DATE_TIME -> "${memberName}$terminator.iso8601WithoutFractionalSeconds()$stringDateTerminator"
-                TimestampFormatTrait.Format.HTTP_DATE -> "${memberName}$terminator.rfc5322()$stringDateTerminator"
-                else -> throw CodegenException("unknown timestamp format: $tsFormat")
-            }
+            val timestampFormat = TimestampHelpers.generateTimestampFormatEnumValue(tsFormat)
+            return "TimestampFormatter(format: .$timestampFormat).string(from: $memberName)$stringDateTerminator"
```
epau, line 49: I don't have an obvious solution to convert this snippet to use CodeWriter for formatting. 
I'd like to keep this as is for now since it simply improves the existing implementation but lmk if you think this should be a blocker

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -118,8 +110,38 @@ abstract class MemberShapeEncodeGenerator(
...
+                    TimestampHelpers.getTimestampFormat(memberShape, targetShape, defaultTimestampFormat)
+                ).generate(writer)
+            }
+            else -> {
+                val extension = getShapeExtension(memberShape, memberName, isBoxed)
+                if (codingKey != null) {
+                    writer.write("try \$L.encode(\$L, forKey: \$L)", containerName, extension, codingKey)
+                } else {
+                    writer.write("try \$L.encode(\$L)", containerName, extension)
+                }
```
ganeshnj, line 144: trivial: you could use inline method like we talked on slack.

ganeshnj (approved):

PR #457, opened 2022-11-02
fix: Provide an encoding customization that allows empty lists to be serialized

epau:
This PR provides a form url encoding customization option that allows for empty lists to be serialized.
Previously, empty lists would never be serialized.

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/665

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/formurl/FormURLEncodeCustomizable.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/formurl/FormURLEncodeCustomizable.kt
@@ -10,4 +10,5 @@ import software.amazon.smithy.model.shapes.Shape
 interface FormURLEncodeCustomizable {
     fun alwaysUsesFlattenedCollections(): Boolean
     fun customNameTraitGenerator(memberShape: Shape, defaultName: String): String
+    fun shouldSerializeEmptyLists(): Boolean
 }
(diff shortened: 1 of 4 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/formurl/FormURLEncodeCustomizable.kt
```diff
@@ -10,4 +10,5 @@ import software.amazon.smithy.model.shapes.Shape
 interface FormURLEncodeCustomizable {
     fun alwaysUsesFlattenedCollections(): Boolean
     fun customNameTraitGenerator(memberShape: Shape, defaultName: String): String
+    fun shouldSerializeEmptyLists(): Boolean
```
ganeshnj, line 13: You could add default implementation here
```
    fun shouldSerializeEmptyLists(): Boolean {
        return true
    }
```
as in all cases it is `true`. ie for EC2Query and AWSQuery

ganeshnj (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/formurl/FormURLEncodeCustomizable.kt
epau: Will do if that's preferable

PR #461, opened 2022-11-03
Jangirg/fix/type casing

ganeshnj:
## Issue \#

Fixes https://github.com/awslabs/aws-sdk-swift/issues/648

## Scope


jbelkins (changes requested):
Sadly, our Swift names can only be as consistent as the service names, which seem to be pretty arbitrarily set.

Approved though, because this will fix some of our naming issues if not all.

jbelkins (approved):

ganeshnj:
> Sadly, our Swift names can only be as consistent as the service names, which seem to be pretty arbitrarily set.
> 
> Approved though, because this will fix some of our naming issues if not all.

That's just how codegen works, there are ways to to smarter things but comes with maintenance cost which I would rather not pay unless absolutely required.

PR #465, opened 2022-11-04
fix: add content-length middleware for operation that has codable conformance (AWS Query)

ganeshnj:
## Issue \#

Fixes https://github.com/awslabs/aws-sdk-swift/issues/619

## Description of changes

Current SDK code checks whether there is a member in the request shape which doesn't target header, query, if yes add the Content-Length middleware because the member will go to body.

For operation like `GetCallerIdentity` in `STS`, it doesn't have any members defined which means there is no Content-Length middleware for it. However, per AWS Query protocol, during serialization `Action` and `Version` is always added to body.
 ```
Action=GetCallerIdentity&Version=2011-06-15
```

With this change, it makes sure if we have a Codable conformance to add something to body of the request, there must exist a Content-Length middleware.

## Scope


jbelkins (commented):
Should there be tests on this?

ganeshnj:
> Should there be tests on this?

Fair call, nothing broke i.e. there are actually no tests for this. Let me see if I can add some.

ganeshnj:
Test case https://github.com/awslabs/aws-sdk-swift/pull/681/files

jbelkins (approved):

PR #467, opened 2022-11-08
fix!: update shape and member extensions to use same casing as rest of the system

ganeshnj:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/648

### Motivation

Some types such as `SessionKeysAbpV1_0_x` were generated without underscore while underscore is a valid char in types. 

### Changes

- Updated logic to preserve underscore.
- Updated extensions methods for Shape and Member types to use consistent casing

### Result

Underscore containing types are generated with underscore. 

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
```diff
@@ -25,6 +25,8 @@ import software.amazon.smithy.model.traits.RequiredTrait
 import software.amazon.smithy.model.traits.StreamingTrait
 import software.amazon.smithy.model.traits.Trait
 import software.amazon.smithy.swift.codegen.getOrNull
+import software.amazon.smithy.swift.codegen.utils.toLowerCamelCase
+import software.amazon.smithy.swift.codegen.utils.toUpperCamelCase
 import software.amazon.smithy.utils.StringUtils
 import kotlin.streams.toList
```
ganeshnj, line 31: this import is quintessential

`❯ ./gradlew ktlintFormat` removes it and build fails horribly.

jbelkins (approved):

epau (approved):

PR #468, opened 2022-11-08
fix: Throw error if a required query param is nil

epau:
This PR updates our query provider logic to now throw an error if a required query parameter is nil. 

## Scope

- [x] Bug fix (breaking change which fixes an issue)


Packages/ClientRuntime/Sources/Networking/ClientError.swift
```diff
@@ -18,31 +19,24 @@ public enum ClientError: Error, Equatable {
...
-            return lhsAuthString == rhsAuthString
-        case (let .retryError(lhsRetryError),
-              let .retryError(rhsRetryError)):
-            return String(reflecting: lhsRetryError) == String(reflecting: rhsRetryError)
+        case let (.networkError(lhsError), .networkError(rhsError)):
+            return String(reflecting: lhsError) == String(reflecting: rhsError)
+        case let (.pathCreationFailed(lhsError), .pathCreationFailed(rhsError)):
+            return lhsError == rhsError
+        case let (.queryItemCreationFailed(lhsError), .queryItemCreationFailed(rhsError)):
+            return lhsError == rhsError
```
epau, line 27: Just wanted to highlight the meaningful change in this snippet since I improved the format of the other cases while i was here.

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/Providers/QueryItemProvider.swift
```diff
@@ -6,5 +6,5 @@
 //
 
 public protocol QueryItemProvider {
-    var queryItems: [URLQueryItem] { get }
+    func queryItems() throws -> [URLQueryItem]
```
ganeshnj, line 9: this is a breaking change. Also you can use throwing property

```suggestion
    var queryItems: [URLQueryItem] { get throws }
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpQueryItemProvider.kt
```diff
@@ -141,6 +142,15 @@ class HttpQueryItemProvider(
...
             }
+            if (queryBinding.member.isRequired()) {
+                writer.openBlock("else {", "}") {
+                    writer.write(
+                        "let message = \"Creating a URL Query Item failed. \$L is required but it is nil\"",
+                        memberName
+                    )
+                    writer.write("throw ClientRuntime.ClientError.queryItemCreationFailed(message)")
+                }
+            }
```
ganeshnj, line 153: we should use a guard let than if let here to go with Swift style when property is required

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
```diff
@@ -103,6 +103,10 @@ fun MemberShape.needsDefaultValueCheck(model: Model, symbolProvider: SymbolProvi
     return isPrimitiveShape && isNotBoxed && defaultValueNotNull
 }
 
+fun MemberShape.isRequired(): Boolean {
+    return (this.hasTrait<RequiredTrait>())
+}
```
ganeshnj, line 108: can be?
```suggestion
fun Shape.isRequired(): Boolean {
    return (this.hasTrait<RequiredTrait>())
}
```

ganeshnj (changes requested):

ganeshnj:
Fixes https://github.com/awslabs/aws-sdk-swift/issues/686

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
epau: Hmm i don't know if it makes sense to put this on shape, since having the required trait means the shape is a member of some shape

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/Providers/QueryItemProvider.swift
epau: Ahh TIL about `get throws` will update. 
This will still be a breaking change though

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
ganeshnj: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html#required-trait

makes sense, it is member trait only.

Packages/ClientRuntime/Sources/Networking/Http/Middlewares/Providers/QueryItemProvider.swift
ganeshnj: for sure it will be, you may want to update commit message with `!` bang

epau:
I think there is room to improve our codegen here and move much of the logic to native swift. I refrained from doing this in this PR to limit the scope to simply fixing the bug and filed this issue to track the follow-up improvement work https://github.com/awslabs/smithy-swift/issues/471

ganeshnj (approved):

PR #476, opened 2022-11-09
feat: add support for wss request signing

atierian:
## Description of changes

Adds a `wss` case to the `ProtocolType` enum to support `wss` request signing + simple unit test.
Confirmed this works by manual testing through `AWSClientRuntime.AWSSigV4Signer.sigV4SignedRequest(requestBuilder:signingConfig:)`.

## Scope

### DIFF
--- a/Packages/ClientRuntime/Sources/Networking/Http/ProtocolType.swift
+++ b/Packages/ClientRuntime/Sources/Networking/Http/ProtocolType.swift
@@ -8,4 +8,5 @@ import Foundation
 public enum ProtocolType: String, CaseIterable {
     case http
     case https
+    case wss
 }
--- a/Packages/ClientRuntime/Tests/ClientRuntimeTests/NetworkingTests/SdkRequestBuilderTests.swift
+++ b/Packages/ClientRuntime/Tests/ClientRuntimeTests/NetworkingTests/SdkRequestBuilderTests.swift
@@ -21,4 +21,21 @@ class SdkRequestBuilderTests: XCTestCase {
         XCTAssertEqual(pathToMatch, updatedPath)
         XCTAssertEqual(url, updatedRequest.endpoint.url?.absoluteString)
     }
+
+    func testBuildWSSRequest() {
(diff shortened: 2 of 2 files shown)


Packages/ClientRuntime/Sources/Networking/Http/ProtocolType.swift
```diff
@@ -8,4 +8,5 @@ import Foundation
 public enum ProtocolType: String, CaseIterable {
     case http
     case https
+    case wss
```
ganeshnj, line 11: I think this fair to add but this doesn't mean the SDK supports wss as protocol. This will only be useful for cases where custom endpoints are provided.

Endpoints vended by the SDK are either https or http. Also, this change needs to be tested for https://github.com/awslabs/aws-sdk-swift/blob/0.5.0/release/AWSApiGatewayManagementApi/ApiGatewayManagementApiClientProtocol.swift#L9-L13 as this is only usecase comes to mind as of now where customer provided endpoints are used.

IIRC, this is expected to fail because of endpoints 2.0 ruleset not supporting wss and the team is working on it.

Could you update what testing have done for it in PR description?

atierian: > I think this fair to add but this doesn't mean the SDK supports wss as protocol. This will only be useful for cases where custom endpoints are provided.

This is the use case I'm working through currently - I have a dependency on ClientRuntime and need to sign some requests that I'm making directly, but will hopefully be making with the implementation behind the design of https://github.com/awslabs/smithy-swift/pull/460 in the future. At the moment, for my use case, I would like to leverage the signing APIs surfaced by the SDK (`sigV4SignedRequest`) rather than rolling my own implementation or dropping down to CRT. 

After typing this out, I'm starting to think that it might actually not be appropriate, at least not for the above use case, to add it here because it can give the false impression that websocket requests are supported. If there is an applicable use case around custom API Gateway endpoints, then I'm happy to add some tests around that.

I'll update the PR description with details on the testing done wrt signing.

PR #477, opened 2022-11-14
Exposes model property in CodegenVisitor

epau:
https://github.com/awslabs/aws-sdk-swift/pull/709

## Description of changes
This PR exposes the `model` property in `CodegenVisitor` to make the codegen behavior in tests more consistent with that in release.

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
@@ -35,7 +35,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Void>() {
 
     private val LOGGER = Logger.getLogger(javaClass.name)
     private val settings: SwiftSettings = SwiftSettings.from(context.model, context.settings)
-    private val model: Model
+    val model: Model
     private val service: ServiceShape
     private val fileManifest: FileManifest = context.fileManifest
     private val symbolProvider: SymbolProvider


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
```diff
@@ -35,7 +35,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Void>() {
 
     private val LOGGER = Logger.getLogger(javaClass.name)
     private val settings: SwiftSettings = SwiftSettings.from(context.model, context.settings)
-    private val model: Model
+    val model: Model
```
jbelkins, line 38: Does `model` only need to be exposed to facilitate testing?  (That's the only place I see it being accessed in https://github.com/awslabs/aws-sdk-swift/pull/709 )

epau: Currently yes, but I'd argue it makes sense outside of a testing context as well as it now exposes the changes it makes to the model it receives as an input. 

Ideally the CodegenVisitor wouldn't mutate the model and instead there'd be some other mechanism responsible for the mutations, but that drastically increases the scope of this issue.

epau: Is that necessary given it's a `val` and therefore immutable (just like a `let` in swift)?

jbelkins (approved):

Issue #449, opened 2022-10-10
Code-generate a `waitFor` method for each waiter


jbelkins:
For each waiter, code-generate a public `waitFor...()` method that provides the entry point to waiter functionality.

The waiter method should be code generated as described in [the Swift Waiters design doc here](https://quip-amazon.com/dLUUAwMak4ka/Waiters-Design#temp:C:MbWc62ebeae967748948cce6af87).

Requirements:
- Generate the method, configuring reusable types as needed, and using the Acceptors generated in #448.
- Add tests as required to ensure that methods are generated correctly.

When complete, this work should merge to a Waiters project branch while awaiting the completion of the rest of the Waiters project.

Issue #448, opened 2022-10-10
Code-generate Acceptors for waiters


jbelkins:
Every Smithy waiter will include a list of Acceptors, which define the conditions that cause a change in waiter state.

Code-generate a Swift expression creating the Acceptors that are defined in a waiter, and make them available as a static variable that contains an array of Acceptors.

Requirements:
- Code generate all possible forms of Acceptor, including JMESPath expressions
- Create a static variable containing an array of Acceptors for each defined waiter
- Include unit tests ensuring that code is generated correctly for various Smithy definitions

When complete, this work should merge to a Waiters project branch while awaiting the completion of the rest of the Waiters project.

PR #478, opened 2022-11-15
feat: Waiter extension & methods

jbelkins:
## Issue \#
Completes #449 

## Description of changes
Generate an extension for each service client with waiters, containing the public `waitUntil...` methods that consumers can use for waiting.

Includes:
- Extension declaration
- Documentation comments for each waiter
- Method signature
- Method body
- Test to verify proper waiter generation (do I need to break this test down further to test specific parts of the generated code?)

Not included:
- Acceptors.  Acceptors are generated as an empty list.  As a result, the waiter methods will compile without error, but will throw an error for invalid `WaiterConfiguration` when called.  Acceptors are to be auto-generated as part of #448.

## Scope


smithy-swift-codegen/src/test/kotlin/WaiterGeneratorTests.kt
```diff
@@ -0,0 +1,61 @@
...
+import software.amazon.smithy.swift.codegen.SwiftSettings
+import software.amazon.smithy.swift.codegen.WaiterGenerator
+import software.amazon.smithy.swift.codegen.core.CodegenContext
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
+
+class WaiterGeneratorTests {
+
+    @Test
+    fun testRenderTestWaiter() {
```
epau, line 14: Can we add a test that verifies operations without the `waitable` trait, do not render a waiter?

jbelkins: Added

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/WaiterGenerator.kt
```diff
@@ -0,0 +1,88 @@
...
+import software.amazon.smithy.swift.codegen.utils.toUpperCamelCase
+import software.amazon.smithy.waiters.WaitableTrait
+import software.amazon.smithy.waiters.Waiter
+
+/**
+ * Generate waiters for supporting operations.  See
+ * https://smithy.io/2.0/additional-specs/waiters.html for details.
+ */
+class WaiterGenerator : SwiftIntegration {
+    override fun enabledForService(model: Model, settings: SwiftSettings): Boolean =
```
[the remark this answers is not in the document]
jbelkins: Added

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/WaiterGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+                waitedOperations.forEach { waitedOperation ->
+                    val waitableTraits = waitedOperation.allTraits.mapNotNull { (_, trait) -> trait as? WaitableTrait }
+                    waitableTraits.forEach { waitableTrait ->
+                        waitableTrait.waiters.forEach { (waiterName, waiter) ->
+                            renderWaiterForOperation(writer, ctx, service, waitedOperation, waiterName, waiter)
+                        }
+                    }
+                }
+            }
+        }
```
epau, line 49: Nit: Some comments describing what we're doing here would be super helpful

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/WaiterGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+    override fun enabledForService(model: Model, settings: SwiftSettings): Boolean {
+        return serviceHasWaiters(model)
+    }
+
+    override fun writeAdditionalFiles(
+        ctx: CodegenContext,
+        protoCtx: ProtocolGenerator.GenerationContext,
+        delegator: SwiftDelegator
+    ) {
+        if (!serviceHasWaiters(ctx.model)) return
```
epau, line 30: Is this line necessary given it should be covered via the SwiftIntegration conformance `enabledForService`

epau (approved):
Looks great! Thanks for the additional tests! ❤️ 

Left two nit comments but nothing to hold this up

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/WaiterGenerator.kt
jbelkins: This is me being defensive... I know that `enabledForService` gets called first, then only if `true` does this method get run.  But since that logic is outside this class, I wanted this class to enforce that behavior itself.

PR #480, opened 2022-11-15
fix: Properly encode empty bodies for json and xml

epau:
This PR properly encodes empty bodies. 
For JSON -> encode as "{}"
for XML -> encode as ""

## Issue \#
[](https://github.com/awslabs/aws-sdk-swift/issues/580)

## Scope

### DIFF
--- a/smithy-swift-codegen/src/test/kotlin/HttpBodyMiddlewareTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/HttpBodyMiddlewareTests.kt
@@ -149,11 +149,23 @@ class HttpBodyMiddlewareTests {
                 Self.Context == H.Context
                 {
                     do {
+                        let encoder = context.getEncoder()
                         if let payload1 = input.operationInput.payload1 {
-                            let encoder = context.getEncoder()
                             let payload1data = try encoder.encode(payload1)
                             let payload1body = ClientRuntime.HttpBody.data(payload1data)
                             input.builder.withBody(payload1body)
+                        } else {
+                            if encoder is JSONEncoder {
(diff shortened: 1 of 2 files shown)


smithy-swift-codegen/src/test/kotlin/HttpBodyMiddlewareTests.kt
```diff
@@ -155,9 +155,17 @@ class HttpBodyMiddlewareTests {
                             let payload1body = ClientRuntime.HttpBody.data(payload1data)
                             input.builder.withBody(payload1body)
                         } else {
-                            let payload1data = try encoder.encode(input.operationInput)
-                            let payload1body = ClientRuntime.HttpBody.data(payload1data)
-                            input.builder.withBody(payload1body)
```
jbelkins, line 160: What was happening before?  Was an empty body failing to encode?  Or was there actually some use in encoding `operationInput`?

epau: Previously, it would attempt to encode the `operationInput`. The smithy specification does NOT dictate that the input struct be encoded if the http payload member is nil. Therefore, this can produce unexpected behavior.

For JSON, this would be fine (which i assume is why this code existed in the first place), as it would encode as `{}`.
But for XML it would produce an unexpected result, since it would encode the input struct such as `<SomeInput></SomeInput>` This is the problem that the issue filed describes https://github.com/awslabs/aws-sdk-swift/issues/580

jbelkins (approved):
Approved, but I concur that encoding/decoding needs a more fundamental overhaul

PR #483, opened 2022-11-21
feat: Code-generate Acceptors for waiters

jbelkins:
## Issue \#
Completes https://github.com/awslabs/smithy-swift/issues/448
Fixes https://github.com/awslabs/aws-sdk-swift/issues/733

## Description of changes
Code generate all acceptors for each waiter.  Support for JMESPath is not complete to the JMESPath spec, but is sufficient to handle all current AWS acceptors.  (Comparable to the JMESPath support in `smithy-kotlin`.)
- For each waiter, a separate, internal-scoped static method is provided that creates a `WaiterConfig` for that waiter.  This config is used by the `waitUntil` method for that waiter to configure its `Waiter` object.
- The `WaiterConfig` includes `minDelay` and `maxDelay` values, plus an array of `Acceptor`s.
- Code is generated that implement all `Acceptor` types.
- For operations that have an `errorType` acceptor, conformance with the `WaiterCodedError` Swift protocol is added to that operation's error to facilitate error type matching.
- Tests for generated `Acceptor`s are in the SDK project, since there is currently no support for testing generated code in this project.  See https://github.com/awslabs/aws-sdk-swift/pull/732

Changes other than new content:
- Split the Waiters Swift integration from the code generation.
- Reorganized existing waiters-related Kotlin files into a `waiters` subfolder.

## Scope


Packages/ClientRuntime/Sources/Middleware/OrderedGroup.swift
```diff
@@ -82,7 +82,7 @@ public struct OrderedGroup<Input, Output, Context: MiddlewareContext> {
         order.insert(relativeTo: relativeTo, position: position, ids: middleware.id)
     }
     
-    func get(id: String)-> AnyMiddleware<Input, Output, Context>? {
+    func get(id: String) -> AnyMiddleware<Input, Output, Context>? {
```
jbelkins, line 85: SwiftLint fix

Packages/ClientRuntime/Sources/Util/PlatformOperationSystemVersion.swift
```diff
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#if os(iOS) || os (watchOS) || os(macOS) || os(tvOS)
+#if os(iOS) || os(watchOS) || os(macOS) || os(tvOS)
```
jbelkins, line 8: SwiftLint fix

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
```diff
@@ -0,0 +1,87 @@
...
+                    renderInputOutputBlockContents(false, matcher.value)
+                }
+                is Matcher.InputOutputMember -> {
+                    renderInputOutputBlockContents(true, matcher.value)
+                }
+                is Matcher.ErrorTypeMember -> {
+                    val errorTypeName = "${waitedOperation.toUpperCamelCase()}OutputError"
+                    var errorEnumCaseName = "${matcher.value.toLowerCamelCase()}"
+                    writer.write("// Match on error case: $errorEnumCaseName")
+                    writer.write("return false")
```
jbelkins, line 50: The `errorType` matcher is essentially disabled (i.e. it never matches) while I determine the best way to match errors; the approach commented out below is too strict.

`errorType` matching will be implemented in a future PR.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+//                    }
+                }
+            }
+        }
+    }
+
+    private fun renderInputOutputBlockContents(includeInput: Boolean, pathMatcher: PathMatcher) {
+        writer.write("// JMESPath expression: ${pathMatcher.path}")
+        writer.write("// JMESPath comparator: ${pathMatcher.comparator}")
+        writer.write("// JMESPath expected value: ${pathMatcher.expected}")
```
jbelkins, line 66: These comments may be removed later, but for now they are useful in interpreting the generated code.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
```diff
@@ -24,6 +24,18 @@ extension WaiterConfiguration {
             case failure
         }
 
+        /// Used as the root value of an `inputOutput` acceptor, which has the `input` and `output` fields
+        /// as its two top level properties.
+        public struct InputOutput {
+            public let input: Input?
+            public let output: Output?
```
epau, line 31: Do these need to be optional? The initializer requires non-nil values.

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+//
+
+import Foundation
+
+public extension Array {
+
+    /// Flattens an array of arrays down to single depth.
+    /// - Parameter transform: A block that transforms the element(s) of this array into an array containing type T.
+    /// - Returns: A single-depth array of type T.
+    func flattenIfPossible<T>(_ transform: (Element) throws -> Array<T>) rethrows -> Array<T> {
```
epau, line 15: Is this necessary? Why can't we just call `flatMap` instead?

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+        return try flatMap { try transform($0) }
+    }
+
+    /// Returns the receiver mapped to an array of a different, non-array type using the provided transform block.
+    /// - Parameters:
+    ///   - unused: An unused integer parameter which is used to affect the precedence of selection of overloads by the Swift compiler.
+    ///   See discussion in the implementation.
+    ///   - transform: A block that transforms each element of the array into a type that is not itself an Array.
+    /// - Returns: The receiver, mapped by the `transform` block.
+    func flattenIfPossible<T>(_ unused: Int = 0, _ transform: (Element) throws -> T) rethrows -> Array<T> {
```
epau, line 25: Is this necessary? Why can't we just call `map` instead?

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+        // the element of an array is also an array, which is not desired behavior.
+        // See: https://forums.swift.org/t/how-to-specify-which-is-the-default-function-when-functions-use-the-same-name-like-array-reversed/39168/5
+        // In the future, if Swift allows us to manually rank overloads or otherwise fixes this issue,
+        // we should eliminate this param & use the new mechanism for overload resolution instead.
+        return try map { try transform($0) }
+    }
+}
+
+/// A value type that is used to facilitate comparison of values returned by JMESPath types, in accordance with the JMESPath spec.
+public enum JMESValue: Equatable, Comparable {
```
epau, line 39: Can we place this type in it's own file?

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /// Creates a `JMESValue` from a type that is raw-representable by a string.
+    /// Typically this will be an enumeration of a limited set of expected string values.
+    /// The resulting value will be a string for purposes of comparing it to other values.
+    /// - Parameter rr: The `RawRepresentable` to represent as a JMESPath String, or as a JMESPath `null` if `rr` is nil.
+    public init<T: RawRepresentable>(_ rr: T?) where T.RawValue == String {
+        guard let string = rr?.rawValue else { self = .null; return }
+        self = .string(string)
+    }
```
epau, line 84: Can we call the string initializer to simplify things? Additionally, prefer to use a clearly named parameter over abbreviations. 

```suggestion
    public init<T: RawRepresentable>(_ value: T?) where T.RawValue == String {
        self.init(value?.rawValue)
    }
```

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Compares two `JMESValue` values, testing for equality per the JMESPath rules.
+    /// - Parameters:
+    ///   - lhs: The left value
+    ///   - rhs: The right value
+    /// - Returns: The result of the Swift comparison of the associated values if `lhs` and `rhs` are the same JMESPath type, or `false` otherwise.
```
epau, line 90:
```suggestion
    /// - Returns: `true` if the associated values of `lhs` and `rhs` are the same JMESPath type and have the same value, otherwise returns `false`.
```

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
```diff
@@ -0,0 +1,120 @@
...
+            return false
+        }
+    }
+
+    /// Compares two `JMESValue` values, testing for order per the JMESPath rules.
+    /// Comparison fails if both left and right are not JMESPath numbers.
+    /// - Parameters:
+    ///   - lhs: The left value
+    ///   - rhs: The right value
+    /// - Returns: The result of the Swift comparison of the associated values if `lhs` and `rhs` are both JMESPath numbers, or `false` otherwise.
```
epau, line 111:
```suggestion
    /// Compares two `JMESValue` values per the JMESPath rules.
    /// This always returns `false` if both the left and right values are not JMESPath numbers.
    /// - Parameters:
    ///   - lhs: The left value
    ///   - rhs: The right value
    /// - Returns: `true`  if `lhs` and `rhs` are both JMESPath numbers and the `lhs` value is less than the `rhs` value, otherwise returns `false`.
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
```diff
@@ -0,0 +1,106 @@
...
+import software.amazon.smithy.waiters.Acceptor
+import software.amazon.smithy.waiters.Matcher
+import software.amazon.smithy.waiters.PathComparator
+import software.amazon.smithy.waiters.PathMatcher
+
+// Renders the initializer for one Smithy waiter acceptor into Swift,
+// using the provided writer & shapes.
+// The Acceptor is defined in Smithy here:
+// https://smithy.io/2.0/additional-specs/waiters.html#acceptor-structure
+class WaiterAcceptorGenerator(
```
epau, line 23: Can you add tests for this that verify the generated code is what we expect? That will also be a big help with reviewing this!

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
jbelkins: These are optional to conform with the "every value is optional" assumption that is baked into JMESPath evaluation.

The initializer takes non-nil values because they will always actually be present when this type is instantiated; an `imputOutput` matcher always fails if there is no output.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    private fun childBlock(forExpression: JmespathExpression): String =
+        forExpression.accept(JMESPathVisitor(writer))
+
+    // Maps the expression result in leftName into a new collection using the right expression.
+    private fun mappingBlock(right: JmespathExpression, leftName: String): String {
+        if (right is CurrentExpression) return leftName // Nothing to map
+        val outerName = bestTempVarName("projection")
+        writer.openBlock("let \$L = Optional.some((\$L ?? []).compactMap { current in", "})", outerName, leftName) {
+            writer.write("let current = Optional.some(current)")
+            val innerResult = addTempVar("innerResult", childBlock(right))
```
epau, line 70: In an effort to prefer explicit over implicit, we should pass in the local property name to the `childBlock` function and then to the new `JMESPathVisitor` instance that is created. This will make the logic easier to read and also easier to change.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    // within them, i.e. to map or filter.  This method is called to render
+    // those expressions.
+    private fun childBlock(forExpression: JmespathExpression): String =
+        forExpression.accept(JMESPathVisitor(writer))
+
+    // Maps the expression result in leftName into a new collection using the right expression.
+    private fun mappingBlock(right: JmespathExpression, leftName: String): String {
+        if (right is CurrentExpression) return leftName // Nothing to map
+        val outerName = bestTempVarName("projection")
+        writer.openBlock("let \$L = Optional.some((\$L ?? []).compactMap { current in", "})", outerName, leftName) {
```
epau, line 68: If the resulting property can be optional, then there is no need to wrap this in an Optional.

```suggestion
        writer.openBlock("let \$L = \$L?.compactMap { current in", "}", outerName, leftName) {
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    override fun visitFunction(expression: FunctionExpression): String {
+        when (expression.name) {
+            "contains" -> {
+                if (expression.arguments.size != 2) { throw Exception("Unexpected number of arguments to $expression") }
+
+                val subject = expression.arguments[0]
+                val subjectName = subject.accept(this)
+                val search = expression.arguments[1]
+                val searchName = search.accept(this)
+                return addTempVar("contains", "\$L.flatMap { \$L?.contains($$0) } ?? false", searchName, subjectName)
```
epau, line 149: I didn't see an example of this case in the `acceptors with generated code` branch. Can we make sure we have a test that hits this code path

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    override fun visitIndex(expression: IndexExpression): String {
+        throw Exception("IndexExpression is unsupported")
+    }
+
+    // Renders a literal of any supported type, wrapped in an optional.
+    override fun visitLiteral(expression: LiteralExpression): String {
+        when (expression.type) {
+            RuntimeType.STRING -> return addTempVar("string", "Optional.some(\"\$L\")", expression.expectStringValue())
+            RuntimeType.NUMBER -> return addTempVar("number", "Optional.some(Double(\$L))", expression.expectNumberValue())
+            RuntimeType.BOOLEAN -> return addTempVar("bool", "Optional.some(\$L)", expression.expectBooleanValue())
```
epau, line 171: It looks like we don't need to wrap these in optionals since the result is always passed into a `JMESValue`. It also feels a bit off to be wrapping literals in optionals when we know they aren't optional.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+                val search = expression.arguments[1]
+                val searchName = search.accept(this)
+                return addTempVar("contains", "\$L.flatMap { \$L?.contains($$0) } ?? false", searchName, subjectName)
+            }
+            "length" -> {
+                if (expression.arguments.size != 1) { throw Exception("Unexpected number of arguments to $expression") }
+
+                val subject = expression.arguments[0]
+                val subjectName = subject.accept(this)
+                return addTempVar("count", "Optional.some(Double(\$L?.count ?? 0))", subjectName)
```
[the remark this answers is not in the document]
epau: Alternatively, if checking the count of a nil object should throw an error then should we be guarding here and throwing an error if the subject is nil?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    private val tempVars = mutableSetOf<String>()
+
+    private fun addTempVar(preferredName: String, content: String, vararg args: Any): String {
+        val name = bestTempVarName(preferredName)
+        writer.writeInline("let \$L = ", name)
+        writer.write(content, *args)
+        return name
+    }
+
+    private fun bestTempVarName(preferredName: String): String =
```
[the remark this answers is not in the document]
jbelkins: Best means "unique, based on the preferred name passed in".  So if you ask for `potato` and it hasn't already been used, it will pass back `potato`.  If there is already a `potato` it will pass back `potato2`, then `potato3` on the next call, etc.

Unless you have a better idea, I'll call it `uniqueTempVarName` and add a comment to describe behavior.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+    override fun visitCurrentNode(expression: CurrentExpression): String {
+        throw Exception("Unexpected current expression outside of flatten expression: $expression")
+    }
+
+    override fun visitExpressionType(expression: ExpressionTypeExpression): String {
+        throw Exception("ExpressionTypeExpression is unsupported")
+    }
+
+    override fun visitField(expression: FieldExpression): String {
+        return subfield(expression, "current")
```
epau, line 112: Like I mentioned in another comment, it'd be great if "current" was an instance property that way we can easily change the value to make generate code that is easier to read. (Giving it a default value of "current" is fine though)

Packages/ClientRuntime/Sources/Waiters/WaiterUtils.swift
epau: This type is great at it's job, but it is an implementation detail. Given that, it'd be great to keep this internal and not expose it publicly, especially since JMES is a real thing and we are only defining this one small piece of it. 

I'd love if we exposed some function instead that used this type under the hood. Something like 
```swift
extension WaiterConfiguration.Acceptor {

    public enum Comparator {
        case equals
        case lessThan
    }
    public static func compareValues(_ lhs: Any?, _ rhs: Any?, comparator: Comparator) -> Bool {
            let lhsValue = JMESValue(lhs)
            let rhsValue = JMESValue(rhs)
            switch comparator {
            case .equals: 
                return lhsValue == rhsValue
            case .lessThan:
                return lhsValue < rhsValue
            }
    }
}
```

And then you just need to add an initializer for JMESValue that takes an `Any?`. Could even add an enum case for "unsupported type" and that way you can define equals and less than explicitly for those cases.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+
+                    // On each waitable trait, get all its waiters and render a waitUntil for each
+                    waitableTraits.forEach { waitableTrait ->
+                        waitableTrait.waiters.forEach { (waiterName, waiter) ->
+                            writer.write("")
+                            WaiterConfigGenerator(writer, ctx, service, waitedOperation, waiterName, waiter).render()
+                            writer.write("")
+                            WaiterMethodGenerator(writer, ctx, service, waitedOperation, waiterName, waiter).render()
+                        }
+                    }
```
epau, line 66: it'd be great to put the config in a separate file (or at least separate extension) from the methods. This way, all the public waiter methods are colocated and easy to view all of them alongside eachother, without the non-public config methods being there. 

Additionally, can we move the config methods to an extension on WaiterConfiguration? That seems like a more natural place for them given they are essentially factory methods for config. Not fussed about this though since they are not public.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+        val service = ctx.model.expectShape<ServiceShape>(ctx.settings.service)
+
+        // Open a new file Waiters.swift to hold the waiter definitions for this service
+        val waiterFilename = "${ctx.settings.moduleName}/Waiters.swift"
+        delegator.useFileWriter(waiterFilename) { writer ->
+            writer.addImport(SwiftDependency.CLIENT_RUNTIME.target)
+            val serviceSymbol = ctx.symbolProvider.toSymbol(service)
+
+            // Only add waiters if Swift version is 5.7 or higher
+            writer.write("#if swift(>=5.7)")
```
epau, line 45: Can we add a comment here describing why we only support waiters for 5.7 or higher

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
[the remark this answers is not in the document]
epau: Nope, and tbh the current name is fine too so defer to w/e you think is best (no pun intended lol). A comment describing that behavior would be great though!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
```diff
@@ -0,0 +1,106 @@
...
+        // Depending on the type of waiter, current is set to establish that scope.
+        if (includeInput) {
+            writer.write(
+                "let current = Optional.some(WaiterConfiguration<\$L, \$L>.Acceptor.InputOutput(input: input, output: unwrappedOutput))",
+                inputTypeName,
+                outputTypeName
+            )
+        } else {
+            writer.write("let current = Optional.some(unwrappedOutput)")
+        }
```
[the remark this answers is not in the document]
epau: Nvm, i see why this is needed now.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        }
+    }
+
+    private fun renderInputOutputBlockContents(includeInput: Boolean, pathMatcher: PathMatcher) {
+        writer.write("// JMESPath expression: ${pathMatcher.path}")
+        writer.write("// JMESPath comparator: ${pathMatcher.comparator}")
+        writer.write("// JMESPath expected value: ${pathMatcher.expected}")
+        writer.write("guard case .success(let unwrappedOutput) = result else { return false }")
```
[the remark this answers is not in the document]
epau: I see why this is needed, this is fine.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
epau: i really don't like this, but i see why it needs to be this way. 

it would be nice to not depend on the java jmes parser, or make changes to it, so that we aren't locked into behaviors like this. I noticed that its forcing our hand with how we generate the code.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
jbelkins: Done, there are now codegen tests for each type of acceptor, though the expected code will change as JMESPath code generation changes.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
jbelkins: Sure thing, I don't se a problem with making them permanent

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterGenerator.kt
```diff
@@ -53,17 +46,23 @@ class WaiterGenerator : SwiftIntegration {
...
                             writer.write("")
                             WaiterConfigGenerator(writer, ctx, service, waitedOperation, waiterName, waiter).render()
                             writer.write("")
-                            WaiterMethodGenerator(writer, ctx, service, waitedOperation, waiterName, waiter).render()
+                            WaiterMethodGenerator(writer, ctx, service, waitedOperation, waiterName).render()
                         }
                     }
+
+                    // If any of this operation's waiters uses an error type matcher, render WaiterTypedError
+                    // conformance for this operation's error
```
epau, line 62: appreciate the comments!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
```diff
@@ -36,7 +36,6 @@ class HttpResponseGenerator(
         httpOperations.forEach {
             httpResponseBindingErrorGenerator.render(ctx, it)
             HttpResponseBindingErrorNarrowGenerator(ctx, it, unknownServiceErrorSymbol).render()
-            WaiterTypedErrorGenerator(ctx, it, unknownServiceErrorSymbol).render()
```
jbelkins, line 39: Removes `WaiterTypedError` conformance from every operation error.  Now `WaiterTypedError` conformance is only added where actually needed.

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
```diff
@@ -0,0 +1,92 @@
...
+            return left == right
+        case (.null, .null):
+            return true
+        default:
+            return false
+        }
+    }
+
+    /// Compares two `JMESValue` values, testing for order per the JMESPath rules.
+    /// Comparison fails if both left and right are not JMESPath numbers.
```
epau, line 79:
```suggestion
    /// Returns false if both left and right are not JMESPath numbers.
```

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
```diff
@@ -0,0 +1,92 @@
...
+    /// - Parameters:
+    ///   - lhs: The left value
+    ///   - rhs: The right value
+    /// - Returns: The result of the Swift comparison of the associated values if `lhs` and `rhs` are both JMESPath numbers, or `false` otherwise.
```
epau, line 83:
```suggestion
    /// - Returns: True if both the `lhs` and `rhs` are numbers and the `lhs` is less than the `rhs`, otherwise returns false.
```

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
```diff
@@ -0,0 +1,92 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A value type that is used to facilitate comparison of values returned by JMESPath types, in accordance with the JMESPath spec.
```
epau, line 10: Can we add a link to the aforementioned JMESPath spec?

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
```diff
@@ -0,0 +1,92 @@
...
+public enum JMESValue: Equatable, Comparable {
```
epau, line 11: Nit: imo it'd be great if we didn't need to expose this publicly since its an implementation detail of the waiters acceptors and instead we just publicly exposed some static function on Acceptors or something. Especially since we just need this type to perform comparisons.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
epau: Can we add a comment here that adds this context since its not obvious, from just looking at the swift code, as to why these need to be optional

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterAcceptorGenerator.kt
jbelkins: `errorType` is now implemented.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/WaiterGenerator.kt
jbelkins: This is removed, since explicit types have been added to make acceptors compile on Swift 5.5+

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/Variable.kt
```diff
@@ -0,0 +1,35 @@
+package software.amazon.smithy.swift.codegen.waiters
+
+import software.amazon.smithy.codegen.core.SymbolProvider
+import software.amazon.smithy.model.shapes.Shape
+
+// Holds the name of a variable defined in Swift, plus the optionality and the Smithy
+// shape equivalent of its type.
+class Variable(
```
epau, line 8: Can we move this file out of a the waiters folder and to somewhere more suitable for general utility objects?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/BufferWriter.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.util.UUID
+
+/**
+ * A class which uses a writer to "record" code into an in-memory buffer,
+ * then "plays it back" onto the writer at a later time.
+ *
+ * Useful when the result of rendering an expression is needed before the expression
+ * is actually to be included in the code.
+ */
+class BufferWriter(
```
epau, line 18: Can we move this file out of a the waiters folder and to somewhere more suitable for general utility objects?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,412 @@
...
+                val subfieldMember = parentMembers.first { it.memberName == expression.name }
+                val subfieldShape = model.expectShape(subfieldMember.target)
+                val subfieldName = symbolProvider.toMemberName(subfieldMember)
+                val fieldOperator = "?.".takeIf { parentVar.isOptional } ?: "."
+                //
+                // Because all fields on Swift models are currently optional, every subfield value will also be
+                // optional.  Later we may have to replace this with logic to determine actual optionality,
+                // so:
+                // val subfieldIsOptional = !subfieldMember.hasTrait<RequiredTrait>()
+                // is temporarily replaced with:
```
epau, line 142: great context!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,412 @@
...
+        return addTempVar(andResultVar, "\$L && \$L", leftExp, rightExp)
+    }
+
+    // Perform a comparison of two values.
+    // The JMESValue type is used to provide conversion and comparison as needed between types
+    // that aren't comparable in "pure Swift" (i.e. Int to Double or String to RawRepresentable
+    // by String.)
+    // The Smithy comparator is a string that just happens to match up with all Swift comparators,
+    // so it is rendered into Swift as-is.
+    override fun visitComparator(expression: ComparatorExpression): Variable {
```
epau, line 169: It appears that we only support two comparators currently `==` and `<`. Can we perform some validation of `expression.comparator` to verify it's one that we support, this way we'll get an error during codegen if a new comparator is added that we don't support. imo that is preferable to getting an error during swift compile time as it makes it much more clear what the actual issue is.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+        val right = expression.right!!
+        return mappingBlock(right, filteredName)
+    }
+
+    // Returns the inner expression unchanged when the inner expression is an array of non-array elements.
+    // Returns the inner expression flattened when the inner expression is an array of arrays.
+    override fun visitFlatten(expression: FlattenExpression): String {
+        val innerName = expression.expression!!.accept(this)
+        return addTempVar("${innerName}OrEmpty", "Optional.some((\$L ?? []).flattenIfPossible { $$0 })", innerName)
```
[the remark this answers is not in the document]
jbelkins: Arrays are now handled as required or optional as appropriate

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/Variable.kt
jbelkins: Moved to `software.amazon.smithy.swift.codegen.utils`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/BufferWriter.kt
jbelkins: Moved to `software.amazon.smithy.swift.codegen.utils`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: I'm sure there will be bugs, but flipping this switch should be all that's needed to support non-optional fields on model objects somewhere down the line

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: If you implement `==` and `<` in Swift, you support all the other Smithy operators because Swift provides all the others through default implementations on the `Equatable` and `Comparable` protocols.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,412 @@
...
+                    throw Exception("Unexpected number of arguments to $expression")
+                }
+                val subjectExp = expression.arguments[0]
+                val subject = subjectExp.accept(this)
+
+                return when (subject.shape) {
+                    is StringShape, is ListShape, is MapShape -> {
+                        val countVar = Variable("count", false, doubleShape)
+                        val optionalityMark = "?".takeIf { subject.isOptional } ?: ""
+                        addTempVar(countVar, "Double(\$L\$L.count ?? 0)", subject.name, optionalityMark)
```
epau, line 299: We likely only want to do the nil coalescing if the subject is optional, otherwise this will result in a warning

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: For reference here are Smithy's comparators: https://github.com/awslabs/smithy/blob/bf9fc069338fe78ed3dfc1f51820a0101dfec12b/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/ast/ComparatorType.java#L21

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,412 @@
...
+        return when (original.shape) {
+            is MapShape -> {
+                val valueShape = model.expectShape(original.shape.value.target)
+                val projectionShape = ListShape.builder()
+                    .id("smithy.swift.synthetic#ObjectProjection")
+                    .member(valueShape.toShapeId())
+                    .build()
+                var projectionVar = Variable("projection", original.isOptional, projectionShape)
+                if (original.isOptional) {
+                    projectionVar = addTempVar(projectionVar, "\$L.flatMap { Array($$0.values) }", original.name)
```
epau, line 368: why `flatMap` here? should this just be `map` instead?

jbelkins: This is the optional version of `flatMap`, not the array version.

jbelkins: https://developer.apple.com/documentation/swift/optional/flatmap(_:)#

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
```diff
@@ -0,0 +1,412 @@
...
+                val subject = expression.arguments[0]
+                val subjectVariable = subject.accept(this)
+                val search = expression.arguments[1]
+                val searchVariable = search.accept(this)
+                val optionalityMark = "?".takeIf { subjectVariable.isOptional } ?: ""
+                val returnValueVar = Variable("contains", false, boolShape)
+                return if (searchVariable.isOptional) {
+                    addTempVar(
+                        returnValueVar,
+                        "\$L.flatMap { \$L\$L.contains($$0) } ?? false",
```
epau, line 273: Why `flatMap` here? Should this just be `map` instead?

jbelkins: This is the optional version of `flatMap`, not the array version.
https://developer.apple.com/documentation/swift/optional/flatmap(_:)#
Evaluates & returns the value of the closure if the receiver is non-nil, returns nil if the receiver is nil.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
epau: Ahh that makes sense, i missed the `Comparable` conformance. 
awesome!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
epau: right, so should we only be using `flatMap` if `subjectVariable` is optional and use `map` otherwise?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
epau: Right, so shouldn't we use `map` here since the closure returns a non-optional?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: Nil-coalescence here is now added only if the subject is Optional.

smithy-swift-codegen/src/test/kotlin/waiters/WaiterAcceptorGeneratorTests.kt
```diff
@@ -0,0 +1,122 @@
...
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.swift.codegen.SwiftSettings
+import software.amazon.smithy.swift.codegen.core.CodegenContext
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
+import software.amazon.smithy.swift.codegen.model.expectShape
+import software.amazon.smithy.swift.codegen.waiters.WaiterAcceptorGenerator
+import software.amazon.smithy.waiters.WaitableTrait
+
+class WaiterAcceptorGeneratorTests {
```
epau, line 26: It'd be nice to have a few more tests covering the variants of acceptors, such as those with mapping and flat mapping and compact mapping etc.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: No, map is not needed in the other case.
I am using Optional `flatMap()` as an easy, compact way to unwrap an optional for use in an expression, but default to `nil` if the optional is `nil`.
When the variable is not optional, I can just use it in the expression directly.

```swift
let result = subject.flatMap { collection.contains($0) } ?? false
```
is just a shorthand way of writing
```swift
let result: Bool
if let subject = subject {
    result = collection.contains(subject)
} else {
    result = false
}
```

epau (commented):
This is looking fantastic! Really great job here! Just a few super small comments

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
epau: Exactly! `map` is what we should be reaching for in your specific example here!

The difference between `Optional.map` and `Optional.flatMap` is that the closures have different return types. 
`map` expects the closure to return a non-optional while `flatMap` expects the closure to return an optional. 

`Optional.map` -> `func map<U>(_ transform: (Wrapped) throws -> U) rethrows -> U?`
`Optional.flatMap` -> `func flatMap<U>(_ transform: (Wrapped) throws -> U?) rethrows -> U?`

So if we know the return type of the closure is non-optional, it is preferable to use map.

smithy-swift-codegen/src/test/kotlin/waiters/WaiterAcceptorGeneratorTests.kt
jbelkins: We already have functional tests for these operations in the other repo.  I added tests there for projection this morning as a result of a build failure on one AWS service.

Really the only thing that we gain with these tests when we have the other ones is coverage of the generated comments.

All these tests do is break when I make superficial changes to the acceptor code, so if I add these it will be the very last thing before I merge.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/Variable.kt
epau: Can we rename it then to something domain specific like `JMESPathVariable`?

smithy-swift-codegen/src/test/kotlin/waiters/WaiterAcceptorGeneratorTests.kt
epau: >Really the only thing that we gain with these tests when we have the other ones is coverage of the generated comments.

We also gain better PR reviews since we easily see the generated code alongside the PR. Additionally, we get test failures that are easier to debug. If a tests fails during codegen, then its obvious that we aren't generating the expected code. If a protocol tests fails, that doesn't necessarily mean we're generating the wrong code but rather that the behavior is unexpected. The fix might be to update the generated code or it might be to update swift code etc. It's really nice to have the failure point narrowed down.

epau: This it non-blocking btw, just think it'd be nice to have.

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
jbelkins: Removed in favor of several overloaded static functions.  See JMESUtils.swift.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/Variable.kt
jbelkins: Renamed.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: The closure returns a non-optional, but the return value of `flatMap` is optional.

Packages/ClientRuntime/Sources/Waiters/JMESValue.swift
jbelkins: This source file is now deleted.

Packages/ClientRuntime/Sources/Waiters/Acceptor.swift
jbelkins: Comment added.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: Changed to `map`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/waiters/JMESPathVisitor.kt
jbelkins: Changed to `map`

epau (approved):

PR #487, opened 2022-12-01
Fix log level parameter being ignored

CharlesJS:
The `level` argument to [this function](https://github.com/awslabs/smithy-swift/blob/ae509e0025ff425df3bf0847413574d0a151d8c0/Packages/ClientRuntime/Sources/Logging/SwiftLog%2BLogAgent.swift#L32-L50) is currently ignored. Instead, the function calls the `logLevel` property on the `SwiftLogger` struct. This means that no matter whether a log is created via any level, whether it's `.debug()` or `.fatal()`, the logs that are generated are always created at whatever level the `SwiftLogger` has in its stored property, which is usually `.info`. I'm pretty sure this is a typo, and it causes any command-line tools that are made using this SDK to output so much debug information that it obscures the intended program output.

## Issue \#
https://github.com/awslabs/smithy-swift/issues/486

## Description of changes
It fixes a typo, so that the `level` parameter is actually used, and logs get logged at the level the caller intended to log them at.

## Scope

### DIFF
--- a/Packages/ClientRuntime/Sources/Logging/SwiftLog+LogAgent.swift
+++ b/Packages/ClientRuntime/Sources/Logging/SwiftLog+LogAgent.swift
@@ -39,7 +39,7 @@ public struct SwiftLogger: LogAgent {
         let mappedDict = metadata?.mapValues { (value) -> Logger.MetadataValue in
             return Logger.MetadataValue.string(value)
         }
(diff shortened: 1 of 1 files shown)


jbelkins:
Closing.  I plan to overhaul logging prior to Swift SDK general availability.

PR #488, opened 2022-12-05
Fixes XML encoding issues

epau:
This PR fixes the following XML encoding and decoding issues 
- Incorrect root key when serializing XML
- Incorrect serialized float and double values when value is not a number or infinity

Additionally, this PR improves our serialization tests. 

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/711

## Description of changes
The root key for XML is now the correct value. Previously, we didn't use the `xmlName` trait if one existed, nor did we account for the fact that our local swift models do not have the same name as their counterparts defined in the smithy model. This PR ensures that we now use the value defined by the `xmlName` trait (if one exists). it also sets the value of the `xmlName` trait to the shape's original name (as defined in the smithy model) if the shape doesn't define it's own explicit `xmlName`. This allows us to different name in swift code for the corresponding structure. 

Previously, we encoded floats and doubles as strings in XML and never defined handling for when the value is not-a-number or infinity. This PR encodes floats and doubles as is and defines how they should be encoded when the value is not-a-number or infinity (according to the smithy spec)

Lastly, this PR improves the accuracy of our serialization tests. It adds a check that verifies the serialized data matches the expected data. Previously, we would simply check that we could decode the data that we encoded which can produce many false-positives.

## Scope

### DIFF
--- a/Packages/SmithyTestUtil/Sources/JSONComparator.swift
+++ b/Packages/SmithyTestUtil/Sources/JSONComparator.swift
@@ -1,0 +1,53 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public struct JSONComparator {
+    /// Returns true if the JSON documents, for the corresponding data objects, are equal.
+    /// - Parameters:
+    ///   - dataA: The first data object to compare to the second data object.
+    ///   - dataB: The second data object to compare to the first data object.
+    /// - Returns: Returns true if the JSON documents, for the corresponding data objects, are equal.
+    public static func jsonData(_ dataA: Data, isEqualTo dataB: Data) throws -> Bool {
+        let jsonDictA = try JSONSerialization.jsonObject(with: dataA)
+        let jsonDictB = try JSONSerialization.jsonObject(with: dataB)
+        return anyValuesAreEqual(jsonDictA, jsonDictB)
+    }
+}
+
+fileprivate func anyDictsAreEqual(_ lhs: [String: Any], _ rhs: [String: Any]) -> Bool {
+    guard lhs.keys == rhs.keys else { return false }
+    for key in lhs.keys {
+        if !anyValuesAreEqual(lhs[key], rhs[key]) {
+            return false
+        }
+    }
+    return true
+}
+
+fileprivate func anyArraysAreEqual(_ lhs: [Any], _ rhs: [Any]) -> Bool {
+    guard lhs.count == rhs.count else { return false }
+    for i in 0..<lhs.count {
+        if !anyValuesAreEqual(lhs[i], rhs[i]) {
+            return false
+        }
+    }
+    return true
+}
+
+fileprivate func anyValuesAreEqual(_ lhs: Any?, _ rhs: Any?) -> Bool {
+    if lhs == nil && rhs == nil { return true }
+    guard let lhs = lhs, let rhs = rhs else { return false }
+    if let lhsDict = lhs as? [String: Any], let rhsDict = rhs as? [String: Any] {
+        return anyDictsAreEqual(lhsDict, rhsDict)
+    } else if let lhsArray = lhs as? [Any], let rhsArray = rhs as? [Any] {
+        return anyArraysAreEqual(lhsArray, rhsArray)
+    } else {
+        return type(of: lhs) == type(of: rhs) && "\(lhs)" == "\(rhs)"
+    }
+}
--- a/Packages/SmithyTestUtil/Sources/XCTAssertions.swift
+++ b/Packages/SmithyTestUtil/Sources/XCTAssertions.swift
@@ -19,9 +19,9 @@ public func XCTAssertJSONDataEqual(
         let data1 = try expression1()
         let data2 = try expression2()
         guard data1 != data2 else { return }
-        let jsonDict1 = try JSONSerialization.jsonObject(with: data1) as! NSDictionary
-        let jsonDict2 = try JSONSerialization.jsonObject(with: data2) as! NSDictionary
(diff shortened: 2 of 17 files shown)


Packages/SmithyTestUtil/Sources/XCTAssertions.swift
```diff
@@ -0,0 +1,45 @@
...
+
+import Foundation
+import XCTest
+
+public func XCTAssertJSONDataEqual(
+    _ expression1: @autoclosure () throws -> Data,
+    _ expression2: @autoclosure () throws -> Data,
+    _ message: @autoclosure () -> String = "",
+    file: StaticString = #filePath,
+    line: UInt = #line
```
jbelkins, line 16: Hmmm, never thought of using `#filePath` and `#line` in test helpers & custom assertions (only ever used it in logging statements.)

Packages/SmithyTestUtil/Sources/XCTAssertions.swift
```diff
@@ -0,0 +1,45 @@
...
+) {
+    do {
+        let data1 = try expression1()
+        let data2 = try expression2()
+        guard data1 != data2 else { return }
+        let jsonDict1 = try JSONSerialization.jsonObject(with: data1) as! NSDictionary
+        let jsonDict2 = try JSONSerialization.jsonObject(with: data2) as! NSDictionary
```
jbelkins, line 23: Why cast to `NSDictionary` here; why not keep it as a Swift type?

Also, are we sure we will never be comparing JSON arrays or fragments here?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/handlers/HttpBodyMiddleware.kt
```diff
@@ -107,7 +109,19 @@ class HttpBodyMiddleware(
                 writer.openBlock("do {", "} catch let err {") {
                     writer.write("let encoder = context.getEncoder()")
                     writer.openBlock("if let $memberName = input.operationInput.$memberName {", "} else {") {
-                        writer.write("let $dataDeclaration = try encoder.encode(\$L)", memberName)
+
+                        val xmlNameTrait = binding.member.getTrait<XmlNameTrait>() ?: target.getTrait<XmlNameTrait>()
+                        if (xmlNameTrait != null) {
+                            val xmlName = xmlNameTrait.value
+                            writer.write("let xmlEncoder = encoder as! XMLEncoder")
```
jbelkins, line 116: Please explain why we should be confident force-casting here will never fail

jbelkins (commented):
Please resolve a few Q's for me. Otherwise this looks good

jbelkins:
Oh, one more thing, see your build failures on `smithy-swift`... Foundation XML types require a special `import` on Linux.

Packages/SmithyTestUtil/Sources/XCTAssertions.swift
epau: Yeah its nice to use here so that the assertion failure will create a nice error message at the proper line in the test. Otherwise, it can be hard to figure out the cause of the error.

Packages/SmithyTestUtil/Sources/XCTAssertions.swift
epau: > Why cast to NSDictionary here; why not keep it as a Swift type?

`NSDictionary` gives us equality checking for free. The bridged swift type would require explicit typing or require a custom algorithm to do a deep equality check. 

> Also, are we sure we will never be comparing JSON arrays or fragments here?

Currently yes given all the currents tests pass.
I'd prefer to defer adding additional handling until it is needed in an effort to not write code that is never used. (adding handling for arrays would be simple but supporting fragments could be complicated. Currently `JSONSerialization` will fail to even generate an object for fragments given we aren't passing in [the option to allow fragments](https://developer.apple.com/documentation/foundation/jsonserialization/writingoptions/3081259-fragmentsallowed))

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/handlers/HttpBodyMiddleware.kt
epau: Good callout, will verify that the service protocol is xml here

jbelkins (approved):

jbelkins (commented):
Just one Q

Packages/SmithyTestUtil/Sources/JSONComparator.swift
```diff
@@ -0,0 +1,53 @@
...
+    /// - Returns: Returns true if the JSON documents, for the corresponding data objects, are equal.
+    public static func jsonData(_ dataA: Data, isEqualTo dataB: Data) throws -> Bool {
+        let jsonDictA = try JSONSerialization.jsonObject(with: dataA)
+        let jsonDictB = try JSONSerialization.jsonObject(with: dataB)
+        return anyValuesAreEqual(jsonDictA, jsonDictB)
+    }
+}
+
+fileprivate func anyDictsAreEqual(_ lhs: [String: Any], _ rhs: [String: Any]) -> Bool {
+    guard lhs.keys == rhs.keys else { return false }
```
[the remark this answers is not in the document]
epau: There is no order here, since `keys` is a struct. I imagine the sequence and collection conformance methods likely will return keys in a non-deterministic order but those aren't in play here.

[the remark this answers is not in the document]
epau: Here's the relevant bits of the equatable conformance for the Keys struct https://github.com/apple/swift/blob/main/stdlib/public/core/Dictionary.swift#L1388-L1400

PR #489, opened 2022-12-06
chore: bump CRT to 0.4.0

jbelkins:
## Description of changes
Use aws-crt-swift version 0.4.0, change version to locked rather than floating to next major

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,7 @@ let package = Package(
         .library(name: "SmithyTestUtil", targets: ["SmithyTestUtil"])
     ],
     dependencies: [
-        .package(url: "https://github.com/awslabs/aws-crt-swift.git", from: "0.3.1"),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -14,7 +14,7 @@ let package = Package(
         .library(name: "SmithyTestUtil", targets: ["SmithyTestUtil"])
     ],
     dependencies: [
-        .package(url: "https://github.com/awslabs/aws-crt-swift.git", from: "0.3.1"),
+        .package(url: "https://github.com/awslabs/aws-crt-swift.git", .exactItem("0.4.0")),
```
epau, line 17: why `exactItem` here? in aws-sdk-swift we use just `exact` https://github.com/awslabs/aws-sdk-swift/blob/0.5.0/Package.swift

epau (approved):

Package.swift
jbelkins: Changed to `.exact()`

PR #491, opened 2022-12-08
feat: Add WaiterTypedError type, conform operation errors to it

jbelkins:
## Description of changes

This PR merges to the waiters feature branch.

Adds a `WaiterTypedError` Swift protocol so that an error can provide its error type.  This will be used in the implementation of the `errorType` acceptor for Smithy waiters, which will match the error based on an expected error type string that will match the `errorCode` property on a `WaiterTypedError`.
- Conforms `SdkError` to `WaiterTypedError`
- Conforms operation errors to `WaiterTypedError`
- Conforms unknown HTTP errors to `WaiterTypedError`
- Adds a test for correct codegen of `WaiterTypedError` compliance of an `OperationError`

As written, this PR generates `WaiterTypedError` compliance for every operation error, not just the ones with `errorType` waiters.  When this work is integrated with Smithy acceptor work, `WaiterTypedError` compliance will be generated only for operations that have a waiter with an `WaiterTypedError` acceptor.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
```diff
@@ -36,6 +36,7 @@ class HttpResponseGenerator(
         httpOperations.forEach {
             httpResponseBindingErrorGenerator.render(ctx, it)
             HttpResponseBindingErrorNarrowGenerator(ctx, it, unknownServiceErrorSymbol).render()
+            CodedErrorGenerator(ctx, it, unknownServiceErrorSymbol).render()
```
jbelkins, line 39: Rendering coded error compliance here does it for every operation.  When this work is integrated with Smithy acceptors, coded error compliance will only be generated for operations that have a waiter with an `errorType` acceptor.

Packages/ClientRuntime/Sources/Networking/CodedError.swift
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import Foundation
+
+public protocol CodedError: Error {
```
epau, line 8: Can we add a doc comment for this type?

Packages/ClientRuntime/Sources/Networking/CodedError.swift
```diff
@@ -0,0 +1,16 @@
...
+
+    /// The `String` error code that identifies the general type of this HTTP service error, or `nil` if the
+    /// error has no known code.
+    ///
+    /// The source of this code is the HTTP response to an API call, but what exact header, body field, etc.
+    /// it comes from is specific to the response encoding used.
```
epau, line 14: We should refrain from documenting protocol methods with details from specific conforming types.
```suggestion
    /// The error code.
```

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
```diff
@@ -5,6 +5,8 @@
 
 /// General Service Error structure used when exact error could not be deduced from the `HttpResponse`
 public struct UnknownHttpServiceError: HttpServiceError, Swift.Equatable {
+    public var _errorCode: String?
```
epau, line 8: Unrelated to this PR:
I wonder why we prefix all the properties here with an underscore? 🤔 
Especially since there properties are public

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
```diff
@@ -19,9 +21,15 @@ public struct UnknownHttpServiceError: HttpServiceError, Swift.Equatable {
...
         self._statusCode = httpResponse.statusCode
         self._headers = httpResponse.headers
         self._message = message
+        self._errorCode = errorCode
     }
 }
+
+extension UnknownHttpServiceError: CodedError {
+
+    public var errorCode: String? { _errorCode }
```
epau, line 34: Can we add docs as to what an error code is (or should be) for an UnknownHttpServiceError?

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
```diff
@@ -19,9 +21,15 @@ public struct UnknownHttpServiceError: HttpServiceError, Swift.Equatable {
 }
 
 extension UnknownHttpServiceError {
-    public init(httpResponse: HttpResponse, message: String? = nil) {
+    public init(httpResponse: HttpResponse, message: String? = nil, errorCode: String? = nil) {
```
epau, line 24: Mind adding docs for this initializer while you're here?

Packages/ClientRuntime/Sources/Networking/SdkError.swift
```diff
@@ -14,3 +14,18 @@ public enum SdkError<E>: Error {
   case unknown(Error?)
 
 }
+
+
+extension SdkError: CodedError {
+
+    public var errorCode: String? {
```
epau, line 21: Can we add docs as to what an error code is (or should be) for a SdkError?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/CodedErrorGenerator.kt
```diff
@@ -0,0 +1,72 @@
...
+            val context = mapOf(
+                "ctx" to ctx,
+                "unknownServiceErrorType" to unknownServiceErrorType,
+                "operationErrorName" to operationErrorName,
+                "errorShapes" to errorShapes
+            )
+            writer.declareSection(CodedErrorGeneratorSectionId, context) {
+                writer.openBlock("extension \$L: CodedError {", "}", operationErrorName) {
+                    writer.write("")
+                    writer.openBlock("public var errorCode: String? {", "}") {
```
epau, line 49: Can we add docs for `errorCode` here?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/CodedErrorGenerator.kt
```diff
@@ -0,0 +1,72 @@
...
+                        writer.write("switch self {")
+                        for (errorShape in errorShapes) {
+                            var errorShapeName = resolveErrorShapeName(errorShape)
+                            var errorShapeType = ctx.symbolProvider.toSymbol(errorShape).name
+                            var errorShapeEnumCase = errorShapeType.decapitalize()
+                            writer.write("case .\$L: return \$S", errorShapeEnumCase, errorShapeName)
```
epau, line 55: Why do we return the error type as a string here? Should we return nil instead?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
epau: Does that mean that the implementation of `CodedErrorGenerator` here will be removed? Assuming it will be implemented somewhere else where it is restricted to only operations that have a waiter with an errorType acceptor?

epau (commented):
Looking really good! Just a few comments and questions

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/CodedErrorGenerator.kt
jbelkins: Returning the error type as a string here prevents me from having to extend every error on the operation with its own `CodedError` compliance... the `GetWidgetOutputError` knows the right error code for every possible error it can return.

Compare this to the output error initializer which switches over the expected values of errorCode to determine what exact error to create for the operation & this method makes more sense.  Here's an (abridged) initializer for an operation from an AWS service:
```swift
extension ModifyDBInstanceOutputError {
    public init(errorType: Swift.String?, httpResponse: ClientRuntime.HttpResponse, decoder: ClientRuntime.ResponseDecoder? = nil, message: Swift.String? = nil, requestID: Swift.String? = nil) throws {
        switch errorType {
        case "AuthorizationNotFound" : self = .authorizationNotFoundFault(try AuthorizationNotFoundFault(httpResponse: httpResponse, decoder: decoder, message: message, requestID: requestID))
        case "CertificateNotFound" : self = .certificateNotFoundFault(try CertificateNotFoundFault(httpResponse: httpResponse, decoder: decoder, message: message, requestID: requestID))
        default : self = .unknown(UnknownAWSHttpServiceError(httpResponse: httpResponse, message: message, requestID: requestID))
        }
    }
}
```

`AuthorizationNotFoundFault` and `CertificateNotFoundFault` don't need to be `CodedError`s because `ModifyDBInstanceOutputError` returns the right error code for them.

jbelkins: The extension I created is basically this switch statement in reverse... convert the enum case to a string instead of convert the string to an enum case.

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
jbelkins: I don't know the purpose either; in the past I have used underscore on private properties to help keep them separate from a related public property.

When we overhaul our error system...

Packages/ClientRuntime/Sources/Networking/CodedError.swift
jbelkins: Added

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
jbelkins: Added

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
jbelkins: Added

Packages/ClientRuntime/Sources/Networking/SdkError.swift
jbelkins: Added

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/CodedErrorGenerator.kt
jbelkins: Added

Packages/ClientRuntime/Sources/Networking/Http/UnknownHttpServiceError.swift
```diff
@@ -19,9 +21,24 @@ public struct UnknownHttpServiceError: HttpServiceError, Swift.Equatable {
...
 
 extension UnknownHttpServiceError {
-    public init(httpResponse: HttpResponse, message: String? = nil) {
+
+
+    /// Creates an `UnknownHttpServiceError` from a HTTP response.
+    /// - Parameters:
+    ///   - httpResponse: The `HttpResponse` for this error.
+    ///   - message: The message associated with this error.  May be omitted.
+    ///   - errorCode: The error code associated with this error.  May be omitted.
```
epau, line 30: Typically docs will say what the value defaults to.
```suggestion
    ///   - message: The message associated with this error. Defaults to nil
    ///   - errorCode: The error code associated with this error.  Defaults to nil
```

epau (approved):
Changes look great!

PR #492, opened 2022-12-12
fix: Fix encoding of nested maps when generating encoders

jbelkins:
## Issue \#
Fixes https://github.com/awslabs/aws-sdk-swift/issues/747

## Description of changes

With the introduction of the `AWSSecurityLake` service, an SDK compile error was encountered when compiling the generated `encode(to:)` method for a model containing this property:
```
        enableAllDimensions: [Swift.String:[Swift.String:[Swift.String]]]? = nil,
```

After investigation it was discovered that the SDK was not generating correct code for nested maps & collections, so the code generation for encoding nested Smithy Maps & Lists was corrected.

Correcting that code exposed a compile error on code for serializing & deserializing `@sparse` collections (i.e. arrays where elements my be nil, or dictionaries where values may be nil).  Codegen was also corrected for `@sparse` data types.

New local protocol tests are being added in https://github.com/awslabs/aws-sdk-swift/pull/779 to verify that these changes build correct Codable methods for nested and/or sparse collections.

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
@@ -7,7 +7,6 @@ package software.amazon.smithy.swift.codegen.integration.serde.json
 import software.amazon.smithy.model.knowledge.NullableIndex
 import software.amazon.smithy.model.shapes.BlobShape
 import software.amazon.smithy.model.shapes.CollectionShape
-import software.amazon.smithy.model.shapes.ListShape
 import software.amazon.smithy.model.shapes.MapShape
 import software.amazon.smithy.model.shapes.MemberShape
 import software.amazon.smithy.model.shapes.Shape
@@ -153,10 +152,7 @@ abstract class MemberShapeEncodeGenerator(
                 }
             }
             when (targetShape) {
-                is CollectionShape -> {
-                    renderEncodeListMember(targetShape, iteratorName, topLevelContainerName, level + 1)
-                }
-                is MapShape -> {
+                is CollectionShape, is MapShape -> {
                     renderEncodeListMember(targetShape, iteratorName, topLevelContainerName, level + 1)
                 }
                 else -> {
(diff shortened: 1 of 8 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -186,31 +171,31 @@ abstract class MemberShapeEncodeGenerator(
...
+                    keyName
                 )
-                renderEncodeMap(ctx, memberName, topLevelContainerName, targetShape.value, level)
+                renderEncodeMap(ctx, memberName, topLevelContainerName, targetShape, level)
             }
             else -> {
                 val isBoxed = ctx.symbolProvider.toSymbol(targetShape).isBoxed() && targetShape.hasTrait<SparseTrait>()
-                val keyEnumName = if (level == 0) ".$memberName" else "${ClientRuntimeTypes.Serde.Key}(stringValue: $dictKey${level - 1})"
-                if (isBoxed && level == 0) {
+                if (isBoxed) {
```
epau, line 193: why does `level` not matter here anymore?

epau (approved):
Just one question, otherwise lgtm

smithy-swift-codegen/src/test/kotlin/StructDecodeGenerationTests.kt
```diff
@@ -91,23 +90,26 @@ class StructDecodeGenerationTests {
                     var encodeContainer = encoder.container(keyedBy: CodingKeys.self)
                     if let intList = intList {
                         var intListContainer = encodeContainer.nestedUnkeyedContainer(forKey: .intList)
-                        for intlist0 in intList {
-                            try intListContainer.encode(intlist0)
+                        for integer0 in intList {
```
ganeshnj, line 93: if there is a list inside this, would `integer0` name collide?

Similarly for others. We must make sure names are unique.

ganeshnj (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
jbelkins: Since I'm not addressing the handling of boxed values in this PR, and since tests pass with or without the condition on `level` here, I am putting this `if` condition back the way it was.

smithy-swift-codegen/src/test/kotlin/StructDecodeGenerationTests.kt
jbelkins: If there is another `integer` it would be suffixed with `1`, `2`, `3`, etc. so name collisions would be prevented.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
jbelkins: Reverted.

ganeshnj (approved):

PR #497, opened 2022-12-21
feat: Updates CRT to 0.5.0+

epau:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/756

## Description of changes
Updates CRT dependency to 0.5.0 and fixes smithy-swift to make it compatible. 

## Scope


Packages/ClientRuntime/Sources/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -6,12 +6,25 @@
 //
 
 public struct DefaultSDKRuntimeConfiguration: SDKRuntimeConfiguration {
```
epau, line 8: We should rename this struct and the protocol to something more descriptive and without "SDK" in the name.
I'll file an issue to address that

Packages/ClientRuntime/Sources/Networking/Http/CRT/SDKDefaultIO.swift
```diff
@@ -14,32 +14,34 @@ import Darwin
 #endif
 
 public final class SDKDefaultIO {
```
epau, line 16: We should rename this type to something more descriptive and without "SDK" in the name. 
I'll file an issue to address this.

Packages/ClientRuntime/Sources/Networking/Http/SdkHttpRequest.swift
```diff
@@ -39,14 +39,14 @@ public class SdkHttpRequest {
 private let allowed = CharacterSet.alphanumerics.union(CharacterSet(charactersIn: "/_-.~"))
 
 extension SdkHttpRequest {
-    public func toHttpRequest(bufferSize: Int = 1024) -> HttpRequest {
```
epau, line 42: This `bufferSize` param was never used, so its now removed.

Packages/ClientRuntime/Sources/Networking/Http/SdkHttpRequest.swift
```diff
@@ -70,22 +70,20 @@ extension SdkHttpRequest: CustomDebugStringConvertible, CustomStringConvertible
...
 
 extension SdkHttpRequestBuilder {
-    public func update(from crtRequest: HttpRequest, originalRequest: SdkHttpRequest) -> SdkHttpRequestBuilder {
+    public func update(from crtRequest: HTTPRequest, originalRequest: SdkHttpRequest) -> SdkHttpRequestBuilder {
         headers = convertSignedHeadersToHeaders(crtRequest: crtRequest)
         methodType = originalRequest.method
         host = originalRequest.endpoint.host
         // TODO: remove extra slash if not needed
-        let pathAndQueryItems = URLComponents(string: crtRequest.path ?? "/")
+        let pathAndQueryItems = URLComponents(string: crtRequest.path)
```
epau, line 78: `path` is no longer optional, so the nil coalescing is no longer needed

Packages/ClientRuntime/Sources/Networking/Streaming/ByteStream.swift
```diff
@@ -9,7 +9,6 @@ import class Foundation.FileHandle
 
 public enum ByteStream {
     case buffer(ByteBuffer)
-    case reader(StreamReader)
```
epau, line 12: I couldn't find a good use case to use this `reader` variant. It appears it was never code generated and `ByteBuffer` is sufficient for streaming requests. Therefore, this variant is removed and the `DataStreamReader` class is also removed.

Please do let me know if you think we should keep it though.

Packages/ClientRuntime/Sources/Retries/SDKRetryer.swift
```diff
@@ -95,3 +100,16 @@ public class SDKRetryer: Retryer {
...
+
+extension AwsCommonRuntimeKit.RetryStrategy {
+    convenience init(options: RetryOptions, eventLoopGroup: EventLoopGroup) throws {
+       try self.init(
+            eventLoopGroup: eventLoopGroup,
+            initialBucketCapacity: options.initialBucketCapacity,
+            maxRetries: options.maxRetries,
+            backOffScaleFactor: options.backOffScaleFactor,
+            jitterMode: options.jitterMode.toCRTType(),
+            generateRandom: nil // we should pass in the options.generateRandom but currently it fails since the underlying closure is a c closure
```
epau, line 112: This is the same behavior currently today, where we end up just passing in `nil` for this closure since we cannot convert swift closures to c closures due to constraints around c closures. 

I've talked to @waahm7 about this and filed https://github.com/awslabs/aws-crt-swift/issues/127 for CRT to expose this as a swift closure. Once that is done, we can update this here to pass in the provided swift closure.

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -6,12 +6,25 @@
 //
 
 public struct DefaultSDKRuntimeConfiguration: SDKRuntimeConfiguration {
```
epau, line 8: We should rename this to something more descriptive and without the terns "Default" and "SDK".
I'll file an issue to address this

Sources/ClientRuntime/Config/SDKRuntimeConfiguration.swift
```diff
@@ -19,29 +19,3 @@ public protocol SDKRuntimeConfiguration {
     var retryer: SDKRetryer {get}
     var endpoint: String? {get set}
 }
-
-public extension SDKRuntimeConfiguration {
```
epau, line 23: This is a good example of when adding default implementations to protocols does not make sense. 
These defaults are all specific to the "default" or "crt based" configuration, and so we should be setting these values there and not at the protocol level.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -88,58 +81,64 @@ public class CRTClientEngine: HttpClientEngine {
...
+            response.body = .stream(.buffer(ByteBuffer(data: responseData)))
+            response.statusCode = makeStatusCode(stream)
 
             continuation.resume(returning: response)
         }
         return requestOptions
     }
+    
+    public func close() async {
+        // no-op
```
epau, line 142: There no longer is anything to close on the CRT side of things and so we just no-op here. 
We can remove the `close` function from the interface as well, and then delete it here but I left it in as it might sense to "close" other http engine implementations.

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -9,7 +9,6 @@ import class Foundation.FileHandle
 
 public enum ByteStream {
     case buffer(ByteBuffer)
-    case reader(StreamReader)
```
epau, line 12: The stream reader case was unused and seems unnecessary since everything can be handled via the ByteBuffer. Please let me know if you think this needs to stay in though.

Sources/ClientRuntime/Retries/SDKRetryer.swift
```diff
@@ -95,3 +93,16 @@ public class SDKRetryer: Retryer {
...
+
+extension AwsCommonRuntimeKit.RetryStrategy {
+    convenience init(options: RetryOptions, eventLoopGroup: EventLoopGroup) throws {
+       try self.init(
+            eventLoopGroup: eventLoopGroup,
+            initialBucketCapacity: options.initialBucketCapacity,
+            maxRetries: options.maxRetries,
+            backOffScaleFactor: options.backOffScaleFactor,
+            jitterMode: options.jitterMode.toCRTType(),
+            generateRandom: nil // we should pass in the options.generateRandom but currently it fails since the underlying closure is a c closure
```
epau, line 105: This maintains the existing behavior where we just pass in `nil` to generateRandom since it is a c closure. An issue has been filed with CRT to change the type exposed so that we can pass in a swift closure. Otherwise, we would need to expose the c closure type to consumers which doesn't make sense as it feels like an implementation detail leaking out.

Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift
```diff
@@ -14,32 +14,40 @@ import Darwin
...
 
 public final class SDKDefaultIO {
-    static weak var privateShared: SDKDefaultIO?
+    public let eventLoopGroup: EventLoopGroup
+    public let hostResolver: HostResolver
+    public let clientBootstrap: ClientBootstrap
+    public let tlsContext: TLSContext
+    public let logger: Logger
     
-    // TODO: revisit this and verify that it is thread safe.
```
jbelkins, line 19: Not the kind of thing that ought to be a "TODO" comment.

jbelkins (approved):
Approved since this gets us onto latest CRT, and the API changes they made do look like an improvement.

For the future, we should have an eye on how to separate CRT from smithy-swift altogether though.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
ganeshnj: would suggest deleting this entirely. We can do before go GA but not later.

Sources/ClientRuntime/Retries/RetryError.swift
```diff
@@ -14,7 +14,7 @@ public enum RetryError {
 }
 
 public extension RetryError {
-    func toCRTType() -> CRTRetryError {
+    func toCRTType() -> AwsCommonRuntimeKit.RetryError {
```
ganeshnj, line 17: isn't is `CommonRuntimeKit` now?

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -88,58 +81,64 @@ public class CRTClientEngine: HttpClientEngine {
...
+    ) throws -> HTTPRequestOptions {
         let response = HttpResponse()
-        let crtRequest = request.toHttpRequest(bufferSize: windowSize)
-        let streamReader: StreamReader = DataStreamReader()
+        let crtRequest = try request.toHttpRequest()
+        var responseData = Data()
+        
+        let makeStatusCode: (HTTPStream) -> HttpStatusCode = { stream in
+            guard
+                let statusCodeInt = try? stream.statusCode(),
```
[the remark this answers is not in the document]
epau: Is it odd? I had thought this was a pretty popular formatting style in the community and is the default way SwiftFormat will format multiline guard statements. 

Would be nice if Swift would prescribe some "standard format" one of these days...

Sources/ClientRuntime/Networking/Http/HttpBody.swift
```diff
@@ -14,29 +14,20 @@ public extension HttpBody {
     static var empty: HttpBody {
         .data(nil)
     }
-}
-
-extension HttpBody {
```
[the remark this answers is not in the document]
epau: iirc i needed public access to `toBytes` but maybe not. I'll double check and update accordingly

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithHttpPayload.kt
```diff
@@ -29,7 +29,7 @@ class HttpResponseTraitWithHttpPayload(
         val isBinaryStream =
             ctx.model.getShape(binding.member.target).get().hasTrait<StreamingTrait>() && target.type == ShapeType.BLOB
         writer.openBlock("if case .stream(let reader) = httpResponse.body {", "} else {") {
-            val extension = if (!isBinaryStream) ".toBytes().toData()" else ""
+            val extension = if (!isBinaryStream) ".toBytes().getData()" else ""
```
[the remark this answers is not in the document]
epau: Yeah i agree, a simple computed var named `data` would be nice here.

Sources/ClientRuntime/Retries/RetryError.swift
epau: Negative, [CommonRuntimeKit](https://github.com/awslabs/aws-crt-swift/blob/f3ae10c9a33bdc9d04246141ef0eb5840b369205/Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift) is a struct while `AwsCommonRuntimeKit` is the module. 

We should clean up these names though. It's odd to have something with a `Kit` suffix that isn't a module.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -1,94 +0,0 @@
-//
-// Copyright Amazon.com Inc. or its affiliates.
-// All Rights Reserved.
-//
-// SPDX-License-Identifier: Apache-2.0
-//
-import AwsCommonRuntimeKit
-import Foundation
-
-public class DataStreamReader: StreamReader {
```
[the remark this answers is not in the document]
epau: Nope. The one place this was used, `ByteBuffer` is sufficient https://github.com/awslabs/smithy-swift/pull/497/files#diff-9ae04687d7469a19d324ef46efc65eada692f0b62c0f49890bc0a6c27afd4358L111

There doesn't seem to be any added functionality provided by the `DataStreamReader` nor `StreamReader` (compared to `ByteBuffer`) and their purposes are unclear. We can simply rely on `ByteBuffer` for streaming data.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -1,20 +0,0 @@
-/*
- * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- * SPDX-License-Identifier: Apache-2.0.
- */
-
-import AwsCommonRuntimeKit
```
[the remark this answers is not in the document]
epau: See this comment https://github.com/awslabs/smithy-swift/pull/497#discussion_r1055682148

Sources/ClientRuntime/Networking/Http/HttpBody.swift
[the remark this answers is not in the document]
epau: Confirmed we need to access `toBytes` outside of this module. 

We no longer need to the `toAwsInputStream` because CRT changed how input streams are defined. Now its a protocol, `IStreamable` that `ByteBuffer` conforms to.

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
ganeshnj: I would push back on removing this as we don't the side effect yet. Let's keep this.

This can be cleanedup when doing streaming improvements.

ganeshnj (changes requested):

Sources/ClientRuntime/Retries/SDKRetryer.swift
```diff
@@ -26,8 +23,9 @@ public class SDKRetryer: Retryer {
     }
     
     public func scheduleRetry(token: RetryToken, error: RetryError) async throws -> RetryToken {
-        let token = try await crtRetryStrategy.scheduleRetry(token: token.crtToken, errorType: error.toCRTType())
-        return RetryToken(crtToken: token)
+        try await crtRetryStrategy.scheduleRetry(token: token.crtToken, errorType: error.toCRTType())
+        // remove this return after CRT updates scheduleRetry to return a token
+        return token
```
waahm7, line 28: I have already updated this.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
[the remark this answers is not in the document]
ganeshnj: Interesting, looks new to me.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -42,15 +42,21 @@ public class DataStreamReader: StreamReader {
...
-            buffer.put(byteBuffer, offset: offset, maxBytes: maxBytes)
-            if !rewind {
-                _availableForRead -= UInt(buffer.length)
-                offset += UInt(buffer.length)
+            let count = Int(maxBytes ?? availableForRead)
+            let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: count)
+            let buffer = UnsafeMutableBufferPointer(start: ptr, count: count)
+            let bytesRead = byteBuffer.read(buffer: buffer)
+            data = Data(buffer)
+            ptr.deallocate()
```
ganeshnj, line 52:
```suggestion
            let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: count)
            let buffer = UnsafeMutableBufferPointer(start: ptr, count: count)
            defer {
                  buffer.deallocate()
                  data.deallocate()
            }
            let bytesRead = byteBuffer.read(buffer: buffer)
            data = Data(buffer)
```

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -42,15 +42,21 @@ public class DataStreamReader: StreamReader {
...
+                        
+            if !rewind, let bytesRead = bytesRead {
+                _availableForRead -= UInt(truncatingIfNeeded: bytesRead)
```
ganeshnj, line 55: shouldn't it fail if truncation happens?

This could lead to loss of data.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -66,14 +72,15 @@ public class DataStreamReader: StreamReader {
...
+            let data = byteBuffer.getData() + buffer.getData()
+            byteBuffer = ByteBuffer(data: data)
+            _availableForRead += UInt(truncatingIfNeeded: buffer.length())
         }
     }
     
     public var contentLength: UInt? {
         withLockingClosure {
-            return byteBuffer.length
+            return UInt(truncatingIfNeeded: byteBuffer.length())
```
ganeshnj, line 83: same if truncation happens, we are giving wrong contentLength.

ganeshnj (changes requested):

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
epau: Yeah good call out, I'll remove the `truncatingIfNeeded`. 
When we tackle streaming improvements, we may want to update the interface to use `UIInt64` as well

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
epau: hmm can you explain this change? from my understand, it is only required that we `deallocate` the `UnsafeMutablePointer` (`ptr`).

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -42,15 +42,21 @@ public class DataStreamReader: StreamReader {
...
-            buffer.put(byteBuffer, offset: offset, maxBytes: maxBytes)
-            if !rewind {
-                _availableForRead -= UInt(buffer.length)
-                offset += UInt(buffer.length)
+            let count = Int(maxBytes ?? availableForRead)
+            let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: count)
+            defer { ptr.deallocate() }
+            let buffer = UnsafeMutableBufferPointer(start: ptr, count: count)
+            let bytesRead = byteBuffer.read(buffer: buffer)
+            data = Data(buffer)
```
waahm7, line 52: We can replace all of this with data and a safer version which doesn't involve allocation/deallocation. Something like the following
```suggestion
            let data = Data(count: count)
            bufferData.withUnsafeMutableBytes { buffer in 
                 byteBuffer.read(buffer: buffer)
            }
```

waahm7: In general, the allocation & deallocation should only happen if we want to manually control lifetime of something. In most cases, we can avoid that.

waahm7 (changes requested):

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
epau: 🤯 
awesome! will update to this

waahm7 (approved):

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
ganeshnj: my bad, didn't realize it was BufferPointer, makes sense.

ganeshnj (approved):

epau:
Waiting until after next release to merge, since this includes a big update from CRT

PR #504, opened 2023-01-05
Adds FoundationTypes and adds function to import individual types from modules

epau:
## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/FoundationTypes.kt
```diff
@@ -0,0 +1,12 @@
+package software.amazon.smithy.swift.codegen
```
ganeshnj, line 1: header missing

ganeshnj (approved):

PR #510, opened 2023-01-11
fix: change socketOptions to mutable to allow mutation on iOS and watchOS

ganeshnj:
## Issue \#

On iOS and watchOS, SDK fails to build.

## Scope


epau (approved):
nice find!

PR #517, opened 2023-01-23
fix: two Swift compile warnings

jbelkins:
## Description of changes
Fixes two different Swift compile warnings thrown when building the AWS SDK.
- `SdkError` no longer conditionally casts `ClientError` to `WaiterTypedError` since `WaiterTypedError` always conforms.
- The `encode(_:)` function on `XMLEncoder` is changed from `open` to `public` to fix `Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead`

After this change, the only compile warnings remaining on the SDK service clients are:
- Deprecated model properties
- The query items property is written as a literal `[]` when no items will be added to it (was going to fix this here, but it needs further investigation)

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpQueryItemProvider.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpQueryItemProvider.kt
@@ -73,13 +73,9 @@ class HttpQueryItemProvider(
         writer.openBlock("extension \$N: \$N {", "}", inputSymbol, ClientRuntimeTypes.Middleware.Providers.QueryItemProvider) {
             writer.openBlock("public var queryItems: [\$N] {", "}", ClientRuntimeTypes.Core.URLQueryItem) {
                 writer.openBlock("get throws {", "}") {
-                    if (queryLiterals.isNotEmpty() || queryBindings.isNotEmpty()) {
-                        writer.write("var items = [\$N]()", ClientRuntimeTypes.Core.URLQueryItem)
-                        generateQueryItems()
-                        writer.write("return items")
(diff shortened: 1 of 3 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpQueryItemProvider.kt
```diff
@@ -73,9 +73,13 @@ class HttpQueryItemProvider(
...
-                    writer.write("var items = [\$N]()", ClientRuntimeTypes.Core.URLQueryItem)
-                    generateQueryItems()
-                    writer.write("return items")
+                    if (queryLiterals.isNotEmpty() || queryBindings.isNotEmpty()) {
+                        writer.write("var items = [\$N]()", ClientRuntimeTypes.Core.URLQueryItem)
+                        generateQueryItems()
+                        writer.write("return items")
+                    } else {
+                        writer.write("return []")
+                    }
```
ganeshnj, line 82: there should be a test on this.

ganeshnj (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpQueryItemProvider.kt
jbelkins: Good call, because in writing the test, I discovered that there is potentially a bug with the query item provider logic.

I am reverting this change in this PR and will investigate further; this fix will be filed in a separate PR once I have diagnosed the issue.

epau (approved):
love it

ganeshnj (approved):

PR #518, opened 2023-01-31
chore: bump aws-crt-swift to 0.5.6

ganeshnj:
## Issue \#

Fixes https://github.com/awslabs/aws-sdk-swift/issues/835

Related https://github.com/awslabs/aws-sdk-swift/pull/862

### Motivation

We have new aws-crt-swift version available which has breaking changes.

### Changes

- Pass bootstrap parameter
- HTTP Client in CRT now accepts updated callbacks which no longer has stream. Hence, updated the callback uses to reflect the changes.

### Result

- No functional changes
- Verified S3 list bucket and get object operation
- 
## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
@@ -103,39 +103,38 @@ public class CRTClientEngine: HttpClientEngine {
         let crtRequest = try request.toHttpRequest()
         let streamReader: StreamReader = DataStreamReader()
 
-        let makeStatusCode: (HTTPStream) -> HttpStatusCode = { stream in
-            guard
-                let statusCodeInt = try? stream.statusCode(),
-                let statusCode = HttpStatusCode(rawValue: statusCodeInt)
-            else { return .notFound }
-            return statusCode
-        }
-
-        let requestOptions = HTTPRequestOptions(request: crtRequest) { [self] (stream, _, httpHeaders) in
-            logger.debug("headers were received")
-            response.statusCode = makeStatusCode(stream)
-            response.headers.addAll(httpHeaders: httpHeaders)
-        } onIncomingHeadersBlockDone: { [self] (stream, _) in
-            logger.debug("header block is done")
-            response.statusCode = makeStatusCode(stream)
-        } onIncomingBody: { [self] (stream, data) in
-            logger.debug("incoming data")
-            response.statusCode = makeStatusCode(stream)
-            let byteBuffer = ByteBuffer(data: data)
-            streamReader.write(buffer: byteBuffer)
-        } onStreamComplete: { [self] (stream, error) in
-            logger.debug("stream completed")
-            if let error = error, error.code != 0 {
-                logger.error("Response encountered an error: \(error)")
-                continuation.resume(throwing: CommonRunTimeError.crtError(error))
-                return
+        let makeStatusCode: (UInt32) -> HttpStatusCode = { statusCode in
+            guard let httpStatusCode = HttpStatusCode(rawValue: Int(statusCode)) else {
+                return .notFound
             }
+            return httpStatusCode
+         }
 
-            response.body = .stream(.reader(streamReader))
-            response.statusCode = makeStatusCode(stream)
-
-            continuation.resume(returning: response)
+        let requestOptions = HTTPRequestOptions(request: crtRequest) { statusCode, headers in
+            response.statusCode = makeStatusCode(statusCode)
+            response.headers.addAll(headers: Headers(httpHeaders: headers))
+        } onResponse: { statusCode, headers in
+            response.statusCode = makeStatusCode(statusCode)
+            response.headers.addAll(headers: Headers(httpHeaders: headers))
+        } onIncomingBody: { bodyChunk in
+            let byteBuffer = ByteBuffer(data: bodyChunk)
+            streamReader.write(buffer: byteBuffer)
+        } onTrailer: { headers in
+            response.headers.addAll(headers: Headers(httpHeaders: headers))
+        } onStreamComplete: { result in
+            streamReader.hasFinishedWriting = true
+            switch result {
+            case .success(let statusCode):
+                response.statusCode = makeStatusCode(statusCode)
+                continuation.resume(returning: response)
+            case .failure(let error):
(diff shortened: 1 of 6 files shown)


Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -70,18 +70,19 @@ extension SdkHttpRequest: CustomDebugStringConvertible, CustomStringConvertible
...
         host = originalRequest.endpoint.host
-        let pathAndQueryItems = URLComponents(string: crtRequest.path)
-        path = pathAndQueryItems?.path ?? "/"
-        queryItems = pathAndQueryItems?.percentEncodedQueryItems ?? [URLQueryItem]()
-
+        if let crtRequest = crtRequest as? HTTPRequest {
+            let pathAndQueryItems = URLComponents(string: crtRequest.path)
+            path = pathAndQueryItems?.path ?? "/"
+            queryItems = pathAndQueryItems?.percentEncodedQueryItems ?? [URLQueryItem]()
+        }
```
epau, line 81: Should we do something instead of silently no-oping when the `crtRequest` is not a HTTPRequest? 

Would it make sense to to still set the `path` and `queryItems` since we have handling for nil values? Or perhaps we should fail? 

```suggestion
        let pathAndQueryItems = (crtRequest as? HTTPRequest).map { URLComponents(string: $0.path) }
        path = pathAndQueryItems?.path ?? "/"
        queryItems = pathAndQueryItems?.percentEncodedQueryItems ?? [URLQueryItem]()
```

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -70,18 +70,19 @@ extension SdkHttpRequest: CustomDebugStringConvertible, CustomStringConvertible
 }
 
 extension SdkHttpRequestBuilder {
-    public func update(from crtRequest: HTTPRequest, originalRequest: SdkHttpRequest) -> SdkHttpRequestBuilder {
+    public func update(from crtRequest: HTTPRequestBase, originalRequest: SdkHttpRequest) -> SdkHttpRequestBuilder {
```
epau, line 73: Nit: Would be nice to add docs while we're here

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -103,39 +103,38 @@ public class CRTClientEngine: HttpClientEngine {
...
-                continuation.resume(throwing: CommonRunTimeError.crtError(error))
-                return
+        } onTrailer: { headers in
+            response.headers.addAll(headers: Headers(httpHeaders: headers))
+        } onStreamComplete: { result in
+            streamReader.hasFinishedWriting = true
+            switch result {
+            case .success(let statusCode):
+                response.statusCode = makeStatusCode(statusCode)
+                continuation.resume(returning: response)
```
epau, line 129: Should we also set `response.body `here, before resuming the continuation? 
Otherwise I think the `response` returned to the awaiting task won't have the correct value for `response.body`.

Tests/ClientRuntimeTests/ClientRuntimeTests/NetworkingTests/ClientErrorTests.swift
```diff
@@ -101,7 +101,7 @@ class ClientErrorTests: XCTestCase {
     }
 
     func test_waiterErrorType_returnsNilForCRTError() async throws {
-        let crtError = CRTError(code: 2)
+        let crtError = CommonRunTimeError.crtError(.init(code: 2))
```
jbelkins, line 104: what is `2`?  I know it likely doesn't matter for this test as long as it results in a non-null error, but just want some context for the value used

jbelkins (changes requested):
One Q to resolve above the other comments

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
ganeshnj: Was thinking around this for sometime but the current way seems correct.

We have 

https://github.com/awslabs/aws-crt-swift/blob/0.5.6/Source/AwsCommonRuntimeKit/http/HTTPRequestBase.swift#L7

The  H1 and H2 has their own implementations

H1

https://github.com/awslabs/aws-crt-swift/blob/0.5.6/Source/AwsCommonRuntimeKit/http/HTTPRequest.swift#L8

H2 

https://github.com/awslabs/aws-crt-swift/blob/0.5.6/Source/AwsCommonRuntimeKit/http/HTTPRequest.swift#L62

What we must do always if the passed object is neither HTTPRequest nor HTTP2Request, it fails. This is a developer error here.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
ganeshnj: ++

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -70,18 +70,29 @@ extension SdkHttpRequest: CustomDebugStringConvertible, CustomStringConvertible
...
-        let pathAndQueryItems = URLComponents(string: crtRequest.path)
-        path = pathAndQueryItems?.path ?? "/"
-        queryItems = pathAndQueryItems?.percentEncodedQueryItems ?? [URLQueryItem]()
-
+        if let crtRequest = crtRequest as? HTTPRequest {
+            let pathAndQueryItems = URLComponents(string: crtRequest.path)
+            path = pathAndQueryItems?.path ?? "/"
+            queryItems = pathAndQueryItems?.percentEncodedQueryItems ?? [URLQueryItem]()
+        } else if crtRequest as? HTTP2Request != nil {
+            assertionFailure("HTTP2Request not supported")
```
ganeshnj, line 88: we will update this during streaming support

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
epau: I see now, this makes sense.

epau (approved):

jbelkins (approved):

PR #524, opened 2023-02-10
refactor: use Stream to send request body for H2

ganeshnj:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/870

### Motivation

Currently, the SDK uses `ByteBuffer` to send request body. This is not efficient because it requires the user to copy the data into a `ByteBuffer`. For Event Streams, the SDK needs to send a stream of Data chunks.

### Changes

Introduces a new async method in StreamReader protocol `func read(upToCount count: Int?) async throws -> Data?` which allows async reading of the request body input.

Add `AsyncThrowingStreamReader` which wraps an `AsyncThrowingStream<Data, Error>`.  This is used to send Event Stream messages.

### Result

The SDK can now send request body using StreamReader for H2. Working of the normal payload is not affected.

## Scope


Sources/ClientRuntime/Networking/Http/HttpBody.swift
```diff
@@ -26,13 +27,37 @@ public extension HttpBody {
...
         switch self {
         case let .data(data):
             return data?.isEmpty ?? true
         case let .stream(stream):
-            return stream.toBytes().getData().isEmpty
+            switch stream {
+            case .buffer(let buffer):
+                return buffer.length() == 0
+            case .reader(let reader):
+                return reader.availableForRead != 0
```
epau, line 59: Nit: It'd be nice to define an `isEmpty` convenience on Stream. It would make our API more consistent with Swift and would make this snippet self-documenting. 

The snippet here would simply turn into: 
```swift 
return stream.isEmpty
```

Sources/ClientRuntime/Networking/Http/HttpContent.swift
```diff
@@ -0,0 +1,33 @@
...
+    public init(stream: StreamReader) {
+        self.stream = stream
+    }
+
+    public func seek(offset: Int64, streamSeekType: AwsCommonRuntimeKit.StreamSeekType) throws {
+        switch streamSeekType {
+        case .begin:
+            try stream.seek(offset: Int(offset))
+        case .end:
+            fatalError("This operation is not supported")
```
epau, line 24: Can we provide a more descriptive error message here? While we will hit this during debugging, stack traces aren't always accurate in production and it'd be good to easily identify this case from the error message.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+
+/// A wrapper around a StreamReader that conforms to the IStreamable protocol
+/// This is used to pass the StreamReader to the CRT HttpMessage
+public class HttpContent: IStreamable {
+    internal let stream: StreamReader
```
epau, line 13: We haven't been explicitly defining methods as `internal` since `internal` is the default. We should either remove `internal` here or file an issue to update all internal APIs to explicitly define `internal`.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
```diff
@@ -0,0 +1,33 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+
+/// A wrapper around a StreamReader that conforms to the IStreamable protocol
+/// This is used to pass the StreamReader to the CRT HttpMessage
+public class HttpContent: IStreamable {
```
epau, line 12: Can we add docs for all public methods in this class?

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
```diff
@@ -0,0 +1,106 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import AwsCommonRuntimeKit
+
+/// A `StreamReader` that reads `AsyncThrowingStream<Data, Error>`
+/// Event Stream messages are sent as a series of `Data` chunks
+public class AsyncThrowingStreamReader: StreamReader {
```
epau, line 13: Can we provide more descriptive error messages for each `fatalError` in this class? 

Additionally, can we add comments to methods that have _real_ implementations? (No need to add docs for methods that simply fatal error)

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -4,47 +4,62 @@
...
 import Foundation
 
+/// A `StreamReader` that reads from a `ByteBuffer`
+/// It allows thread safe reading, writing and seeking
 public class DataStreamReader: StreamReader {
     private var _availableForRead: UInt
-
     public var availableForRead: UInt {
-        return _availableForRead
+        get {
```
epau, line 16: While we're here, can we convert `DataStreamReader` to an actor and remove the locks? It would simplify the implementation, make it less bug prone, and keeps the code updated with modern swift practices.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
```diff
@@ -4,47 +4,62 @@
...
         set {
-            withLockingClosure {
+            lock.withLockingClosure {
                 _hasFinishedWriting = newValue
             }
         }
     }
 
-    private var byteBuffer: ByteBuffer
+    internal var byteBuffer: ByteBuffer
```
epau, line 42: Similar comment as above regarding `internal` usage.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -7,14 +7,31 @@ import AwsCommonRuntimeKit
 import class Foundation.FileHandle
 import class Foundation.FileManager
 
+/// Protocol for reading and writing to a stream.
 public protocol StreamReader: AnyObject {
```
epau, line 11: ooc, why the AnyObject restriction?

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -7,14 +7,31 @@ import AwsCommonRuntimeKit
...
+
+    /// Seek to a specific offset in the stream.
+    /// - Parameter offset: The offset to seek to. This is relative to the beginning of the stream.
+    func seek(offset: Int) throws
+
+    /// Close the stream.
+    /// - Parameter error: The error that caused the stream to close.
     func onError(error: ClientError)
+
+    /// Write a buffer to the stream.
```
epau, line 34:
```suggestion
    /// Writes a buffer to the stream.
```

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -7,14 +7,31 @@ import AwsCommonRuntimeKit
...
     func read(maxBytes: UInt?, rewind: Bool) -> ByteBuffer
-    func seek(offset: Int)
+
+    /// Seek to a specific offset in the stream.
+    /// - Parameter offset: The offset to seek to. This is relative to the beginning of the stream.
+    func seek(offset: Int) throws
+
+    /// Close the stream.
+    /// - Parameter error: The error that caused the stream to close.
     func onError(error: ClientError)
```
epau, line 32: Nit: Suggest renaming this function to match the intention as defined by the comment. 

```suggestion
    /// Closes the stream when an error occurs.
    /// - Parameter error: The error that caused the stream to close.
    func closeWithError(_ error: ClientError)
```

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -7,14 +7,31 @@ import AwsCommonRuntimeKit
 import class Foundation.FileHandle
 import class Foundation.FileManager
 
+/// Protocol for reading and writing to a stream.
 public protocol StreamReader: AnyObject {
+    /// The number of bytes that are available for reading.
     var availableForRead: UInt { get }
+
+    /// Whether or not the stream has finished writing.
+    /// This is used to determine when to stop reading from the stream.
```
epau, line 16:
```suggestion
    /// Returns true if the stream has finished writing, otherwise returns false
```

The line `/// This is used to determine when to stop reading from the stream.` is a bit misleading here since it's an implementation detail and this is the protocol. IMO this comment should only be added to the implementation.

Sources/ClientRuntime/PrimitiveTypeExtensions/NSLock+Extensions.swift
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension NSLock {
```
epau, line 10: I'm hoping we can remove this in favor of using `actor`

Sources/ClientRuntime/Util/Async.swift
```diff
@@ -0,0 +1,53 @@
...
+        do {
+            let val = try await f()
+            box.result = .success(val)
+        } catch {
+            box.result = .failure(error)
+        }
+        semaphore.signal()
+    }
+    semaphore.wait()
+    return try box.result!.get()
```
epau, line 36: Can we just use Result here and remove `BoxedThrowingResult`? IMO its simpler 

```suggestion
    var result: Result<ResultType, Error>? = nil
    let semaphore = DispatchSemaphore(value: 0)
    Task {
        do {
            let val = try await f()
            result = .success(val)
        } catch {
            result = .failure(error)
        }
        semaphore.signal()
    }
    semaphore.wait()
    return try result!.get()
```

Sources/ClientRuntime/Util/Async.swift
```diff
@@ -0,0 +1,53 @@
...
+""")
+public func _unsafeWait<ResultType>(_ f: @escaping () async -> ResultType) -> ResultType {
+    let box = BoxedResult<ResultType>()
+    let semaphore = DispatchSemaphore(value: 0)
+    Task {
+        box.result = await f()
+        semaphore.signal()
+    }
+    semaphore.wait()
+    return box.result!
```
epau, line 52: Can just use the `ResultType` here and remove `BoxedResult`? Similar to the above comment, imo this is simpler. 

```suggestion
    var result: ResultType? = nil
    let semaphore = DispatchSemaphore(value: 0)
    Task {
        result = await f()
        semaphore.signal()
    }
    semaphore.wait()
    return result!
```

Sources/ClientRuntime/Util/Async.swift
```diff
@@ -0,0 +1,53 @@
...
+}
+
+private class BoxedResult<ResultType> {
+    var result: ResultType?
+}
+
+/// Unsafely awaits a throwing async function from a synchronous context.
+@available(*, deprecated, message: """
+    This is a temporary workaround for the lack of async support.
+    It will be migrated to structured concurrency in the future
```
epau, line 21: When "in the future" do you think this will be migrated to structured concurrency? What's preventing us from doing it now?

Tests/ClientRuntimeTests/NetworkingTests/AsyncThrowingStreamReaderTests.swift
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import XCTest
+@testable import ClientRuntime
+
+final class AsyncThrowingStreamReaderTests: XCTestCase {
+    func testRead() {
```
epau, line 10: Can we add another test that verifies the locking behavior by spinning up multiple concurrent calls to `read` and verifying the output is as expected?

Tests/ClientRuntimeTests/NetworkingTests/DataStreamReaderTests.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import ClientRuntime
+
+final class DataStreamReaderTests: XCTestCase {
+    func testRead() throws {
+        let sut = DataStreamReader(data: Data([0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A]))
```
epau, line 13: Nit: Can we move this out of the test so that it can be reused?

Tests/ClientRuntimeTests/NetworkingTests/StreamContentTests.swift
```diff
@@ -0,0 +1,68 @@
...
+
+final class StreamContentTests: XCTestCase {
+
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
```
epau, line 20: lets remove if unimplemented

Tests/ClientRuntimeTests/NetworkingTests/DataStreamReaderTests.swift
```diff
@@ -0,0 +1,76 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import ClientRuntime
+
+final class DataStreamReaderTests: XCTestCase {
```
epau, line 11: Can we either provide more descriptive names for the test functions? Perhaps following a format like "test_whenABC_thenXYZ"
Or add comments that describe the behavior the test is verifying

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public func read(maxBytes: UInt?, rewind: Bool = false) -> AwsCommonRuntimeKit.ByteBuffer {
+        if rewind {
+            fatalError("Rewind is not supported with async streams")
+        }
+
+        return lock.withLockingClosure {
+            return _unsafeWait {
+                return await self.read(maxBytes: maxBytes)
```
jbelkins, line 53: See warning in the [NSLock docs](https://developer.apple.com/documentation/foundation/nslock):

> The NSLock class uses POSIX threads to implement its locking behavior. When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior.

It seems likely that after `await` this block will continue execution on a different thread.

jbelkins (changes requested):

Sources/ClientRuntime/PrimitiveTypeExtensions/NSLock+Extensions.swift
jbelkins: Seconded, see my comment above on the use of `async` while locked.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
ganeshnj: sure

Sources/ClientRuntime/Networking/Http/HttpContent.swift
ganeshnj: I thought without docs, protocol methods automatically gets protocol docs but no :(

ganeshnj: done

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
ganeshnj: done.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
ganeshnj: almost forgot there are Actors now. 100%

ganeshnj: done

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
ganeshnj: if I have to guess, we don't want to copy data, and by mistake we could end up implementing it via a struct which is going to be a bad idea.

Backing/internal stream must always be a reference type.

Sources/ClientRuntime/Util/Async.swift
ganeshnj: underlying CRT APIs. I'm going to create issues for them.

Sometimes protocol properties like `CustomDebugStringConvertible` which can't be made async implementation.

Sources/ClientRuntime/Util/Async.swift
ganeshnj: Thats because we need a type that can be manipulated outside concurrently. Result is enum which is not possible to be altered if declared outside the concurrent block ie Task.

Sources/ClientRuntime/Util/Async.swift
ganeshnj: nope, same as above.

Tests/ClientRuntimeTests/NetworkingTests/DataStreamReaderTests.swift
ganeshnj: I moved the data part but not the reader as we don't one test interfere another.

Tests/ClientRuntimeTests/NetworkingTests/StreamContentTests.swift
ganeshnj: done

Tests/ClientRuntimeTests/NetworkingTests/DataStreamReaderTests.swift
ganeshnj: these tests check multiple read, write seek calls so harder to name them, but I have added comments what individual step does.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
ganeshnj: how about just `close`?

Sources/ClientRuntime/PrimitiveTypeExtensions/NSLock+Extensions.swift
ganeshnj: fixed.

Tests/ClientRuntimeTests/NetworkingTests/AsyncThrowingStreamReaderTests.swift
ganeshnj: now we use actor, we don't want to test actor.

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
ganeshnj: no longer applicable.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
ganeshnj: for now removing internal but would like to discuss what team think on adding it explicitly.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
ganeshnj: actually now thinking more, I would like it to be as it is. Because we have exposed this in Input models.

We may do bigger refactor for customer facing structs.

Sources/ClientRuntime/Networking/Streaming/DataStreamReader.swift
ganeshnj: Dropped the idea to use actor, because we have exposed StreamReader interface in public models which confirms to Codable.

Coddle is not async protocol and it is not possible to confirm it without a hacky solution.

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
jbelkins: How is this no longer applicable?  We still make an `async` call while holding a lock, and there is no guarantee that execution continues on the same thread after resuming.

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
```diff
@@ -0,0 +1,106 @@
...
+                }
+            } catch {
+                // TODO: handle error
+                fatalError("TODO: handle error \(error)")
+            }
+        }
+        return ByteBuffer(data: data)
+    }
+
+    public func seek(offset: Int) throws {
```
jbelkins, line 95: The fact that so many methods from the `StreamReader` protocol fail if called suggests that this protocol is not the right abstraction for an asynchronous stream, or that the protocol needs to be broken down into narrower sub-protocols that certain sub-types of streams may adopt.

jbelkins (changes requested):
I feel like this PR is hammering the async round peg into a synchronous square hole, adopting ill-suited existing conventions to new, async means of passing data.

Examples of things that suggest we need to adapt more fully to the async model of providing data:
- Async stream readers call `fatalError` for a large chunk of the methods on the `StreamReader` protocol.
- We use semaphores to block a thread while asynchronously collecting the data from a stream before continuing.

I would prefer to update to async-friendly interfaces in conjunction with this change.

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
ganeshnj: I made the comment after converting StreamReader to actor which didn't work in the end.

ganeshnj: Also, why it needs to be guaranteed that execution must continue on the same thread?

Sources/ClientRuntime/Util/Async.swift
epau: Ahh thats right

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
epau: While yes, acquiring the lock and releasing the lock must happen on the same thread, I don't see any threading issues here. 

`_unsafeWait` will execute the async closure in a task on some random thread, but `_unsafeAwait` will return the value in the same thread that it was called on. This is because `_unsafeWait` uses a semaphore to block the calling thread until the async task finishes.

Sources/ClientRuntime/Networking/Http/AsyncThrowingStreamReader.swift
epau: I agree with Josh here, although this might be out of scope for this PR and this feature, but we should create an issue to fix this up.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
epau: Definitely not opposed to it! Just want consistency

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
epau: Gotcha, I think this is unnecessary and is an implementation detail leaking to the interface. It is quite possible to conform to this protocol without needing a class (a struct could always have a property that is a reference type where mutated values are stored). If this protocol is going to be very prescriptive about the implementation, then it seems like its too rigid to be a protocol and instead should be a concrete object that allows for configuration/customization. 

Related to Josh's comment earlier regarding StreamReader being the wrong abstraction, we should tackle this in follow-up work.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
epau: Fine to leave it for now since there are a few comments regarding fixing this up in follow-up work.

Tests/ClientRuntimeTests/NetworkingTests/DataStreamReaderTests.swift
epau: I see comments have been added to some, which are very helpful thank you! Please add to each test, since most of the tests are a bit unclear in what the expected behavior is and so I think comments would go a long way improving their maintainability moving forward. 

Two examples of improvements:

```swift
// Tests that `availableForRead` returns the correct value and
// that it decrements accordingly after each call to `read()`
func testAvailableForRead() 
```

```swift
// Tests that `contentLength` returns the correct value and 
// that the `contentLength` doesn't change after `read()` is executed.
func testContentLength()
```

epau (approved):
Thanks for testing out the actor route! 
Just a nit comment asking for more comments on tests and a request to open an issue to improve the `StreamReader` abstraction. 

Otherwise looks good to me.

Sources/ClientRuntime/Networking/Streaming/StreamReader.swift
```diff
@@ -7,14 +7,53 @@ import AwsCommonRuntimeKit
...
+    /// Write a buffer to the stream.
+    /// - Parameter buffer: The buffer to write to the stream.
     func write(buffer: ByteBuffer)
 }
+
+extension StreamReader {
+    public var isEmpty: Bool {
+        // swiftlint:disable empty_count
+        return availableForRead == 0
+        // swiftlint:enable empty_count
```
epau, line 51: Does the linter complain here? `availableForRead` is an `UInt` and so i don't think it even has a `isEmpty`.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -123,15 +151,44 @@ public class CRTClientEngine: HttpClientEngine {
...
+                        continue
+                    }
+                    try await writeData(data: data, endOfStream: false)
+                }
+            }
+        case .none:
+            break
+        }
+
+        try await writeData(data: .init(), endOfStream: true)
```
epau, line 193: Can you add a comment to describe why this is necessary?

epau:
Closing this as it has been replaced by implementation in https://github.com/awslabs/smithy-swift/pull/537

PR #530, opened 2023-02-23
refactor: streams

ganeshnj:
### Motivation

The SDK doesn't stream request and responses rather it buffers them in memory. This is a problem for large requests and responses.

### Changes

Introduced `HttpContent` which conforms to `IStreamable` (a CRT protocol) required for streaming requests. `HttpContent` can be created from `HttpBody` which can be either a Data or a Stream.

Introduced `Stream` protocol and number of implementations for different stream types.

- `BufferedStream` - allows thread safe reading and writing while keeping only a fixed unread amount of data in memory.
- `FileStream` - allows reading and writing from/to a file.
- `CacheStream` -  allows thread safe reading and writing on base stream. This is useful for streams that don't support seeking.

### Result

Now we can stream requests and responses.

## Scope


Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+//
+
+import Foundation
+
+/// A `Stream` implementation that buffers data in memory.
+/// The buffer size depends on the amount of data written and read.
+/// Note: This class is thread-safe.
+public class BufferedStream: Stream {
+
+    /// The length of the stream, if known.
```
epau, line 15: The underlying buffer is not nullable and so what case(s) is the length unknown?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+    public var length: Int?
+
+    /// The current position in the stream.
+    public var position: Data.Index
```
epau, line 19: Should this be `private(set)` so that it can only be modified via operations on the stream (read, write, seek etc)?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Whether the stream is empty.
+    public var isEmpty: Bool {
+        return buffer.isEmpty
+    }
+
+    /// Whether the stream is seekable.
```
epau, line 26: Imo we should prefer a format of `Returns a if x,y,z otherwise returns b` for doc comments on simple booleans. 
Such as `Returns true if the stream is seekable, otherwise returns false. `
Imo this is a bit more clear.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// Although this stream is seekable, seeking to a position that is not in the buffer will cause the stream to throw an error.
+    public var isSeekable: Bool = false
+
+    var isClosed: Bool
+
+    var buffer = Data()
```
epau, line 32: Even though this is internal, should this be `private(set)` to prevent code outside of this class, from modifying this value?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+    let lock = NSLock()
```
epau, line 33: Can we make this private?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// Whether the stream is empty.
+    public var isEmpty: Bool {
+        return buffer.isEmpty
+    }
+
+    /// Whether the stream is seekable.
+    /// Although this stream is seekable, seeking to a position that is not in the buffer will cause the stream to throw an error.
+    public var isSeekable: Bool = false
+
+    var isClosed: Bool
```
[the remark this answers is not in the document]
epau: Additionally, should we expose this publicly? I find it a bit odd for a property to be injected in the initializer but not exposed on the type (at least exposed for reading).

Sources/ClientRuntime/Networking/Http/HttpBody.swift
```diff
@@ -4,23 +4,50 @@
  */
 import AwsCommonRuntimeKit
 
-public enum HttpBody: Equatable {
+public enum HttpBody {
     case data(Data?)
-    case stream(ByteStream)
```
ganeshnj, line 9: can we remove ByteStream completely?

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -10,20 +10,27 @@ public struct DeserializeMiddleware<Output: HttpResponseBinding,
...
+            if (200..<300).contains(response.httpResponse.statusCode.rawValue) {
+                let output = try Output(httpResponse: copiedResponse.httpResponse, decoder: decoder)
+                copiedResponse.output = output
+                return copiedResponse
+            } else {
+                /// if the response is a stream, we need to cache the stream so that it can be read again
+                /// error deserialization reads the stream multiple times to first deserialize the protocol error
+                /// eg. [RestJSONError](https://github.com/awslabs/aws-sdk-swift/blob/d1d18eefb7457ed27d416b372573a1f815004eb1/Sources/Core/AWSClientRuntime/Protocols/RestJSON/RestJSONError.swift#L38,
+                /// and then the service error eg. [AccountNotFoundException](https://github.com/awslabs/aws-sdk-swift/blob/d1d18eefb7457ed27d416b372573a1f815004eb1/Sources/Services/AWSCloudTrail/models/Models.swift#L62)
+                if case let .stream(stream) = copiedResponse.httpResponse.body {
```
ganeshnj, line 29: check is seekable, no need of caching stream.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /// Reads all data from the stream.
+    /// It uses `read(upToCount:)` to read data in chunks.
+    /// - Returns: Data read from the stream, or nil if the stream is closed and no data is available.
+    public func readToEnd() throws -> Data? {
+        var data = Data()
+
+        while let next = try read(upToCount: Int.max) {
```
ganeshnj, line 77: use recursive lock

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,124 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A `Stream` implementation that buffers data in memory.
+/// The buffer size depends on the amount of data written and read.
```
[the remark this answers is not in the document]
ganeshnj: correct.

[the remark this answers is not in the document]
ganeshnj: updated docs on this.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: yes updated.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: +1

Sources/ClientRuntime/Networking/Http/HttpBody.swift
ganeshnj: not yet and probably will not, as we talked on the call.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: done

ganeshnj: perhaps not buffer (tests)  but made if private(set)

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -101,37 +101,49 @@ public class CRTClientEngine: HttpClientEngine {
...
         let makeStatusCode: (UInt32) -> HttpStatusCode = { statusCode in
             HttpStatusCode(rawValue: Int(statusCode)) ?? .notFound
          }
 
         let requestOptions = HTTPRequestOptions(request: crtRequest) { statusCode, headers in
+            self.logger.debug("headers were received")
             response.statusCode = makeStatusCode(statusCode)
             response.headers.addAll(headers: Headers(httpHeaders: headers))
         } onResponse: { statusCode, headers in
+            self.logger.debug("header block is done")
```
epau, line 115: Nit: 
```suggestion
            self.logger.debug("Headers received")
```

epau, line 115: What is the difference between this callback and the one above? 
One is "headers were received" and the other is "header block is done". Can we provide a better log string for the latter? I'm unsure what a "header block" is.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
```diff
@@ -0,0 +1,102 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+
+/// A class that implements the `IStreamable` protocol for `HttpBody`.
+/// It acts as a bridge between AWS SDK and CRT.
+class HttpContent: IStreamable {
```
epau, line 12: Something like `HttpBodyStreamable` is a bit more clear on the intention and removes the synonymous naming we have currently, where Body and Content have basically the same meaning.

Alternatively, there's also doing something like `AnyStreamable<T>` where the initiailizer takes closures for each of the IStreamable functions. This makes it really simply to compose anything into a streamable. Likely overkill rn and so i prefer simply changing the name.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,128 @@
...
+/// Note: This class is thread-safe.
+/// Note: if data is not read from the stream, the buffer will grow indefinitely until the stream is closed.
+///       or reach the maximum size of a `Data` object.
+public class BufferedStream: Stream {
+
+    /// Returns the length of the stream, if known
+    public private(set) var length: Int?
```
epau, line 18: We kind of touched on this but I forget where we ended up. 
Does this need to be optional? Doesn't every stream start with a length of 0 and gets incremented as more data is written to the stream. Can a stream never have a length? What does an unknown length mean?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,128 @@
...
+    /// Returns true if the stream is empty, false otherwise
+    public var isEmpty: Bool {
+        return buffer.isEmpty
+    }
+
+    /// Returns true if the stream is seekable, false otherwise
+    /// Note: This stream is seekable, but seeking to a position that is not in the buffer will cause the stream to throw an error.
+    public var isSeekable: Bool {
+        return false
+    }
```
epau, line 32: Does this value ever change or is it static?

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: Not every stream has length. This becomes more prominent when event streams comes in picture. We don't know how many events customer is suppose to send.

Based on its nullability, we decide the content length header

https://github.com/awslabs/smithy-swift/blob/jangirg/refactor/streams/Sources/ClientRuntime/Networking/Http/Middlewares/ContentLengthMiddleware.swift#L23

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,128 @@
...
+
+        return lock.withLockingClosure {
+            // if we're closed and there's no data left, return nil
+            // this will signal the end of the stream
+            if isClosed && data.isEmpty {
+                return nil
+            }
+
+            return data
+        }
```
epau, line 93: Now that we have the recursive lock, let's add the entire contents in the locking closure 

```suggestion
        return lock.withLockingClosure {
            var data = Data()

            while let next = try read(upToCount: Int.max) {
                data.append(next)
            }
            
            // if we're closed and there's no data left, return nil
            // this will signal the end of the stream
            if isClosed && data.isEmpty {
                return nil
            }

            return data
        }
```

Sources/ClientRuntime/Networking/Http/HttpContent.swift
ganeshnj: `StreamableHttpBody` ?

Streamable suffix sounds odd.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
epau: And should this value be `true` since we have defined an implementation for seeking?

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -7,63 +7,63 @@
...
-        case (let .buffer(lhsBuffer), let .buffer(rhsBuffer)):
-            return lhsBuffer.getData() == rhsBuffer.getData()
-        case (let .reader(lhsReader), let .reader(rhsReader)):
-            return lhsReader === rhsReader
+        case (.data(let lhsData), .data(let rhsData)):
+            return lhsData == rhsData
+        case (.stream(let lhsStream), .stream(let rhsStream)):
+            let lhsData = try? lhsStream.readToEnd()
+            let rhsData = try? rhsStream.readToEnd()
+            return lhsData == rhsData
```
epau, line 64: Here's another spot that would benefit from moving the equality logic to Stream

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
ganeshnj: https://github.com/awslabs/aws-crt-swift/blob/1279ecd11c952f75006c44144986a22fa0fb630e/Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift#L8

updated some logs to make it clearer.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: here is bit controversy, you can only seek in the buffer but this property tells that whole stream is seekable.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: that would block the read or write, depending what comes first.

Here the producer and consumer are 1:1. And we never expect multiple but they can be on different threads.


`readToEnd` `write` `readToEnd` `write` `write` `readToEnd`

something like this.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: answered in some other thread.

Sources/ClientRuntime/Networking/Streaming/Stream.swift
```diff
@@ -0,0 +1,63 @@
...
+public enum StreamError: Error {
+    case invalidOffset(String)
+    case notSupported(String)
+}
+
+extension Stream {
+    func isEqual(to other: Stream) throws -> Bool {
+        let selfData = try readToEnd()
+        let otherData = try other.readToEnd()
+        return selfData == otherData
```
epau, line 61: Rather than make this throwing, should we swallow the error and assume equality is false in the case where reading either of the streams fails? That simplifies things and seems to be a reasonable tradeoff in this scenario.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
epau: Hmm. I don't have full context but length feels like the wrong mechanism to dictate the headers. It would be much simpler to make length non-null, and use a value of 0 in the nil cases. This makes things much easier to reason about. 

Perhaps we should add a property to stream to dictate the encoding headers (`streamType` or `preferredHeader`), or the middleware maps the stream to the proper header type (if event stream -> use x header else use y header). That would also make things much easier reason about as it isn't clear, from reading the code alone, why length would dictate the header value.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
epau: Its also a bit odd to have this value return false yet we define an implementation for the public seek method. 
Something isn't quite right here.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
epau: I don't think I understand. My suggested change is needed to prevent race conditions, where the stream may be written to from Thread A, while Thread B is in the middle of reading to end. Without making my suggested change, there is no need for a recursive lock.

Sources/ClientRuntime/Networking/Streaming/Stream.swift
ganeshnj: that seems wrong to me, consumers should care about it then here. Having to bubble the error up forces us to decide better whether do we want to force/swallow the error.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: talked offline, updated it per our discussion.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: made some changes around this, no longer applicable.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
ganeshnj: per our offline chat, length is nullable, however we stopped setting it to 0 now.

Sources/ClientRuntime/Networking/Http/HttpContent.swift
ganeshnj: done

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -0,0 +1,114 @@
...
+    /// Returns the current position in the stream
+    public private(set) var position: Data.Index
+
+    /// Returns true if the stream is empty, false otherwise
+    public var isEmpty: Bool {
+        return buffer?.isEmpty == true
+    }
+
+    /// Returns true if the stream is seekable, false otherwise
+    /// Note: This stream is seekable, but seeking to a position that is not in the buffer will cause the stream to throw an error.
```
epau, line 29: Nit: I think we can remove this now.

```suggestion
```

Sources/ClientRuntime/Networking/Streaming/FileStream.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    let fileHandle: FileHandle
+
+    /// Returns the current position of the stream.
+    var position: Data.Index
+
+    /// Returns true if length is zero, false otherwise.
+    var isEmpty: Bool {
+        return length == 0
```
epau, line 29: Should this also return true if the length is nil?

Sources/ClientRuntime/Networking/Streaming/FileStream.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /// Returns true if the stream is seekable, false otherwise
+    var isSeekable: Bool = true
```
epau, line 33:
```suggestion
    let isSeekable: Bool = true
```

Sources/ClientRuntime/Networking/Streaming/CachingStream.swift
```diff
@@ -0,0 +1,148 @@
...
+        base.length
+    }
+
+    /// Returns true if the base stream is empty, false otherwise
+    public var isEmpty: Bool {
+        base.isEmpty
+    }
+
+    /// Returns true if the base stream is seekable, false otherwise
+    public var isSeekable: Bool {
```
epau, line 29:
```suggestion
    public let isSeekable: Bool  = true
```

Sources/ClientRuntime/Networking/Streaming/Stream.swift
epau: Hmm, i don't particularly like that we will crash in the case where this throws, since we are force trying in the places we use this. We shouldn't crash a customers app because reading a stream failed.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithHttpPayload.kt
```diff
@@ -25,62 +25,76 @@ class HttpResponseTraitWithHttpPayload(
...
-                ShapeType.BLOB -> {
-                    if (isBinaryStream) {
-                        writer.write("self.\$L = ByteStream.from(data: data)", memberName)
-                    } else {
-                        writer.write("self.\$L = data", memberName)
-                    }
+
+                // For binary streams, we need to set the member to the stream directly.
+                // this allows us to stream the data directly to the user
+                // without having to buffer it in memory.
```
epau, line 72: This comment mentions binary streams, yet a few liens above it appears we handle binary streams by setting body to `.data(data)`. Do we handle binary streams in multiple ways?

epau (approved):

PR #531, opened 2023-03-03
feat!: Remove Hashable conformance from generated errors & models

jbelkins:
## Issue \#
[#899](https://github.com/awslabs/aws-sdk-swift/issues/899)

## Description of changes
Removes `Hashable` from all errors and models since it is not used at all by the SDK.
- Structs and errors are no longer generated with the `Hashable` protocol applied (Smithy `enum`s and `intEnum`s remain `Hashable`)
- The `HashableShapeTransformer` is removed
- The custom `HashableTrait` is removed
- Code generation tests are adjusted for removal of `Hashable`

This is a breaking change because public struct and error types that were previously `Hashable` no longer are.

## Scope
- [x] New feature (breaking change which adds functionality)


epau (changes requested):
Can we keep the Hashable conformance for enums and int enums? 
IMO from the customer POV, itd be a bit unexpected that these enums aren't Hashable since we're the reason they don't automatically conform due to the associated enum case for handling backwards compatibility.

epau (approved):
Looks good, thanks for tackling!

PR #535, opened 2023-03-17
feat!: Add Smithy error name to generated errors

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/338

(Note that this PR merges to the retry feature branch)

## Description of changes

These error changes facilitate the classification of errors during retry.

- The `ServiceError` protocol is extended with a property that provides the name (minus namespace) for the underlying Smithy error type.
- Code-generated errors are generated with implementations for Smithy error type. 
- Unknown error is provided with a property for Smithy error type.
- Operation error enums are extended with a method that allows for the easy extraction of the underlying `ServiceError`.

This PR is a breaking change because it adds a field to a public protocol.

## Scope


Sources/ClientRuntime/Networking/ServiceError.swift
```diff
@@ -10,6 +10,10 @@ public protocol ServiceError {
     var _isThrottling: Bool { get set}
     var _type: ErrorType { get set }
     var _message: String? { get set }
+
+    /// The non-namespaced name of the Smithy shape for this error,
+    /// or `nil` if the Smithy type is not known.
+    var _smithyErrorTypeName: String? { get }
```
epau, line 16: This seems more like a type-level property, all instances of the same type should have the same value for this. Given that, can we make this static? Additionally, lets adjust the name so `smithy` isn't in it. 
Lastly, the only case that is the unknown case, therefore can we provide a reasonable default in that case instead of nil? I'd even argue that the unknown errors shouldn't conform to `ServiceError`.

```suggestion
    /// The non-namespaced name of the model for this error,
    static var _modelName: String { get }
```

Sources/ClientRuntime/Networking/UnknownServiceError.swift
```diff
@@ -10,10 +10,13 @@ public struct UnknownServiceError: ServiceError, Equatable {
     public var _isThrottling: Bool = false
     public var _type: ErrorType = .unknown
     public var _message: String?
+    public var _errorType: String?
```
epau, line 13: Can we add a comment describing what error type is here?

Sources/ClientRuntime/Networking/ServiceError.swift
jbelkins: I agree that the unknown errors shouldn't conform to `ServiceError` and I think that should be reformed in the error revamp.  But I don't want to change that here because it will increase the scope of these error changes beyond what is needed for retry, and IMO it is best saved for the error overhaul.

So I'll make this a static property of type `String?` for now.

Sources/ClientRuntime/Networking/UnknownServiceError.swift
jbelkins: Added

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -171,6 +171,21 @@ class ServiceGenerator(
...
+                    writer.write("switch self {")
+                    for (errorShape in errorShapes) {
+                        val errorShapeName = symbolProvider.toSymbol(errorShape).name
+                        writer.write("case .\$L(let error): return error", errorShapeName.decapitalize())
+                    }
+                    val unknownServiceErrorType = unknownServiceErrorSymbol.name
+                    writer.write("case .unknown(let error): return error")
+                    writer.write("}")
+                }
+            }
```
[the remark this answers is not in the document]
jbelkins: Test added.  There were no tests whatsoever for the operation output error, so I added that too.

epau (approved):

PR #537, opened 2023-03-21
feat: codegen and integration

ganeshnj:
## Scope


Sources/ClientRuntime/EventStream/Header.swift
```diff
@@ -34,8 +34,8 @@ extension EventStream {
         case bool(Bool)
         case byte(Int8)
         case int16(Int16)
-        case int32(Int32)
-        case int64(Int64)
```
ganeshnj, line 38: short - Int16
int - ..
long - ...

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -84,30 +84,56 @@ public class CRTClientEngine: HttpClientEngine {
...
+        case .unknown:
+            fatalError()
+        }
     }
 
-    public func makeHttpRequestStreamOptions(
-        _ request: SdkHttpRequest,
-        _ continuation: StreamContinuation
-    ) throws -> HTTPRequestOptions {
+    private func makeHttpRequestStreamOptions(
```
ganeshnj, line 124: add docs

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -84,30 +84,56 @@ public class CRTClientEngine: HttpClientEngine {
...
+        case .version_2:
+            self.logger.debug("Using HTTP/2 connection")
+            let crtRequest = try request.toHttp2Request()
+
+            return try await withCheckedThrowingContinuation { (continuation: StreamContinuation) in
+                let requestOptions = makeHttpRequestStreamOptions(request: crtRequest,
+                                                                  continuation: continuation,
+                                                                  http2ManualDataWrites: true)
+                Task {
+                    let stream = try connection.makeRequest(requestOptions: requestOptions) as! HTTP2Stream
```
ganeshnj, line 110: do is missing

Sources/ClientRuntime/Networking/Http/HttpVersion.swift
```diff
@@ -0,0 +1,6 @@
+//
```
ganeshnj, line 1: delete it

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
```diff
@@ -0,0 +1,151 @@
...
+
+        public func read(upToCount count: Int) throws -> ClientRuntime.Data? {
+            fatalError("read(upToCount:) is not supported by AsyncStream backed streams")
+        }
+
+        public func readToEnd() throws -> ClientRuntime.Data? {
+            fatalError("readToEnd() is not supported by AsyncStream backed streams")
+        }
+
+        public func readAsync(upToCount count: Int) async throws -> Data? {
```
ganeshnj, line 101: test case for it

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
```diff
@@ -0,0 +1,151 @@
...
+
+            return data
+        }
+
+        public func write(contentsOf data: ClientRuntime.Data) throws {
+            fatalError("write(contentsOf:) is not supported by AsyncStream backed streams")
+        }
+
+        public func close() throws {
+            // no-op
```
epau, line 148: Can you provide context as to why this is a no-op? iirc its because there is nothing for us to "close" in CRT.

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
```diff
@@ -0,0 +1,151 @@
...
+        public func makeAsyncIterator() -> AsyncIterator {
+            AsyncIterator(stream: stream,
+                          messageEncoder: messageEncoder,
+                          requestEncoder: requestEncoder,
+                          messageSinger: messageSinger)
+        }
+
+        // MARK: Stream
+
+        public var position: ClientRuntime.Data.Index = 0
```
epau, line 83: Can we add doc comments for all public properties and functions? I imagine we have corresponding docs on the protocol but we must bring also add comments here as well.

Sources/ClientRuntime/EventStream/Header.swift
epau: Just noting here, I think where we left off in our chat is that we want to keep the explicit types here as is to avoid any casting issues.

Sources/ClientRuntime/EventStream/Message.swift
```diff
@@ -22,3 +22,52 @@ extension EventStream {
...
+        }
+        return headers
+    }
+}
+
+extension Array where Element == EventStream.Header {
+    public func value(name: String) -> EventStream.HeaderValue? {
+        return self.first(where: { $0.name == name })?.value
+    }
+}
```
epau, line 73: Can we move these to the `Header.swift` file?
And please add doc comments for public functions here.

Sources/ClientRuntime/EventStream/Message.swift
```diff
@@ -22,3 +22,52 @@ extension EventStream {
...
+            case .uuid(let value):
+                headers.add(name: header.name, value: value.description)
+            }
+        }
+        return headers
+    }
+}
+
+extension Array where Element == EventStream.Header {
+    public func value(name: String) -> EventStream.HeaderValue? {
```
epau, line 70: Nit: IMO this function name is a bit too generic and depends too much on the array instance name to reason about the code. Can we add something header specific to the name?

```suggestion
    public func value(forHeader name: String) -> EventStream.HeaderValue? {
```

Sources/ClientRuntime/EventStream/MessageDecoderStream.swift
```diff
@@ -6,5 +6,26 @@
...
 
 /// Stream adapter that decodes input data into `EventStream.Message` objects.
-public protocol MessageDecoderStream: AsyncSequence {
+public protocol MessageDecoderStream: AsyncSequence where Event == Element {
+    associatedtype Event
+}
+
+extension MessageDecoderStream {
+    /// Returns an `AsyncThrowingStream` that decodes input data into `Event` objects.
+    public func toAsyncStream() -> AsyncThrowingStream<Event, Error> {
```
epau, line 15: Would we expect this function to ever be overridden by conforming types? 
If yes, can we also make this function part of the protocol conformance to avoid unexpected behavior. 

If `toAsyncStream` is not part of the protocol conformance and a conforming type overrides the function, the implementation that is called depends on the casted type of the instance.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -42,9 +42,9 @@ public class CRTClientEngine: HttpClientEngine {
             )
 
             var socketOptions = SocketOptions(socketType: .stream)
-    #if os(iOS) || os(watchOS)
+#if os(iOS) || os(watchOS)
```
epau, line 45: does `tvOS` matter here?

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
epau: Just a reminded to make sure to describe `http2ManualDataWrites`, what it does and why its needed

Sources/ClientRuntime/Networking/Http/CRT/HTTP2Stream+HttpBody.swift
```diff
@@ -0,0 +1,29 @@
...
+    /// Writes the HttpBody to the stream asynchronously
+    /// There is not recommended size for the data to write. The data will be written in chunks of 1024 bytes.
+    /// - Parameter body: The body to write
+    /// - Throws: Throws an error if the write fails
+    func write(body: HttpBody) async throws {
+        switch body {
+        case .data(let data):
+            try await writeData(data: data ?? .init(), endOfStream: true)
+        case .stream(let stream):
+            while let data = try await stream.readAsync(upToCount: 1024) {
```
epau, line 20: Can we move the bytes value to a property. This has the benefit of making it easier to understand what the value represents and also easier to change.

Sources/ClientRuntime/Networking/Http/HttpContext.swift
```diff
@@ -72,6 +84,9 @@ public class HttpContextBuilder {
     let hostPrefix = AttributeKey<String>(name: "HostPrefix")
     let logger = AttributeKey<LogAgent>(name: "Logger")
     let partitionID = AttributeKey<String>(name: "PartitionID")
+    public static let messageEncoder = AttributeKey<MessageEncoder>(name: "MessageEncoder")
+    public static let messageDecoder = AttributeKey<MessageDecoder>(name: "MessageDecoder")
+    public static let messageSigner = AttributeKey<MessageSigner>(name: "MessageSigner")
```
epau, line 89: Do these need to be public given the values are accessed through functions on HttpContext?

Additionally, it'd be great if you could also update the existing keys to static and reuse them in the Context as well. While not part of event streams, it should be low-effort and would make everything consistent.

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -18,7 +18,9 @@ public struct DeserializeMiddleware<Output: HttpResponseBinding,
             let response = try await next.handle(context: context, input: input) // call handler to get http response
             var copiedResponse = response
             if (200..<300).contains(response.httpResponse.statusCode.rawValue) {
-                let output = try Output(httpResponse: copiedResponse.httpResponse, decoder: decoder)
+                let output = try Output(httpResponse: copiedResponse.httpResponse,
+                                        decoder: decoder,
+                                        messageDecoder: context.getMessageDecoder())
```
epau, line 23: Only event streams require the message decoder/encoder, it'd be great if we didn't need to do this for all. I almost think that event-streams should have its own serde middlewares, especially since they have their own behavior (serialize headers -> sign request -> serialize chunk -> sign chunk -> serialize chunk -> sign chunk etc.....)

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -49,6 +49,17 @@ extension SdkHttpRequest {
         httpRequest.body = StreamableHttpBody(body: body)
         return httpRequest
     }
+
+    public func toHttp2Request() throws -> HTTPRequestBase {
```
epau, line 53: Why do we return a `HTTPRequestBase` here but the `toHttpRequest()` returns a `HTTPRequest`?

Can you add comments describing what makes this a http2 request? It looks almost identical to `toHttpRequest` except that body is nil. Why does setting `body` to nil make this a http2 request?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SymbolVisitor.kt
```diff
@@ -202,6 +207,13 @@ class SymbolVisitor(private val model: Model, swiftSettings: SwiftSettings) :
         if (nullableIndex.isMemberNullable(shape, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1)) {
             symbol = symbol.toBuilder().boxed().build()
         }
+
+        val isEventStream = targetShape.isStreaming && targetShape.isUnionShape
```
epau, line 211: why is union a requirement for event stream?

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpAWSJson11ProtocolGenerator.kt
```diff
@@ -106,6 +115,14 @@ class MockHttpAWSJson11ProtocolGenerator : HttpBindingProtocolGenerator() {
         // Intentionally empty
     }
 
+    override fun generateMessageMarshallable(ctx: ProtocolGenerator.GenerationContext) {
+        // Intentionally empty
+    }
+
+    override fun generateMessageUnmarshallable(ctx: ProtocolGenerator.GenerationContext) {
+        // Intentionally empty
+    }
```
epau, line 124: Why is it intentionally empty though?

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpEC2QueryProtocolGenerator.kt
```diff
@@ -108,6 +108,14 @@ class MockHttpEC2QueryProtocolGenerator : HttpBindingProtocolGenerator() {
         // Intentionally empty
     }
 
+    override fun generateMessageMarshallable(ctx: ProtocolGenerator.GenerationContext) {
+        // Intentionally empty
+    }
+
+    override fun generateMessageUnmarshallable(ctx: ProtocolGenerator.GenerationContext) {
+        // Intentionally empty
```
epau, line 116: Similar to an above question, why is it intentionally empty?

Sources/ClientRuntime/EventStream/Header.swift
ganeshnj: right we are going to use the explicit types than depending on the Swift runtime behavior. Because runtime behavior can lead to issues on the service side which has specified type for each date type.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
ganeshnj: updated.

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
ganeshnj: `close` is used when we have read all the data when receiving the response.

In this case, this stream is used for sending the data and we have no use case of that. Closing is marked by calling the finish on the continuation of the async stream which is base stream for it.

Sources/ClientRuntime/EventStream/Message.swift
ganeshnj: actually not needed anymore

Sources/ClientRuntime/EventStream/MessageDecoderStream.swift
ganeshnj: I don't expect it to be overridden because it is generic and works on the Element than any specific type.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
ganeshnj: I just reformat it. tvOS story is same as we have other places.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -84,30 +84,56 @@ public class CRTClientEngine: HttpClientEngine {
...
+                                                                  http2ManualDataWrites: true)
+                Task {
+                    let stream = try connection.makeRequest(requestOptions: requestOptions) as! HTTP2Stream
+                    try stream.activate()
+                    do {
+                        try await stream.write(body: request.body)
+                    } catch {
+                        self.logger.error(error.localizedDescription)
+                    }
+                }
```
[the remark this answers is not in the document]
ganeshnj: already fixed.

Sources/ClientRuntime/Networking/Http/CRT/HTTP2Stream+HttpBody.swift
ganeshnj: done

Sources/ClientRuntime/Networking/Http/HttpContext.swift
ganeshnj: they are access across packages. So yes, we need them public.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
ganeshnj: added comment for it

THIRD-PARTY-LICENSES
```diff
@@ -0,0 +1,678 @@
+** swift-collections; version 1.0.2 -- https://github.com/apple/swift-collections
```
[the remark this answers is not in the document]
ganeshnj: bad commit

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpEC2QueryProtocolGenerator.kt
ganeshnj: something to cover in sync

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpAWSJson11ProtocolGenerator.kt
```diff
@@ -47,7 +47,16 @@ class MockJsonHttpBindingResolver(
...
     }
 }
-class MockAWSJson11HttpProtocolCustomizations() : DefaultHttpProtocolCustomizations()
+class MockAWSJson11HttpProtocolCustomizations() : DefaultHttpProtocolCustomizations() {
+    override fun renderEventStreamAttributes(
+        ctx: ProtocolGenerator.GenerationContext,
+        writer: SwiftWriter,
+        op: OperationShape,
+    ) {
+        TODO("Not yet implemented")
```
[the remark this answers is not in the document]
ganeshnj: nope, used for test only for mocking purpose. We have tests in aws-sdk-codegen with implementation.

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
epau: Makes sense, can you add this as a comment here?

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpAWSJson11ProtocolGenerator.kt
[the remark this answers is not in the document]
epau: So instead of a TODO can we just add a `// no op` comment?

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -84,30 +84,72 @@ public class CRTClientEngine: HttpClientEngine {
...
-    public func makeHttpRequestStreamOptions(
-        _ request: SdkHttpRequest,
-        _ continuation: StreamContinuation
-    ) throws -> HTTPRequestOptions {
+    /// Creates a `HTTPRequestOptions` object that can be used to make a HTTP request
+    /// - Parameters:
+    ///   - request: The `HTTPRequestBase` object that contains the request information
+    ///   - continuation: The continuation that will be resumed when the request is complete
+    ///   - http2ManualDataWrites: Whether or not the request is using HTTP/2 manual data writes, defaults to `false`
+    ///     If not set to true, HTTP/2 manual data writes will be disabled and result in a runtime error.
```
epau, line 138:
```suggestion
    ///     If set to false, HTTP/2 manual data writes will be disabled and result in a runtime error.
```

Can you provide more information in what else is a factor here in causing the runtime error. If this flag defaults to false, reading the comment makes it sound like this function will always result in a runtime error then...

Sources/ClientRuntime/Networking/Http/CRT/HTTP2Stream+HttpBody.swift
```diff
@@ -0,0 +1,34 @@
...
+
+extension HTTP2Stream {
+    /// Returns the recommended size for the data to write
+    /// when using manual writes to HTTP2Stream
+    var manualWriteBufferSize: Int {
+        return 1024
+    }
+
+    /// Writes the HttpBody to the stream asynchronously
+    /// There is not recommended size for the data to write. The data will be written in chunks of 1024 bytes.
```
epau, line 18:
```suggestion
    /// There is not recommended size for the data to write. The data will be written in chunks of `manualWriteBufferSize` bytes.
```

Sources/ClientRuntime/Networking/Http/CRT/HTTP2Stream+HttpBody.swift
```diff
@@ -0,0 +1,34 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+
+extension HTTP2Stream {
+    /// Returns the recommended size for the data to write
```
epau, line 11:
```suggestion
    /// Returns the recommended size, in bytes, for the data to write
```

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -73,13 +72,13 @@ public class BufferedStream: Stream {
     /// It uses `read(upToCount:)` to read data in chunks.
     /// - Returns: Data read from the stream, or nil if the stream is closed and no data is available.
```
epau, line 73:
```suggestion
    /// It uses `read(upToCount:)` to read data in chunks.
    /// This function will block and wait until the stream is closed before returning the data read from the stream.
    /// - Returns: Data read from the stream, or nil if the stream is closed and no data is available.
```

Sources/ClientRuntime/EventStream/Message.swift
ganeshnj: actually name is right way to call the "key"

https://www.rfc-editor.org/rfc/rfc9110.html#fields

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpAWSJson11ProtocolGenerator.kt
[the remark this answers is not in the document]
ganeshnj: I would prefer TODO, as if this codepath is hit we get an exception and do the right thing.

Else we will loose the track where the issue is.

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
ganeshnj: fixed.

Sources/ClientRuntime/Networking/Http/CRT/HTTP2Stream+HttpBody.swift
```diff
@@ -8,14 +8,14 @@
...
-    /// Returns the recommended size for the data to write
+    /// Returns the recommended size, in bytes, for the data to write
     /// when using manual writes to HTTP2Stream
     var manualWriteBufferSize: Int {
         return 1024
     }
 
     /// Writes the HttpBody to the stream asynchronously
-    /// There is not recommended size for the data to write. The data will be written in chunks of 1024 bytes.
+    /// There is not recommended size for the data to write. The data will be written in chunks of `manualWriteBufferSize` bytes.
```
epau, line 18: typo
```suggestion
    /// There is no recommended size for the data to write. The data will be written in chunks of `manualWriteBufferSize` bytes.
```

epau (approved):
🔥

PR #539, opened 2023-03-22
fix: Reading from a `DataStreamReader` with `rewind: true` properly rewinds the stream allowing for subsequent reads.

epau:
This PR fixes an issue where subsequent reads from a DataStreamReader would return an unexpected value when the preceding read specified `rewind: true`. 

Passing in `rewind: true` when reading from a DataStreamReader should allow the next read to also read the same data. 

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/865

## Scope

### DIFF
--- a/Tests/ClientRuntimeTests/StreamingTests/DataStreamReaderTests.swift
+++ b/Tests/ClientRuntimeTests/StreamingTests/DataStreamReaderTests.swift
@@ -1,0 +1,37 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@testable import ClientRuntime
+import AwsCommonRuntimeKit
+import XCTest
+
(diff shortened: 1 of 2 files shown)


Tests/ClientRuntimeTests/StreamingTests/DataStreamReaderTests.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+@testable import ClientRuntime
+import AwsCommonRuntimeKit
+import XCTest
+
+class DataStreamReaderTests: XCTestCase {
+    let testData = Data([0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A])
+    
+    func testRead() {
```
ganeshnj, line 15: could you add one test on write -> read with rewind?

ganeshnj (approved):

jbelkins (approved):

PR #540, opened 2023-03-28
chore: Bumps crt version

epau:
This PR updates the version of the crt dependency to the current latest release `0.8.0`

## Scope

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #541, opened 2023-03-29
feat: Improves configuring imports for codegen

epau:
## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ImportDeclarations.kt
```diff
@@ -8,8 +8,22 @@ package software.amazon.smithy.swift.codegen
 class ImportDeclarations {
     private val imports = mutableSetOf<ImportStatement>()
 
-    fun addImport(packageName: String, isTestable: Boolean = false) {
-        imports.add(ImportStatement(packageName, isTestable))
+    fun addImport(packageName: String, isTestable: Boolean = false, internalSPIName: String? = null) {
+        val existingImport = imports.find { it.packageName == packageName }
```
jbelkins, line 12: Do we need to search for existing imports by package name _and_ internal SPI name?

i.e. you can `import ABC` and `import @_spi(DEF) ABC` in the same source file, right?

jbelkins:
also, if we are proceeding with this change, can we close #540?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ImportDeclarations.kt
epau: Nope. There'd be no reason to `import ABC` and `import @_spi(DEF) ABC` in the same source file.

`import @_spi(DEF) ABC` will import all of `ABC` plus it'll import anything that is marked with `@_spi(DEF)`.

jbelkins: What if we imported:
```
import @_spi(DEF) ABC
import @_spi(GHI) ABC
```
These are two distinct imports but this code would wipe out one of these two, no?

epau: Correct, this currently doesn't support adding multiple SPI internal names. I can add support for that if its something we want

PR #545, opened 2023-04-10
feat: Forces http2 for bidirectional event streams

epau:
This PR forces http2 for bidirectional event streams and updates the CRT http engine, to use a http2 specific connection manager when forcing http2. 

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/870

## Scope

### DIFF
--- a/Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
+++ b/Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
@@ -60,7 +60,11 @@ extension EventStream {
         }
 
         public func makeAsyncIterator() -> AsyncIterator {
-            AsyncIterator(stream: stream, messageDecoder: messageDecoder, responseDecoder: responseDecoder)
+            AsyncIterator(
+                stream: stream,
+                messageDecoder: messageDecoder,
+                responseDecoder: responseDecoder
+            )
         }
     }
 }
--- a/Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
+++ b/Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
@@ -6,39 +6,39 @@
 //
 
 extension EventStream {
-    
+
     /// Stream adapter that decodes input data into `EventStream.Message` objects.
     public struct DefaultMessageDecoderStream<Event: MessageUnmarshallable>: MessageDecoderStream {
         public typealias Element = Event
-        
+
         let stream: Stream
         let messageDecoder: MessageDecoder
         let responseDecoder: ResponseDecoder
-        
+
         public init(stream: Stream, messageDecoder: MessageDecoder, responseDecoder: ResponseDecoder) {
             self.stream = stream
             self.messageDecoder = messageDecoder
             self.responseDecoder = responseDecoder
         }
-        
+
         public struct AsyncIterator: AsyncIteratorProtocol {
             let stream: Stream
             let messageDecoder: MessageDecoder
             let responseDecoder: ResponseDecoder
-            
+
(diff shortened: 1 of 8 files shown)


Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
```diff
@@ -6,61 +6,65 @@
 //
 
 extension EventStream {
-
+
```
epau, line 9: Formatting changes are a result of auto formatting the entire file in xcode via (cmd + shift + i)

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
```diff
@@ -13,39 +13,44 @@ extension EventStream {
...
+            private var streamIterator: AsyncThrowingStream<Event, Error>.Iterator
+            
+            init(
+                stream: AsyncThrowingStream<Event, Error>,
+                messageEncoder: MessageEncoder,
+                requestEncoder: RequestEncoder,
+                messageSinger: MessageSigner
+            ) {
                 self.stream = stream
+                self.streamIterator = stream.makeAsyncIterator()
```
epau, line 46: Previously, we were creating a new iterator everytime `next()` is called. Now we create an iterator once during initi.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -17,6 +17,7 @@ public class CRTClientEngine: HttpClientEngine {
         private let windowSize: Int
         private let maxConnectionsPerEndpoint: Int
         private var connectionPools: [Endpoint: HTTPClientConnectionManager] = [:]
+        private var http2ConnectionPools: [Endpoint: HTTP2StreamManager] = [:]
```
jbelkins, line 20: Should this be a separate map of connection pools from `connectionPools`, or should connection type be a property of `Endpoint` so that all connections (HTTP 1.1 & 2.0) can be stored together?

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -83,6 +124,7 @@ public class CRTClientEngine: HttpClientEngine {
     public func execute(request: SdkHttpRequest) async throws -> HttpResponse {
         let connectionMgr = try await serialExecutor.getOrCreateConnectionPool(endpoint: request.endpoint)
         let connection = try await connectionMgr.acquireConnection()
+
         self.logger.debug("Connection was acquired to: \(String(describing: request.endpoint.url?.absoluteString))")
         switch connection.httpVersion {
```
jbelkins, line 129: The fact that the code for HTTP/1.1 & HTTP/2 is so close to one another makes me wonder if there's some common abstraction that both can adhere to, that would simplify this code

jbelkins (commented):
Overall, there's a lot of code duplication for introducing HTTP/2.0 that suggests an abstraction can be put in place around transmission protocol.  Wondering if you have thoughts on whether such an abstraction would be beneficial.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
epau: Yeah i agree, but things could get complicated trying to find a common abstraction. Classic example of `duplication is far cheaper than the wrong abstraction`. IMO it's worth looking into a common abstraction as part of SRA adherence tracked by https://github.com/awslabs/aws-sdk-swift/issues/698 and https://github.com/awslabs/aws-sdk-swift/issues/702

jbelkins (approved):

PR #553, opened 2023-05-05
fix: Make data streams completely async

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/974

## Description of changes
Changes the method of reading streams from intermittent polling to true continuation.  While stream is open, CPU sits at idle unless the stream is being read from or written.

- `BufferedStream` is made totally asynchronous so it does not need to be polled.  Reads that cannot be immediately fulfilled are stored in the form of a continuation, and fulfilled once data is available.
- The `CachingStream` type is eliminated because its purpose (allowing data to be kept in-memory for subsequent access after it is streamed) can be fulfilled without a dedicated stream type.
- An async method of reading to the end of a stream has been added, and used in code where possible.
- Equality between streams has been simplified to not require reading the whole stream and blocking a thread.
- Protocol tests now compare streaming data by reading the actual data instead of simply testing equality between streams (this is necessary because the definition of "equatable" for streams was changed to identity.)

## Scope


Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -107,4 +131,21 @@ public class BufferedStream: Stream {
...
     }
+
+    // If there is a client waiting to read data, and there is unread data remaining in the buffer,
+    // read the data and pass it to the reader via the continuation.
+    // Continue until there are no clients or no data.
+    private func _flushToNextClient() {
+        while !clients.isEmpty {
+            let (continuation, count) = clients[0]  // Don't remove the client from the array yet
+            do {
+                guard let data = try _read(upToCount: count), !data.isEmpty else { return }
```
epau, line 142: If `data` is `nil`, we should still resume the continuation as `nil` marks the end of the stream.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -107,4 +131,21 @@ public class BufferedStream: Stream {
             isClosed = true
         }
     }
+
+    // If there is a client waiting to read data, and there is unread data remaining in the buffer,
+    // read the data and pass it to the reader via the continuation.
+    // Continue until there are no clients or no data.
+    private func _flushToNextClient() {
+        while !clients.isEmpty {
```
epau, line 139: We should wrap this in the lock, to prevent any bugs from human error where someone may call this function without using a lock.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -32,6 +32,10 @@ public class BufferedStream: Stream {
     private(set) var buffer: Data?
     private let lock = NSRecursiveLock()
 
+    /// Continuations that are awaiting data, and the byte count to be delivered to each.
+    /// Access to this value should only occur while `lock` is locked.
+    private var clients: [(UnsafeContinuation<Data?, Error>, Int)] = []
```
epau, line 37: IMO `clients` is an overloaded term, perhaps use `callers` or `receivers` instead.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -47,24 +51,42 @@ public class BufferedStream: Stream {
...
-            let endPosition = position.advanced(by: toRead)
-            let chunk = buffer?[position..<endPosition]
+        try lock.withLockingThrowingClosure {
+            try _read(upToCount: count)
+        }
+    }
 
-            // remove the data we just read
-            buffer?.removeFirst(toRead)
+    private func _read(upToCount count: Int) throws -> Data? {
```
epau, line 59: Similar to another comment, we should use the lock in this function to prevent future bugs caused by human error.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
epau: Can we ever have multiple concurrent clients? Should this be a single item such as `(UnsafeContinuation<Data?, Error>, Int)?` instead of an array?

epau (changes requested):
If no tests are failing, we likely are missing a few tests cases particularly around verifying the stream is closed.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
jbelkins: `readers` is better, I'll use that.

Also, what would be the behavior if we don't accept multiple suspended readers?  Fail with an error if there is already a reader suspended?  Return nil or empty data to the existing suspended reader then replace it with the new one?  Maintaining a queue sidesteps this question.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
jbelkins: The intent is that `private func _read(upToCount:)` (as opposed to the `public` version, without the leading `_`) may be called when the lock is already held; this allows other methods that already hold the lock (such as `_serviceFirstReaderIfPossible()` to perform a read without deadlocking.

epau: `NSRecursiveLock` removes the risk of a deadlock in this case (which is why that type is being used for the lock instead of `NSLock`). It allows for the lock to be acquired multiple times by the same thread without causing a deadlock.

Sources/ClientRuntime/EventStream/DefaultMessageDecoderStream.swift
```diff
@@ -33,24 +33,22 @@ extension EventStream {
             }
 
             mutating public func next() async throws -> Event? {
```
jbelkins, line 35: Simplified this loop a bit by moving the `while` condition to the end of the loop

Sources/ClientRuntime/Networking/Http/HttpBody.swift
```diff
@@ -16,11 +16,7 @@ extension HttpBody: Equatable {
         case (.data(let lhsData), .data(let rhsData)):
             return lhsData == rhsData
         case (.stream(let lhsStream), .stream(let rhsStream)):
-            do {
-                return try lhsStream.isEqual(to: rhsStream)
-            } catch {
-                return false
-            }
+            return lhsStream === rhsStream
```
jbelkins, line 19: Stream equatability isn't used for any purpose other than to allow streams to be part of input & output shapes, which are `Equatable` by default.  The old equality implementation for streams synchronously read them to the end, which was unnecessary & added complexity.  So I have changed the stream equality check to a simple pointer comparison.

Sources/ClientRuntime/Networking/Http/HttpBody.swift
```diff
@@ -71,12 +87,15 @@ public extension HttpBody {
 }
 
 extension HttpBody: CustomDebugStringConvertible {
+    
     public var debugDescription: String {
```
jbelkins, line 91: Improved the debug description a little bit.

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -27,9 +27,8 @@ public struct DeserializeMiddleware<Output: HttpResponseBinding,
                 // error deserialization reads the stream multiple times to first deserialize the protocol error
                 // eg. [RestJSONError](https://github.com/awslabs/aws-sdk-swift/blob/d1d18eefb7457ed27d416b372573a1f815004eb1/Sources/Core/AWSClientRuntime/Protocols/RestJSON/RestJSONError.swift#L38,
                 // and then the service error eg. [AccountNotFoundException](https://github.com/awslabs/aws-sdk-swift/blob/d1d18eefb7457ed27d416b372573a1f815004eb1/Sources/Services/AWSCloudTrail/models/Models.swift#L62)
-                if case let .stream(stream) = copiedResponse.httpResponse.body, !stream.isSeekable {
-                    copiedResponse.httpResponse.body = .stream(CachingStream(base: stream))
-                }
+                let bodyData = try await copiedResponse.httpResponse.body.readAllData()
+                copiedResponse.httpResponse.body = .data(bodyData)
```
jbelkins, line 31: This code does the same as the previous code did, which is to read the body & save it in-memory so it can be used again.  But here it is done without needing the `CachingStream` type to do it, which was complex & not totally thread safe, and is eliminated in this PR.

Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift
```diff
@@ -5,66 +5,142 @@
...
-import Foundation
+import struct Foundation.Data
+import class Foundation.NSRecursiveLock
 
 /// A `Stream` implementation that buffers data in memory.
 /// The buffer size depends on the amount of data written and read.
 /// Note: This class is thread-safe.
 /// Note: if data is not read from the stream, the buffer will grow indefinitely until the stream is closed.
 ///       or reach the maximum size of a `Data` object.
 public class BufferedStream: Stream {
```
jbelkins, line 16: This class is substantially changed internally, but it functions the same from the perspective of the caller.

Instead of polling intermittently to see if there is data to be read, this class now stores a continuation queue for callers wanting to read from the stream & resumes the continuations when it is appropriate to resume them.

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -40,6 +40,28 @@ extension ByteStream {
     }
 }
 
+extension ByteStream {
```
jbelkins, line 43: Extension provides asynchronous read-to-end for `ByteStream`.

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -54,16 +76,13 @@ extension ByteStream: Codable {
...
         switch (lhs, rhs) {
         case (.data(let lhsData), .data(let rhsData)):
             return lhsData == rhsData
         case (.stream(let lhsStream), .stream(let rhsStream)):
-            do {
-                return try lhsStream.isEqual(to: rhsStream)
-            } catch {
-                return false
-            }
+            return lhsStream === rhsStream
```
jbelkins, line 85: Same use of identity for "equality" as in `HttpBody` up above

Sources/ClientRuntime/Networking/Streaming/FileStream.swift
```diff
@@ -70,10 +81,21 @@ class FileStream: Stream {
         }
     }
 
+    func readToEndAsync() async throws -> Data? {
```
jbelkins, line 84: Same as `readAsync(upToCount)` above, blocking file operation is wrapped in a continuation.

Sources/ClientRuntime/Networking/Streaming/Stream.swift
```diff
@@ -61,48 +65,10 @@ public enum StreamError: Error {
...
-
 extension Stream {
     public func seek(toOffset offset: Int) throws {
         guard isSeekable else {
             throw StreamError.notSupported("Seeking is not supported.")
         }
     }
 }
-
-extension Stream {
```
jbelkins, line 80: This was the not-really-asynchronous / polling stream read method that was causing CPU problems.

Instead of providing this method in a protocol extension, it's now implemented individually in the two stream types.

Sources/ClientRuntime/PrimitiveTypeExtensions/NSLock+Extension.swift
```diff
@@ -5,28 +5,18 @@
...
-     /// - Returns: The return value of the closure
-     func withLockingThrowingClosure<T>(closure: () throws -> T) throws -> T {
-         lock()
-         defer {
-             unlock()
-         }
-         return try closure()
-     }
- }
+    /// Executes a closure while holding the lock.
```
jbelkins, line 12: Replaced a throwing and a non-throwing implementation of the same operation with a single one that rethrows.

Tests/ClientRuntimeTests/NetworkingTests/CRTClientEngineIntegrationTests.swift
```diff
@@ -80,7 +80,8 @@ class CRTClientEngineIntegrationTests: NetworkingTestUtils {
         let response = try await httpClient.execute(request: request)
         XCTAssertNotNil(response)
         if case let HttpBody.stream(unwrappedStream) = response.body {
-            XCTAssert(try unwrappedStream.readToEnd()?.count == 1024)
```
jbelkins, line 83: Using the new `readToEndAsync()` method I introduced in this test class.

Tests/ClientRuntimeTests/NetworkingTests/NetworkingTestUtils.swift
```diff
@@ -27,7 +27,7 @@ class NetworkingTestUtils: XCTestCase {
      Create a mock HttpRequest with valid data payload
      */
     func setMockHttpDataRequest() {
-        var headers = Headers(["header-item-name": "header-item-value"])
+        let headers = Headers(["header-item-name": "header-item-value"])
```
jbelkins, line 30: Fixed some "change var to let" warnings

Tests/ClientRuntimeTests/NetworkingTests/Streaming/BufferedStreamTests.swift
```diff
@@ -6,65 +6,154 @@
 //
 
 import XCTest
-@testable import ClientRuntime
+import ClientRuntime
 
 final class BufferedStreamTests: XCTestCase {
```
jbelkins, line 11: Cleaned up this test class and added some new tests for the new async methods I added to this stream.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestResponseGenerator.kt
```diff
@@ -147,18 +150,27 @@ open class HttpProtocolUnitTestResponseGenerator protected constructor(builder:
...
             writer.write("XCTAssert(\$L === \$L)", expectedMemberName, actualMemberName)
         } else if ((shape.isDoubleShape || shape.isFloatShape)) {
             val stringNodes = test.params.stringMap.values.map { it.asStringNode().getOrNull() }
             if (stringNodes.isNotEmpty() && stringNodes.mapNotNull { it?.value }.contains("NaN")) {
-                val suffix = if (symbolProvider.toSymbol(member).isBoxed()) "?" else ""
                 writer.write("XCTAssertEqual(\$L$suffix.isNaN, \$L$suffix.isNaN)", expectedMemberName, actualMemberName)
             } else {
                 writer.write("XCTAssertEqual(\$L, \$L)", expectedMemberName, actualMemberName)
             }
+        } else if (shape.isBlobShape && shape.hasTrait<StreamingTrait>()) {
```
jbelkins, line 166: This renders an `XCAssert` for comparing two data streams.  It gets the data to local variables then compares them.

The final result, as rendered to a protocol test:
```
        // Compare blobs by reading them both to data
        let expectedBlobData = try await expected.blob?.readData()
        let actualBlobData = try await actual.blob?.readData()
        XCTAssertEqual(expectedBlobData, actualBlobData)
```

(btw `blob` is the Smithy equivalent of Swift `Data`.)

Sources/ClientRuntime/Networking/Http/HttpBody.swift
jbelkins: As it turns out, this did require a change to protocol tests to compare stream contents in testing (see below). But I'd rather have this code in tests than in production.

waahm7 (approved):

PR #557, opened 2023-05-11
fix: Don't change HTTP request components after they have been signed

jbelkins:
## Issue \#
#554

## Description of changes

This PR has companion integration tests in https://github.com/awslabs/aws-sdk-swift/pull/992.

Escaping is being added on requests' URL path after they are signed.  This causes requests to fail when certain characters requiring escaping are being used in the path.
- When converting `SdkHttpRequest` (a Smithy type) to `HttpRequest` (a CRT type), escaping is currently applied to the resulting path, even after the request has been codesigned.  Instead, escaping will now only applied when selected; and that will only be when creating `HttpRequest` for codesigning.  At other times after codesigning has been applied, no change will be made to the URL path.
- When updating the original request with the code signature after signing, the path was being split using `URLComponents` in a way that could add escaping; that has been fixed by adding the path into `URLComponents` using `percentEscapedPath` instead of `path`.
- When generating a URL on an `Endpoint`, set path on the URL using the `percentEscapedPath` component, so that further escaping is not applied.

## Scope


Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -129,7 +129,7 @@ public class CRTClientEngine: HttpClientEngine {
         switch connection.httpVersion {
         case .version_1_1:
             self.logger.debug("Using HTTP/1.1 connection")
-            let crtRequest = try request.toHttpRequest()
+            let crtRequest = try request.toHttpRequest(escaping: false)
```
epau, line 132: do we also need to do this for the http2 requests which we create in two places below?

jbelkins: Yes we do, good callout.  I'm going to change the `toHTTPRequest()` method to default to not escaping, though, since escaping is more the "special case" than not escaping.

epau (approved):

PR #558, opened 2023-05-16
fix: Fix tests & add CI coverage for unit tests on Linux

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1005

## Description of changes
Unit tests for `smithy-swift` are crashing on Linux + Swift 5.5 & 5.6 and two tests are not passing on any Linux + Swift.

To diagnose:
- A matrix of Linux + Swift combinations was created and CI is run on each, in addition to the existing CI which runs on Mac.  These will be marked as "Required" to pass once this PR merges.

To fix:
- Two tests used the `wait(for:timeout:)` method in asynchronous context.  This method is incompatible with async because it blocks the main thread and can cause deadlock.  The tests did not actually require waiting on expectations since the code under test executes in a deterministic order, so the use of expectations & `wait(for:timeout:)` was removed from each test.
- Two tests that do not pass on Linux were disabled on that platform, and will be fixed in the future.

## Scope

### DIFF
--- a/Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
+++ b/Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
@@ -24,7 +24,7 @@ class OperationStackTests: HttpRequestTestBase {
 
         var stack = OperationStack<MockInput, MockOutput, MockMiddlewareError>(id: "Test Operation")
         stack.initializeStep.intercept(position: .before, middleware: MockInitializeMiddleware(id: "TestInitializeMiddleware", callback: { _, _ in
-            currExpectCount = self.checkAndFulfill(currExpectCount, 1)
+            self.checkOrder(&currExpectCount, 1)
         }))
 
         stack.serializeStep.intercept(position: .before, middleware: MockSerializeMiddleware(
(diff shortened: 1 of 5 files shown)


Tests/SmithyTestUtilTests/RequestTestUtilTests/HttpRequestTestBaseTests.swift
```diff
@@ -155,7 +155,6 @@ class HttpRequestTestBaseTests: HttpRequestTestBase {
 
     // Mocks the code-generated unit test which includes testing for forbidden/required headers/queries
     func testSayHello() async throws {
-        let deserializeMiddleware = expectation(description: "deserializeMiddleware")
```
waahm7, line 158: Same as above. We can keep the expectation.

Tests/ClientRuntimeTests/SerializationTests/SerializationUtilsTests/TimestampSerdeUtilsTests.swift
```diff
@@ -31,6 +31,8 @@ class TimestampSerdeUtilsTests: XCTestCase {
 
     // MARK: - Encoding Tests
 
+    // TODO: Fix this test on Linux
```
waahm7, line 34: This seems like a weird todo. Maybe at least we should investigate why this is failing and add the details.

Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
```diff
@@ -13,13 +13,6 @@ class OperationStackTests: HttpRequestTestBase {
 
     func testMiddlewareInjectableInit() async throws {
         var currExpectCount = 1
-        let defaultTimeout = 2.0
-        let expectInitializeMiddleware = expectation(description: "initializeMiddleware")
```
waahm7, line 17: I faced a similar issue that expectation is not available on Linux. I solved it by using `XCTestExpectation` instead. We can use something similar if we need expectation in future.

Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
```diff
@@ -31,58 +24,49 @@ class OperationStackTests: HttpRequestTestBase {
 
         var stack = OperationStack<MockInput, MockOutput, MockMiddlewareError>(id: "Test Operation")
         stack.initializeStep.intercept(position: .before, middleware: MockInitializeMiddleware(id: "TestInitializeMiddleware", callback: { _, _ in
-            currExpectCount = self.checkAndFulfill(currExpectCount, 1, expectation: expectInitializeMiddleware)
+            currExpectCount = self.checkAndFulfill(currExpectCount, 1)
```
waahm7, line 27: Debatable: We are still using expectation terminology without expectations. Maybe we can just replace it with two lines something like 
rename `currExpectCount` to `currCallbacksCount` and doing 

```suggestion
	  XCTAssert(currCallbackCount, 1)
	  currCallbackCount += 1
```
Or
renaming the function to something  like

`checkExpectedCountAndIncrement`

waahm7 (approved):

Tests/ClientRuntimeTests/SerializationTests/SerializationUtilsTests/TimestampSerdeUtilsTests.swift
jbelkins: I don't want to spend more time on this right now, so I filed https://github.com/awslabs/aws-sdk-swift/issues/1006 so that these disabled tests are tracked & can be prioritized against other work.

Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
jbelkins: Expectations are available but the asynchronous-safe wait function (`await fulfillment(of:timeout`) is missing on Linux.  No idea why it's not there.

And in any case, these tests are async but not concurrent (i.e. no new `Task`s are created) and all code is guaranteed to execute before the test ends.  No expectations required to verify that all steps are completed in the expected order.

Tests/ClientRuntimeTests/ClientRuntimeTests/MiddlewareTests/OperationStackTests.swift
jbelkins: I cleaned this up a bit, plus I added a check at the end (verifying the counter value) to prove that the callbacks were actually made.

Tests/ClientRuntimeTests/SerializationTests/SerializationUtilsTests/TimestampSerdeUtilsTests.swift
jbelkins: I left a comment by each disabled test with a link to the ticket as well.

PR #567, opened 2023-06-23
fix: Properly escape HTTP paths for greedy label types

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1043

## Description of changes
Services besides S3 that were substituting text into URL paths via the `httpLabel` trait were ending up with double-encoded  paths, which caused operations to fail.  After investigating:
- S3 keys which were being substituted into URL paths were not being escaped at all, which was the default treatment for "greedy" URL paths.  Instead of not escaping, greedy labels are now escaped with the exception of the usual allowed characters minus forward-slash (greedy labels are inherently expected to contain forward slashes since they may span more than one URL path component).
- Since escaping no longer is performed as part of converting a `smithy-swift` HTTP request to a CRT HTTP request, the option to escape when that conversion is made is removed.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Endpoint.swift
+++ b/Sources/ClientRuntime/Networking/Endpoint.swift
@@ -11,7 +11,6 @@ public struct Endpoint: Hashable {
     public let protocolType: ProtocolType?
     public let host: String
     public let port: Int16
-
     public let headers: Headers?
     public let properties: [String: AnyHashable]
 
@@ -67,18 +66,13 @@ extension Endpoint {
         components.host = host
         components.percentEncodedPath = path
         components.percentEncodedQuery = queryItemString
-
         return components.url
     }
 
     var queryItemString: String? {
         guard let queryItems = queryItems else { return nil }
         return queryItems.map { queryItem in
-            if let value = queryItem.value {
-                return "\(queryItem.name)=\(value)"
-            } else {
-                return queryItem.name
-            }
+            return [queryItem.name, queryItem.value].compactMap { $0 }.joined(separator: "=")
         }.joined(separator: "&")
     }
 }
--- a/Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
+++ b/Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
@@ -29,11 +29,10 @@ public class SdkHttpRequest {
 
 extension SdkHttpRequest {
 
-    public func toHttpRequest(escaping: Bool = false) throws -> HTTPRequest {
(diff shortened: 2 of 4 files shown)


Tests/ClientRuntimeTests/NetworkingTests/HttpRequestTests.swift
```diff
@@ -90,16 +90,7 @@ class HttpRequestTests: NetworkingTestUtils {
         XCTAssert(updatedRequest.queryItems?.contains(ClientRuntime.URLQueryItem(name: "signedthing", value: "signed")) ?? false)
     }
 
-    func testPathInInHttpRequestIsEscapedPerRFC3986() throws {
```
jbelkins, line 93: Test no longer needed since HttpRequest no longer does any escaping

Sources/ClientRuntime/PrimitiveTypeExtensions/String+Extensions.swift
```diff
@@ -100,8 +100,8 @@ extension String {
 }
 
 extension String {
-    public func urlPercentEncoding() -> String {
-        self.urlPercentEncodedForQuery
+    public func urlPercentEncoding(encodeForwardSlash: Bool = true) -> String {
+        encodeForwardSlash ? urlPercentEncodedForQuery : urlPercentEncodedForPath
```
jbelkins, line 104: This function now optionally encodes forward slashes

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -29,11 +29,10 @@ public class SdkHttpRequest {
 
 extension SdkHttpRequest {
 
-    public func toHttpRequest(escaping: Bool = false) throws -> HTTPRequest {
```
jbelkins, line 32: Escaping of paths is no longer performed anywhere but at the point the path is created.  As a result, SdkHttpRequest no longer has an option to encode or not.

Sources/ClientRuntime/Networking/Endpoint.swift
```diff
@@ -67,18 +66,13 @@ extension Endpoint {
...
 
     var queryItemString: String? {
         guard let queryItems = queryItems else { return nil }
         return queryItems.map { queryItem in
-            if let value = queryItem.value {
-                return "\(queryItem.name)=\(value)"
-            } else {
-                return queryItem.name
-            }
+            return [queryItem.name, queryItem.value].compactMap { $0 }.joined(separator: "=")
```
jbelkins, line 75: lol we don't get paid by the line of code round here

PR #576, opened 2023-08-09
fix: Add default value handling for a scalar member of server response struct

sichanyoo:
This PR addresses the deserialization error where a default value is not correctly assigned to fields when they're missing from server response.

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/602

## Description of changes
Issue #602 is resolved by properly assigning the default value (`false` in this case) to [deleteMarker](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeletedObject.html#:~:text=Contents-,DeleteMarker,-Indicates%20whether%20the) property (a boolean member) of [DeletedObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeletedObject.html) when it is not found in S3 server response due to the responding bucket not being versioned.

With this change, when the deserializer tries to decode server response to [S3:deleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) from a non-versioned S3 bucket, in each [DeletedObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeletedObject.html), default value is correctly assigned to be `false` to delete markers.

Same applies for any other scalar member in any other server response structure that has a default value, if that member is missing from the server response struct.

A test case is added along with the change, as the change did not break any existing tests.

All changes were tested by building the `smithy-swift` project, as well as generating all the services using the changed code generator and running `swift test` as well as protocol tests on the `aws-sdk-swift` project.

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
@@ -29,6 +29,7 @@ import software.amazon.smithy.swift.codegen.integration.serde.TimestampDecodeGen
 import software.amazon.smithy.swift.codegen.integration.serde.TimestampHelpers
 import software.amazon.smithy.swift.codegen.integration.serde.xml.collection.CollectionMemberCodingKey
 import software.amazon.smithy.swift.codegen.integration.serde.xml.collection.MapKeyValue
+import software.amazon.smithy.swift.codegen.model.getTrait
 import software.amazon.smithy.swift.codegen.model.hasTrait
 import software.amazon.smithy.swift.codegen.model.isBoxed
 import software.amazon.smithy.swift.codegen.model.recursiveSymbol
@@ -300,16 +301,12 @@ abstract class MemberShapeDecodeXMLGenerator(
         val decodeVerb = if (memberTargetSymbol.isBoxed() && !isUnion || (member.hasTrait<DefaultTrait>())) "decodeIfPresent" else "decode"
         val decodedMemberName = "${memberNameUnquoted}Decoded"
 
-        var defaultValNilCoalescing = ""
-        if (member.hasTrait<DefaultTrait>()) {
-            val defaultTraitVal = member.getTrait(DefaultTrait::class.java).get().toNode()
-            if (defaultTraitVal.isStringNode) {
-                defaultValNilCoalescing = "?? \"${defaultTraitVal}\""
-            } else if (defaultTraitVal.toString().equals("null")) {
-                defaultValNilCoalescing = "?? nil"
-            } else {
-                defaultValNilCoalescing = "?? $defaultTraitVal"
-            }
+        val defaultTrait = if (member.hasTrait(DefaultTrait::class.java)) member.getTrait(DefaultTrait::class.java).get().toNode() else null
+        val defaultValNilCoalescing = when {
+            defaultTrait == null -> ""
+            defaultTrait.isStringNode() -> "?? \"${defaultTrait}\""
+            defaultTrait.toString().equals("null") -> "?? nil"
+            else -> "?? $defaultTrait"
         }
 
         if (unkeyed) {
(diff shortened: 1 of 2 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -296,12 +297,25 @@ abstract class MemberShapeDecodeXMLGenerator(
...
+        if (member.hasTrait<DefaultTrait>()) {
+            val defaultTraitVal = member.getTrait(DefaultTrait::class.java).get().toNode()
+            if (defaultTraitVal.isStringNode) {
+                defaultValNilCoalescing = "?? \"${defaultTraitVal}\""
+            } else if (defaultTraitVal.toString().equals("null")) {
+                defaultValNilCoalescing = "?? nil"
+            } else {
+                defaultValNilCoalescing = "?? $defaultTraitVal"
+            }
+        }
```
dayaffe, line 313: We should be able to take advantage of Kotlin syntax here to make the code more readable. For example, instead of defining a default variable and keeping track of several if statements we should be able to simplify it to something like:

```
val defaultValNilCoalescing = member.getTrait(DefaultTrait::class.java)?.let { trait ->
    val defaultTraitVal = trait.get().toNode()
    when {
        defaultTraitVal.isStringNode -> "?? \"${defaultTraitVal}\""
        defaultTraitVal.toString().equals("null") -> "?? nil"
        else -> "?? $defaultTraitVale"
    }
} ?: ""
```

_Note: I have not tested this against your branch so a line or two may need modification_

dayaffe (changes requested):
In general this approach looks good! Please take another look at lines 303-313 to see if we can make it more readable by taking advantage of Kotlin syntax. Also, I was wondering what the reasoning is for translating "null" to "nil"?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
sichanyoo: Yes, this change would make the code flow better! Working on the change rn, will push changes soon. Also, the reason for translating "null" to "nil" is because we're generating Swift code, and in Swift the keyword for "null" is "nil"!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
```diff
@@ -296,12 +299,22 @@ abstract class MemberShapeDecodeXMLGenerator(
...
         }
-        val decodeVerb = if (memberTargetSymbol.isBoxed() && !isUnion) "decodeIfPresent" else "decode"
+        val decodeVerb = if (memberTargetSymbol.isBoxed() && !isUnion || (member.hasTrait<DefaultTrait>())) "decodeIfPresent" else "decode"
         val decodedMemberName = "${memberNameUnquoted}Decoded"
+
+        val defaultValNilCoalescing = member.getTrait(DefaultTrait::class.java).getOrNull()?.let { trait ->
+            val defaultVal = trait.toNode()
+            when {
+                defaultVal.isStringNode() -> "?? \"$defaultVal\""
+                defaultVal.toString().equals("null") -> "?? nil"
```
jbelkins, line 309: Is there a better way to detect a null default value?  I am always wary of detecting a type by string like this... how does a null default actually get modeled by Smithy?

jbelkins (changes requested):
I like the tests here, since they cover several default value cases and not just the one the bug ticket flagged.

Just address the question about null defaults and @dayaffe's comments and this should be good to merge.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/MemberShapeDecodeXMLGenerator.kt
sichanyoo: That's a great point, I missed out on that. There exists a `isNullNode()` method for Smithy's Node types, so I've changed the code to use that method.

jbelkins: https://smithy.io/2.0/spec/type-refinement-traits.html#smithy-api-default-trait

Now I'm sort of wondering if `null` is a valid default value:
> The `@default` trait on root-level shapes cannot be set to `null`.

(Not exactly sure what they mean by "root level")

jbelkins: lol the answer is right above...

> The `@default` trait on a structure member can be set to `null` to explicitly indicate that the member has no default value or to override the default value requirement of a targeted shape.

so never mind, we do need the `null` case here.

dayaffe (approved):
LGTM

jbelkins (approved):

PR #578, opened 2023-08-10
fix: Add documentation generation for operations.

sichanyoo:
This PR adds in-line documentation generation for operations.

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1033

## Description of changes

Issue #1033 is resolved by adding documentation generation for errors thrown by operations.
When a service client protocol and service client are generated, for each and every operation, the code generator now generates in-line documentation for parameters, return type, and each and every error thrown by that operation (both errors common to that service, and errors bound to that operation), in the standard Swift in-line documentation format accepted by [Swift DocC](https://www.swift.org/documentation/docc/), like the following :

```
    /// This operation is cool.
    ///
    /// - Parameter TestStreamOpInput : [no documentation found]
    ///
    /// - Returns: `TestStreamOpOutputResponse` : [no documentation found]
    ///
    /// - Throws: One of the exceptions listed below __Possible Exceptions__.
    ///
    /// __Possible Exceptions:__
    /// - `SomeError` : You don't have permission.
```

Error name is extracted from name of the error structs from the corresponding operation shape.
Error cause is extracted from documentation on the error struct from the the corresponding service model file. If no documentation is provided for a given error struct in the original model file, "[no documentation found]" is appended as the error cause.

Below are example code-generated output, as well as what it looks like in Swift DocC:

Generated in-line documentation:

image

Quick-help view of in-line doc:

image

Documentation built with generated in-line doc:

image

This change was tested using tests in the `smithy-swift` project, as well as codegen tests in `aws-sdk-swift` project.

## Scope


jbelkins (changes requested):
A couple things:
- We should change the format to the following:
```
    /// - Throws: `BucketAlreadyExists`: The requested bucket name is not available.
    ///
    /// `BucketAlreadyOwnedByYou`: The bucket you tried to create already exists, and you own it.
```
Note the backticks around the exception name and the empty comment line between exceptions.  This renders as follows in Xcode inline docs:

Screenshot 2023-08-11 at 3 32 45 PM

- We should add `Parameters` and `Returns` to the inline function comment docs as well, for completeness.  We can write a separate ticket for that feature though.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -68,6 +73,42 @@ class ServiceGenerator(
...
         }
+
+        /**
+         * Helper method for generating in-line documentation for operation
+         */
+        private fun renderOperationDoc(model: Model, service: ServiceShape, op: OperationShape, writer: SwiftWriter) {
+            writer.writeShapeDocs(op)
+            writer.writeAvailableAttribute(model, op)
+
+            writer.writeSingleLineDocs { write("") }
```
dayaffe, line 84: Instead of writing `writer.writeSingleLineDocs { write("") }` we could take advantage of Kotlin's Local Functions. In general it is not good to overuse these but I think this is a case where it makes sense to improve readability.

Inside of this function we can declare a local function inside of renderOperationDoc 
```
fun writeEmptyLine() {
    writer.writeSingleLineDocs { write("") }
}
```

And now within the scope of this function we can write `writeEmptyLine()` instead of `writer.writeSingleLineDocs { write("") }` at all occurrences.

https://kotlinlang.org/docs/functions.html#local-functions

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -68,6 +73,42 @@ class ServiceGenerator(
...
+            if (op.getErrors(service).size > 0) {
+                writer.writeSingleLineDocs { write("") }
+                writer.writeSingleLineDocs { write("- Throws: `${op.toUpperCamelCase() + "Error"}` : Place-holder wrapper object for possible exceptions listed below.") }
+                writer.writeSingleLineDocs { write("") }
+                writer.writeSingleLineDocs { write("__Possible Exceptions:__") }
+            }
+
+            for (error in op.getErrors(service)) {
+                writer.writeDocs("\\- \\`${error.name}\\` : ${retrieveMemberShapeDoc(error.toShapeId(), model)}")
+            }
```
dayaffe, line 99: This is probably a matter of preference but I would prefer

```
val errorDocs = op.getErrors(service).joinToString("\n").forEach { error ->
    "\\- \\`${error.name}\\` : ${retrieveMemberShapeDoc(error.toShapeId(), model)}"
}
writer.writeDocs(errorDocs)
```

The above code utilizes Kotlin syntax as well as separates the concerns of string formation and doc writing

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -68,6 +73,42 @@ class ServiceGenerator(
...
+        /**
+         * Helper method to grab documentation for operation's member shapes (input, output, error(s)
+         */
+        private fun retrieveMemberShapeDoc(shapeId: ShapeId, model: Model): String {
+            val docTrait = model.getShape(shapeId).get().getTrait(DocumentationTrait::class.java).getOrNull()
+            return when {
+                docTrait == null -> "[no documentation found]"
+                else -> docTrait.value
+            }
+        }
```
dayaffe, line 111: Nice use of when here and .getOrNull() on a java class :)

dayaffe (changes requested):
Just a few more Kotlin syntax suggestions!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
sichanyoo: Given that it seems to be a matter of preference as you suggested, I think I'll keep the for loop format as I don't think changing this is worth the possible unexpected behavior / quirks of SwiftWriter I came across last week. I've changed all other syntax comments to reflect your comments tho :D

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -68,6 +73,46 @@ class ServiceGenerator(
...
+
+            writeEmptyLine()
+            writer.writeDocs("\\- Parameter ${op.inputShape.name} : ${retrieveMemberShapeDoc(op.inputShape, model)}")
+
+            writeEmptyLine()
+            writer.writeDocs("\\- Returns: \\`${op.outputShape.name}\\` : ${retrieveMemberShapeDoc(op.outputShape, model)}")
+
+            if (op.getErrors(service).isNotEmpty()) {
+                writeEmptyLine()
+                writer.writeSingleLineDocs { write("- Throws: `${op.toUpperCamelCase() + "Error"}` : Place-holder wrapper object for possible exceptions listed below.") }
```
jbelkins, line 96: What is this about a "place-holder wrapper"?  We should be throwing the individual exceptions unwrapped.

jbelkins (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
sichanyoo: Got on huddle call to determine desired error output, reflected in PR post.

jbelkins (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
```diff
@@ -68,6 +71,45 @@ class ServiceGenerator(
...
+
+        /**
+         * Helper method to grab documentation for operation's member shapes (input, output, error(s)
+         */
+        private fun retrieveMemberShapeDoc(shapeId: ShapeId, model: Model): String {
+            val docTrait = model.getShape(shapeId).get().getTrait(DocumentationTrait::class.java).getOrNull()
+            return when {
+                docTrait == null -> "[no documentation found]"
+                else -> docTrait.value
+            }
```
dayaffe, line 111: I thought about this line some more and since all we're doing here is a null check we can simplify this to 1 line!

`return docTrait?.value ?: "[no documentation found]"`

dayaffe (commented):
One more change to make use of Kotlin syntax and then I am good to approve!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
sichanyoo: Reflected suggestion in code. Resolving conversation.

dayaffe (approved):
LGTM

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ServiceGenerator.kt
jbelkins: Elvis operator FTW

PR #581, opened 2023-08-29
fix/Make operations correctly model service errors

sichanyoo:
## Issue \#

Addresses the [Issue #1080](https://github.com/awslabs/aws-sdk-swift/issues/1080)

## Description of changes

Adds pre-processing to the Smithy model before code-generation starts.
Puts errors defined at service level ("service errors") to all operations contained by the service.

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
@@ -25,6 +25,7 @@ import software.amazon.smithy.swift.codegen.integration.CustomDebugStringConvert
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
 import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
 import software.amazon.smithy.swift.codegen.model.AddOperationShapes
+import software.amazon.smithy.swift.codegen.model.FlattenServiceErrorsOntoOperations
 import software.amazon.smithy.swift.codegen.model.NestedShapeTransformer
 import software.amazon.smithy.swift.codegen.model.RecursiveShapeBoxer
 import software.amazon.smithy.swift.codegen.model.hasTrait
@@ -88,6 +89,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Void>() {
         resolvedModel = AddOperationShapes.execute(resolvedModel, settings.getService(resolvedModel), settings.moduleName)
         resolvedModel = RecursiveShapeBoxer.transform(resolvedModel)
         resolvedModel = NestedShapeTransformer.transform(resolvedModel, settings.getService(resolvedModel))
+        resolvedModel = FlattenServiceErrorsOntoOperations.transform(resolvedModel, settings.getService(resolvedModel))
         return resolvedModel
     }
 
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/FlattenServiceErrorsOntoOperations.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/FlattenServiceErrorsOntoOperations.kt
@@ -1,0 +1,29 @@
+package software.amazon.smithy.swift.codegen.model
+
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.knowledge.TopDownIndex
+import software.amazon.smithy.model.shapes.ServiceShape
+
+object FlattenServiceErrorsOntoOperations {
+    /**
+     * Puts error structures defined at service level to every operation under that service.
+     * Pre-processing is required to ensure operations correctly model service errors in addition to operation errors.
+     */
+    fun transform(model: Model, service: ServiceShape): Model {
+        // Get service errors. If none exist, return unmodified model
+        val serviceErrors = if (service.errors == null || service.errors.isEmpty()) return model else service.errors
+
+        val topDownIndex: TopDownIndex = TopDownIndex.of(model)
+        val operations = topDownIndex.getContainedOperations(service)
+        val modelBuilder: Model.Builder = model.toBuilder()
+
+        // Append service errors to each and every operation
+        for (op in operations) {
+            modelBuilder.addShape(
+                op.toBuilder().addErrors(serviceErrors).build()
+            )
+        }
+
+        return modelBuilder.build()
+    }
+}


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/FlattenServiceErrorsOntoOperations.kt
```diff
@@ -0,0 +1,29 @@
...
+        val modelBuilder: Model.Builder = model.toBuilder()
+
+        // Append service errors to each and every operation
+        for (op in operations) {
+            modelBuilder.addShape(
+                op.toBuilder().addErrors(serviceErrors).build()
+            )
+        }
+
+        return modelBuilder.build()
```
dayaffe, line 27: This code can be made into a bit more idiomatic kotlin. 
1. we have a return hidden in line 14, generally this is not good we want to try to keep possible exit scenarios explicit and easy to find. One way to do this is to replace `if x return y else z` with kotlin's takeUnless function
2. kotlin's forEach is preferred over a for loop

Maybe we can try re-writing this to following:
```
fun transform(model: Model, service: ServiceShape): Model {
    // Get service errors if any exist
    service.errors?.takeUnless { it.isEmpty() }?.let { serviceErrors ->
        val topDownIndex: TopDownIndex = TopDownIndex.of(model)
        val operations = topDownIndex.getContainedOperations(service)
        val modelBuilder: Model.Builder = model.toBuilder()

        // Append service errors to each and every operation
        operations.forEach { op ->
            modelBuilder.addShape(
                op.toBuilder().addErrors(serviceErrors).build()
            )
        }

        return modelBuilder.build()
    } ?: model // If service.errors is null or empty, return unmodified model
}
```

dayaffe (changes requested):
Logic looks good! Requesting some Kotlin code updates

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/FlattenServiceErrorsOntoOperations.kt
sichanyoo: Changed style to be more idiomatic to Kotlin as suggested.

dayaffe (approved):
Looks good!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
```diff
@@ -88,6 +89,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Void>() {
         resolvedModel = AddOperationShapes.execute(resolvedModel, settings.getService(resolvedModel), settings.moduleName)
         resolvedModel = RecursiveShapeBoxer.transform(resolvedModel)
         resolvedModel = NestedShapeTransformer.transform(resolvedModel, settings.getService(resolvedModel))
+        resolvedModel = FlattenServiceErrorsOntoOperations.transform(resolvedModel, settings.getService(resolvedModel))
```
jbelkins, line 92: While this approach would work, and even has a certain elegance about it, my concern is with how this will scale:
- For a number of errors `m` and a number of operations `n`, this will grow the generated codebase at a rate of `m * n`.  For a service with many operations (which we have in spades) and with many errors modeled at the service level (which we don't right now but I could easily see happening in the future) this will expand the generated codebase excessively.

An alternate approach:
- Code-generate a static method for each service that attempts to turn a response into one of the errors defined on the service.  The method should either return an error, or `nil` if no error modeled on the service can be created from the response.
- When parsing an error from each operation on the service, call that static method.  If an error is returned, throw it; else continue parsing the error the same way we currently do.

The advantage of this approach is that it would grow the codebase only proportionally to the number of service errors.

jbelkins (changes requested):
See my lone comment; though this would work, I think an alternate approach would be better suited.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
```diff
@@ -33,11 +33,17 @@ class HttpResponseGenerator(
...
+            httpResponseBindingErrorGenerator.renderOperationError(ctx, it, unknownServiceErrorSymbol)
         }
 
-        val modeledErrors = httpOperations.flatMap { it.errors }.map { ctx.model.expectShape(it) as StructureShape }.toSet()
+        val modeledOperationErrors = httpOperations.flatMap { it.errors }.map { ctx.model.expectShape(it) as StructureShape }.toSet()
+        var modeledServiceErrors: MutableSet<StructureShape> = mutableSetOf()
+        for (serviceError in ctx.service.errors) {
+            modeledServiceErrors.add(ctx.model.expectShape(serviceError) as StructureShape)
+        }
+        val modeledErrors = modeledOperationErrors + modeledServiceErrors
```
jbelkins, line 46: Same Q here as above, do we need to combine errors here?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/CodegenVisitor.kt
sichanyoo: Decided to go with alternate approach as suggested here after offline discussion.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
sichanyoo: Yes, operation errors and service errors need to be combined into a set for us to code generate initializers for all the errors that can be thrown by an operation in that service.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
```diff
@@ -320,7 +320,11 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
             .flatMap { it.errors }
             .map { ctx.model.expectShape(it) }
             .toSet()
-        return operationErrorShapes.filter { shapes -> shapes.members().any { it.isInHttpBody() } }.toMutableSet()
+        var serviceErrorShapes: MutableSet<Shape> = mutableSetOf()
+        for (serviceError in ctx.service.errors) {
+            serviceErrorShapes.add(ctx.model.expectShape(serviceError) as StructureShape)
+        }
+        return operationErrorShapes.filter { shapes -> shapes.members().any { it.isInHttpBody() } }.toMutableSet() + serviceErrorShapes
```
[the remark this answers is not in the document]
sichanyoo: Yes, adding service errors to returned operation shapes here ensures that the body struct and decodable extension are also correctly code generated for service error structures as well.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
```diff
@@ -320,7 +320,11 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
             .flatMap { it.errors }
             .map { ctx.model.expectShape(it) }
             .toSet()
-        return operationErrorShapes.filter { shapes -> shapes.members().any { it.isInHttpBody() } }.toMutableSet()
+        var serviceErrorShapes: MutableSet<Shape> = mutableSetOf()
+        for (serviceError in ctx.service.errors) {
+            serviceErrorShapes.add(ctx.model.expectShape(serviceError) as StructureShape)
+        }
```
dayaffe, line 326:
```
val serviceErrorShapes = ctx.service.errors.map { 
    ctx.model.expectShape(it) as StructureShape 
}.toMutableSet()
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
dayaffe, line 46: Spacing would help with readability here :) 
```
val modeledOperationErrors = httpOperations
    .flatMap { it.errors }
    .map { ctx.model.expectShape(it) as StructureShape }
    .toSet()
```

And for modeledServiceErrors we can achieve similar syntax to what you used in modeledOperationErrors with map
```
val modeledServiceErrors = ctx.service.errors
    .map { ctx.model.expectShape(it) as StructureShape }
    .toSet()
```

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpResponseBindingErrorGenerator.kt
```diff
@@ -34,4 +34,8 @@ class MockHttpResponseBindingErrorGenerator : HttpResponseBindingErrorGeneratabl
             }
         }
     }
+
+    override fun renderServiceError(ctx: ProtocolGenerator.GenerationContext) {
+        // Intentionally empty
```
dayaffe, line 39: Can you say more why this should be empty?

dayaffe (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
sichanyoo: Changed as suggested.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/HttpResponseGenerator.kt
sichanyoo: Changed as suggested.

smithy-swift-codegen/src/test/kotlin/mocks/MockHttpResponseBindingErrorGenerator.kt
sichanyoo: Discussed offline.

dayaffe (approved):
Looks great! Please fix failing CI. If further changes are required re-request review.

jbelkins (changes requested):
In the case where there are no modeled errors on the service, we should omit the method for `topLevelServiceErrorMembers` and the calls to it.

This is consistent with elsewhere in codegen, where components are omitted where they are an empty set or otherwise serve no function.

jbelkins (approved):

Issue #583, opened 2023-09-01
`SdkHttpRequest().headers` is  immutable


atierian:
### Describe the bug

The `headers` property of `SdkHttpRequest` is  immutable since `0.20.0`.
This unnecessarily prevents the callers from injecting headers using a custom `HttpClientEngine`.

### Expected Behavior

Ideally, expose an easier way to inject headers on a request level.
But in the absence of that, please make `headers` mutable.

### Current Behavior

`SdkHttpRequest().headers` is immutable.

### Reproduction Steps

```swift
class MyHTTPClientEngineProxy: HttpClientEngine {
    let base: HttpClientEngine

    init(base: HttpClientEngine) {
        self.base = base
    }

    func execute(request: SdkHttpRequest) async throws -> HttpResponse {
        // ❌ Cannot use mutating member on immutable value: 'headers' is a get-only property
        request.headers.add(name: "foo", value: "bar")
        return try await base.execute(request: request)
    }
}
```

### Possible Solution

Add a setter to `headers`.
- https://github.com/smithy-lang/smithy-swift/pull/584

### AWS SWIFT SDK version used

0.24.0

### Compiler and Version used

n/a

### Operating System and version

Darwin

sichanyoo:
Addressed in PR: https://github.com/smithy-lang/smithy-swift/pull/588

Where a convenience method `toBuilder()` is added to `SdkHttpRequest` to allow seamless update of a built `SdkHttpRequest` object.

sichanyoo:
Closed.

sichanyoo:
Re-opened; will be closed with the next minor version release of Swift SDK.

atierian:
Thanks for taking this on @sichanyoo.
If I'm not mistaken, this doesn't resolve our issue though. The new `toBuilder().with...().build()` API is creating a new instance of `SdkHttpRequest`, not mutating the existing one.

In other words, this
```swift
func execute(request: SdkHttpRequest) async throws -> HttpResponse {
    let newRequest = request.toBuilder().withHeader(name: "foo", value: "bar").build()
    return try await base.execute(request: newRequest)
}
```
is the same as this, which we can already do today
```swift
func execute(request: SdkHttpRequest) async throws -> HttpResponse {
    var headers = request.headers
    headers.add(name: "foo", value: "bar")
    let newRequest = SdkHttpRequest(
        method: request.method,
        endpoint: request.endpoint,
        headers: headers,
        queryItems: request.queryItems,
[...]

PR #588, opened 2023-09-18
fix: Add toBuilder() to SdkHttpRequest to easily allow customizations

sichanyoo:
## Issue \#

https://github.com/smithy-lang/smithy-swift/issues/583

## Description of changes

Adds a convenience method `toBuilder()` to `SdkHttpRequest` to allow customers to easily obtain a builder object that has all the fields from `SdkHttpRequest`. This will allow customers to update / customize any fields in the builder, then call `build()` on the builder to obtain a mutated `SdkHttpRequest` object they want.

## Scope


Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -25,6 +25,18 @@ public class SdkHttpRequest {
...
     }
+    
+    public func toBuilder() -> SdkHttpRequestBuilder {
+        return SdkHttpRequestBuilder()
+            .withBody(self.body)
+            .withMethod(self.method)
+            .withHeaders(self.headers)
+            .withPath(self.path)
+            .withHost(self.host)
+            .withQueryItems(self.queryItems ?? [])
```
jbelkins, line 36: Nil query items & empty query items are not the same thing.  We should preserve the `queryItems` value in the original request as-is.

jbelkins (changes requested):

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
sichanyoo: Changed to handle the nil `queryItems` value correctly.

jbelkins (approved):

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -25,6 +25,21 @@ public class SdkHttpRequest {
...
+            .withMethod(self.method)
+            .withHeaders(self.headers)
+            .withPath(self.path)
+            .withHost(self.host)
+            .withPort(self.endpoint.port)
+            .withProtocol(self.endpoint.protocolType ?? .https)
+        if let qItems = self.queryItems {
+            builder.withQueryItems(qItems)
+        }
+        return builder
```
dayaffe, line 41: Does passing an empty list cause problems? We could do instead:

```
return SdkHttpRequestBuilder()
            .withBody(self.body)
            .withMethod(self.method)
            .withHeaders(self.headers)
            .withPath(self.path)
            .withHost(self.host)
            .withPort(self.endpoint.port)
            .withProtocol(self.endpoint.protocolType ?? .https)
            .withQueryItems(qItems ?? [])
```

dayaffe (changes requested):
Please fix swiftlint and see if toBuilder can be simplified. If not then all good!

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
sichanyoo: That's actually what I did originally, but form Josh's comment having empty query items and having nil for query items is different - so I changed it to current state.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -25,6 +25,21 @@ public class SdkHttpRequest {
         self.endpoint = endpoint
         self.body = body
     }
+
```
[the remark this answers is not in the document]
sichanyoo: Fixed.

dayaffe (approved):
LGTM

PR #590, opened 2023-09-20
fix: Use original member name for serialization of URL-encoded forms

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1121

## Description of changes
When generating the serializer for `x-www-url-encoded` forms, the "lowerCamelCase" form of the member name is used when serializing timestamps.  This can cause errors at the server because the field name does not have the correct capitalization.
- This change uses the original member name to construct the form, so that URL-encoded forms containing timestamps may be read.
- Protocol tests for proper request serialization are in the corresponding `aws-sdk-swift` PR: https://github.com/awslabs/aws-sdk-swift/pull/1137

Also: Improve the messaging when a protocol test fails with non-matching query string parameters.

## Scope


sichanyoo (approved):
Lgtm

dayaffe (approved):
LGTM

PR #593, opened 2023-09-25
feat: add stored property for adding additional headers in SdkHttpRequest

dayaffe:
Provide the ability to add additional headers to an existing request of type SdkHttpRequest.

## Issue \#
[#1130](https://github.com/awslabs/aws-sdk-swift/issues/1130)
#583 

## Description of changes
Cause: `headers` is already a mutable property but it is computed without any stored values. This means that a customer who attempts to modify headers will either have no effect (because computed properties are read-only) or because it points to an instance of `endpoint.headers` whose headers are immutable. 

Solution: Add `additionalHeaders` stored property which is mutable and provide add/removal methods.

Risk: Computed properties ensure that the request is never modified in-transit. In order to reduce any risk associated I have chosen to add a new property that is mutable, stores values, and provided my own setters rather than modify the existing property to no longer be computed. Additionally, we should remove this functionality once Swift SDK SRA-compliant with interceptors ([#307](https://github.com/awslabs/aws-sdk-swift/issues/307))

How to use:
```
class MyHTTPClientEngineProxy: HttpClientEngine {
    let base: HttpClientEngine

    init(base: HttpClientEngine) {
        self.base = base
    }

    func execute(request: SdkHttpRequest) async throws -> HttpResponse {
        request.withHeader(name: "foo", value: "bar")
        return try await base.execute(request: request)
    }
}
```

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
+++ b/Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
@@ -46,11 +46,11 @@ public class SdkHttpRequest {
         return builder
     }
 
-    public func withHeader(name: String, value: String) -> Void {
+    public func withHeader(name: String, value: String) {
         self.additionalHeaders.add(name: name, value: value)
     }
 
-    public func withoutHeader(name: String) -> Void {
+    public func withoutHeader(name: String) {
         self.additionalHeaders.remove(name: name)
     }
 }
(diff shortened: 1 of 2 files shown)


sichanyoo (approved):
Lgtm

jbelkins (approved):
I approve this, as an interim fix to restore what Amplify needs, but this should be removed once reference architecture features that allow for headers to be set are implemented.

dayaffe:
This PR should be reverted once [#307](https://github.com/awslabs/aws-sdk-swift/issues/307) is implemented.

PR #595, opened 2023-09-27
feat: update swift versions to min 5.7 and max 5.9

dayaffe:
## Issue \#
[#1132](https://github.com/awslabs/aws-sdk-swift/issues/1132)
[#1133](https://github.com/awslabs/aws-sdk-swift/issues/1133)

## Description of changes
Update min version to swift 5.7 and max to 5.9

## Scope

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -50,10 +50,10 @@ jobs:
       fail-fast: false
       matrix:
         swift:
-          - 5.5-amazonlinux2
-          - 5.6-focal
           - 5.7-amazonlinux2
-          - 5.8-jammy
+          - 5.7-focal
+          - 5.9-amazonlinux2
+          - 5.9-jammy
     container:
       image: swift:${{ matrix.swift }}
     steps:
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.7
 
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.5
 
 import PackageDescription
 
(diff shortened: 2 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -50,10 +50,10 @@ jobs:
       fail-fast: false
       matrix:
         swift:
-          - 5.5-amazonlinux2
-          - 5.6-focal
           - 5.7-amazonlinux2
-          - 5.8-jammy
+          - 5.7-focal
+          - 5.9-amazonlinux2
+          - 5.9-jammy
```
sichanyoo, line 56: Not entirely sure if this matters, but according to the ticket https://github.com/awslabs/aws-sdk-swift/issues/1133, we should have `5.7-jammy` as well. I'd assume `5.7-focal` here should be replaced with `5.7-jammy`. @jbelkins could you confirm?

sichanyoo (commented):
Left a comment on swift container image

sichanyoo (approved):
Lgtm

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.7
```
jbelkins, line 1: I don't want to change this just yet (this will fail package resolution if at least the Swift version specified is not used). We will also need to update Catapult before we can do this; I will take on Catapult next week.

jbelkins (changes requested):

smithy-swift-codegen/src/test/kotlin/TestUtils.kt
```diff
@@ -344,7 +344,7 @@ fun getSettingsNode(
         .withMember("sdkId", Node.from(sdkId))
         .withMember("author", Node.from("Amazon Web Services"))
         .withMember("gitRepo", Node.from("https://github.com/aws-amplify/amplify-codegen.git"))
-        .withMember("swiftVersion", Node.from("5.5.0"))
+        .withMember("swiftVersion", Node.from("5.7.0"))
```
jbelkins, line 347: lol idk what this file does

dayaffe: Looks like its part of settings used in tests

dayaffe: I am going to revert this since it seems related to swift-tools-version

jbelkins (approved):

PR #596, opened 2023-09-29
fix: Correct serde for restJson1 event streams that contain errors

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1159

## Description of changes
Fixes a compile error that affects event streams on the `restJson1` protocol, which were discovered with the launch of the AWS Bedrock service.
- Corrects the serializer for an error received via an event stream to use the `properties` namespace.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -97,7 +97,7 @@ abstract class MemberShapeEncodeGenerator(
                 renderEncodeMap(ctx, memberName, topLevelContainerName, targetShape, level)
             }
             else -> {
-                renderSimpleShape(targetShape, memberName, containerName, null, false)
+                renderSimpleShape(targetShape, memberName, containerName, null, false, "")
```
dayaffe, line 100: rather than passing in `""` each time, what if we make path an optional parameter to `renderSimpleShape`?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
```diff
@@ -107,7 +107,8 @@ abstract class MemberShapeEncodeGenerator(
         memberName: String,
         containerName: String,
         codingKey: String?,
-        isBoxed: Boolean
+        isBoxed: Boolean,
+        path: String
```
dayaffe, line 111: As mentioned above, what if we say `path: String = ""` or `path: String? = null` and handle accordingly

dayaffe (changes requested):
Looks good! But I think we should try to avoid forcing many instances of passing in an empty string and instead handle the null case internally or default to ""

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
```diff
@@ -173,7 +173,8 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
                             generateCodingKeysForMembers(ctx, writer, httpBodyMembers)
                             writer.write("")
                         }
-                        renderStructEncode(ctx, shape, shapeMetadata, httpBodyMembers, writer, defaultTimestampFormat, "")
+                        val path = "properties.".takeIf { shape.hasTrait<ErrorTrait>() } ?: ""
```
sichanyoo, line 176: Related to David's comments - working with empty string literals and passing it into methods doesn't seem ideal. Maybe we could do something like:
`val path = "properties.".takeIf { shape.hasTrait<ErrorTrait>() } ?: null`
...and refactor `renderStructEncode` to take in an optional parameter with default value `""`.

sichanyoo (changes requested):
Left a comment regarding passing empty string literals as argument to functions.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/json/MemberShapeEncodeGenerator.kt
jbelkins: Path is now an optional String

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
jbelkins: Path is now an optional String

dayaffe (approved):
LGTM

sichanyoo (approved):
Lgtm

PR #597, opened 2023-10-03
feat: support initial-response in RPC based event streams

dayaffe:
## Issue \#
[#1163](https://github.com/awslabs/aws-sdk-swift/issues/1163)

## Description of changes
Render initial-response processing from AWSMessageDecoder (Companion PR: [#1165](https://github.com/awslabs/aws-sdk-swift/pull/1165)). All initial-response members should be included as part of Smithy model and are processed as individual members.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithoutHttpPayload.kt
```diff
@@ -133,4 +139,58 @@ class HttpResponseTraitWithoutHttpPayload(
...
+            write("} else {")
+            indent()
+
+            initialResponseMembers.forEach { responseMember ->
+                val responseMemberName = ctx.symbolProvider.toMemberName(responseMember.member)
+                write("self.$responseMemberName = nil")
+            }
+
+            dedent()
+            write("}")
```
sichanyoo, line 179: Make use of utility methods on SwiftWriter to keep codegen more readable + keep style uniform.
E.g., instead of manually indenting and dedenting, you can use `openBlock()` and `closeBlock()`.

```
writer.apply {
     openBlock("if let initialDataWithoutHttp = await messageDecoder.awaitIniitialResponse() {", "} else {") {
          // In if-block
          write("let decoder = JSONDecoder()")
          openBlock("do {", "} catch {") {
               // In do-block
               write("let response = try decoder.decode([String: String].self, from: initialDataWithoutHttp)")
               initialResponseMembers.forEach { responseMember ->
                    val responseMemberName = ctx.symbolProvider.toMemberName(responseMember.member)
                    write("self.$responseMemberName = response[\"$responseMemberName\"].map { 
                         value in KinesisClientTypes.Tag(value: value) 
                    }")
               }
          }
          // In catch-block
          write("print(\"Error decoding JSON: \\(error)\")")
          initialResponseMembers.forEach { responseMember ->
               val responseMemberName = ctx.symbolProvider.toMemberName(responseMember.member)
               write("self.$responseMemberName = nil")
          }
          // Close do-catch block
          closeBlock("}")
     }
     // In outer else-block
     initialResponseMembers.forEach { responseMember ->
                val responseMemberName = ctx.symbolProvider.toMemberName(responseMember.member)
                write("self.$responseMemberName = nil")
     }
     // Close if-else block
     closeBlock("}")
}
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithoutHttpPayload.kt
```diff
@@ -49,16 +49,19 @@ class HttpResponseTraitWithoutHttpPayload(
         val bodyMembersWithoutQueryTrait = bodyMembers
             .filter { !it.member.hasTrait(HttpQueryTrait::class.java) }
             .toMutableSet()
+        val initialResponseMembers = bodyMembers.filter {
+            val targetShape = it.member.targetOrSelf(ctx.model)
+            targetShape?.hasTrait(StreamingTrait::class.java) == false
+        }.toMutableSet()
```
sichanyoo, line 55: nit: Maybe use `toSet()` instead of `toMutableSet()` if it's not gonna be modified / shouldn't be modified except at this point in code?

sichanyoo (changes requested):
A nit and a comment on SwiftWriter utility methods

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithoutHttpPayload.kt
dayaffe: Tried this out and no matter the configuration openBlock / closeBlock seem to produce spacing issues. Reverting this change for now.

sichanyoo: Discussed offline and decided on adding a helper method to code-generator for this use case in a separate PR.

sichanyoo (approved):
Lgtm!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithoutHttpPayload.kt
```diff
@@ -49,16 +49,19 @@ class HttpResponseTraitWithoutHttpPayload(
         val bodyMembersWithoutQueryTrait = bodyMembers
             .filter { !it.member.hasTrait(HttpQueryTrait::class.java) }
             .toMutableSet()
+        val initialResponseMembers = bodyMembers.filter {
+            val targetShape = it.member.targetOrSelf(ctx.model)
+            targetShape?.hasTrait(StreamingTrait::class.java) == false
```
jbelkins, line 54: Just so I understand, if `targetShape` is `nil`, the whole left side of this expression evaluates to `nil`, which does not equal `false` so this entire expression is `false`.  Correct?

(that's how Swift would work, guessing Kotlin is the same when there is an optional boolean)

jbelkins (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/httpResponse/bindingTraits/HttpResponseTraitWithoutHttpPayload.kt
dayaffe: Correct! If `targetShape` is null the whole expression will evaluate to `false` because `null == false` is `false`

jbelkins: Just a nit, but `hasTrait<StreamingTrait>()` provides the same info with a little less boilerplate.  Or are they subtly different?

dayaffe: I believe they are the same but I was following the example used elsewhere in this file (see above on line 50 `it.member.hasTrait(HttpQueryTrait::class.java)`). Which is preferred? I could switch over the other instances in this file too if that method is preferred

PR #599, opened 2023-10-04
fix!: Remove suffix 'Response' from operation output type names

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1169

## Description of changes
In the past, operations could return either an `OutputResponse` or an `OutputError`.  SDK error handling was refactored to eliminate the `OutputError` enumeration, but the `OutputResponse` suffix remains on operation output structures.

To provide for more concise & intuitive naming, and to be consistent with both operation input structures and with the practice in other AWS SDKs, output structures will have the `Response` suffix removed from their name.

This is a breaking change because references to operation output structures will have to ba changed to refer to the new name.

## Scope


sichanyoo (approved):
Lgtm

PR #604, opened 2023-10-18
feat: add support for requiresLength trait and Transfer-Encoding: Chunked

dayaffe:
SEP: `/accepted/shared/requires-length-trait.md`
Smithy Docs: https://smithy.io/2.0/spec/streaming.html?highlight=requireslength#requireslength-trait

Add support for `@requiresLength` trait and properly set header `Transfer-Encoding: Chunked`

## Issue \#
[#1177](https://github.com/awslabs/aws-sdk-swift/issues/1177)

## Description of changes
- If requiresLength trait is set we should throw an error if we cannot calculate the length of the stream
- Change "Transfer-Encoded" header to "Transfer-Encoding"
- Transfer-Encoding header should be set if requiresLength trait is not present and the payload is marked with unsignedPayload trait
- Transfer-Encoding header should be removed in Http/2.0 requests since it is not supported. It is only supported in Http/1.1 requests.
- Tests for ContentLengthMiddleware.swift logic
- Tests for codegenerated client from smithy operations

## Scope


Tests/ClientRuntimeTests/NetworkingTests/ContentLengthMiddlewareTests.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    private func forceEmptyStream() {
+        // Force stream length to be nil
+        stack.finalizeStep.intercept(position: .before, id: "set nil stream length") { (context, input, next) -> OperationOutput<MockOutput> in
+            input.body = .stream(BufferedStream()) // Set the stream length to nil
+            return try await next.handle(context: context, input: input)
+        }
+    }
+        
+    private func AssertHeadersArePresent(expectedHeaders: [String: String]) async throws -> Void {
```
jbelkins, line 72: When creating a reusable test assertion, it is helpful to have it capture the file and line so that the test failure can be marked in the right place in the IDE window.

See this explanation: https://www.iosdev.recipes/xctest/custom-assertions-file-and-line/

dayaffe: Updated function to pass file & line

Sources/ClientRuntime/Networking/Http/Middlewares/ContentLengthMiddleware.swift
```diff
@@ -22,8 +29,12 @@ public struct ContentLengthMiddleware<OperationStackOutput: HttpResponseBinding>
         case .stream(let stream):
             if let length = stream.length {
                 input.headers.update(name: "Content-Length", value: String(length))
+            } else if !requiresLength && unsignedPayload {
+                // only for HTTP/1.1 requests, will be removed in all HTTP/2 requests
+                input.headers.update(name: "Transfer-Encoding", value: "Chunked")
             } else {
-                input.headers.update(name: "Transfer-Encoded", value: "Chunked")
+                let errorMessage = unsignedPayload ? "operation requires length" : "sigv4 requires length"
```
sichanyoo, line 36: I think the error messages should have more details to make debugging easier for customers, e.g., "Stream operation [OPERATION NAME] requires content length." Or "SigV4 signing requires content length for [OPERATION NAME]."

Tests/ClientRuntimeTests/NetworkingTests/ContentLengthMiddlewareTests.swift
```diff
@@ -0,0 +1,85 @@
...
+                                              next: MockHandler(handleCallback: { (_, input) in
+                                                  for (key, value) in expectedHeaders {
+                                                      XCTAssert(input.headers.value(for: key) == value, file: file, line: line)
+                                                  }
+                                                  let httpResponse = HttpResponse(body: HttpBody.none, statusCode: HttpStatusCode.ok)
+                                                  let mockOutput = try! MockOutput(httpResponse: httpResponse, decoder: nil)
+                                                  let output = OperationOutput<MockOutput>(httpResponse: httpResponse,
+                                                                                           output: mockOutput)
+                                                  return output
+                                              }))
```
sichanyoo, line 83: I think it would be good if `MockHandler` construction is factored outside of the `handleMiddleware` method call to make this more readable.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
```diff
@@ -91,9 +94,8 @@ fun formatHeaderOrQueryValue(
     memberShape: MemberShape,
     location: HttpBinding.Location,
     bindingIndex: HttpBindingIndex,
-    defaultTimestampFormat: TimestampFormatTrait.Format
+    defaultTimestampFormat: TimestampFormatTrait.Format,
```
sichanyoo, line 97: nit: Why are there `,` added after last argument of function calls? (Not just here, but a dozen other places as well)

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
```diff
@@ -403,6 +405,31 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
+                        targetShape.isBlobShape &&
+                        targetShape.hasTrait<RequiresLengthTrait>()
+                }
+            }
+        }
+        return false
+    }
+
+    // Checks for @unsignedPayload trait on an operation
+    private fun hasUnsignedBody(op: OperationShape): Boolean {
```
sichanyoo, line 429: nit: Maybe rename the function to `hasUnsignedBodyTrait` for uniformity with `hasRequiresLengthTrait` and more context with just one more word

sichanyoo: Or is this even needed? It does make it read a bit more smoother, but I'm unsure whether introducing a new method here is a good idea given the readability and amount of functionality provided by the function doesn't seem to outweigh the extra bandwidth in having to remember to use this function instead of just calling `hasTrait<[TraitName]>()`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/ContentLengthMiddleware.kt
```diff
@@ -20,17 +20,20 @@ class ContentLengthMiddleware(val model: Model, private val alwaysIntercept: Boo
...
-                operationStackName,
-                middlewareStep.stringValue(),
-                position.stringValue(),
-                ClientRuntimeTypes.Middleware.ContentLengthMiddleware
-            )
+            val middlewareArgs = if (requiresLength || unsignedPayload) {
+                "requiresLength: $requiresLength, unsignedPayload: $unsignedPayload"
+            } else {
+                ""
+            }
```
sichanyoo, line 31: I think this could be written in a more idiomatic way by doing:
```
val str = "requiresLength: $requiresLength, unsignedPayload: $unsignedPayload"
val middlewareArgs = str.takeIf {requiresLength || unsignedPayload} ?: ""
```
*Factored out `str` to separate val just to avoid line breaks in one long statement.

sichanyoo (changes requested):
Left some comments for suggesting some small changes.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
dayaffe: I installed ktlint locally which started doing that every time I did a lint on the file, will check that kotlin style guide suggests

Sources/ClientRuntime/Networking/Http/Middlewares/ContentLengthMiddleware.swift
dayaffe: I think in this case adding the operation name would be redundant since ContentLengthMiddleware() is code-generated into relevant operations. I took the error messages as specified in the SEP through I agree more verbose error messages are almost always better. Take a look at the codegen test below to see what I mean and let me know what we think. If we want the operation name it would have to be passed in as an additional argument to the middleware.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpBindingProtocolGenerator.kt
dayaffe: Changing to `hasUnsignedPayloadTrait`. I think I like it in a helper for readability + maintainability should the condition change. Keeping logic inside of where we append/instantiate middlewares seemed a bit hairy

Sources/ClientRuntime/Networking/Http/Middlewares/ContentLengthMiddleware.swift
sichanyoo: I don't think it'll be redundant because even though the middleware is code-generated into each operation, if multiple operations were called on services in an application written by SDK end-user, they would have to track down inside which operation this middleware failed due to insufficient error message.

The name of the operation is actually passed within the middleware context object `context` to the `handle` operation, and you can access it easily using `context.getOperation()` - no code-gen stage input is required.

dayaffe: Awesome I agree, thanks!

sichanyoo (approved):
LGTM, just gotta fix the swiftlint warning on line being too long

jbelkins (approved):

PR #605, opened 2023-10-18
feat: middleware changes

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1145

## Description of changes

- Adds `AuthSchemeMiddleware`
  - This middleware is responsible for retrieving auth options defined for the model for the operation at hand, and "loading" one of those auth options into an auth scheme if the auth scheme is supported by the SDK AND identity resolver for that auth scheme is configured on the service config. 
- Adds `SignerMiddleware`
  - This middleware is responsible for signing the request using signer retrieved from the resolved auth scheme, saved to the middleware context by `AuthSchemeMiddleware` earlier in the operation stack.

Accompanying PR: https://github.com/awslabs/aws-sdk-swift/pull/1180

## Scope


jbelkins (changes requested):
Both of these middlewares have logic in them, there should be unit tests for it.

(The middlewares themselves look fine though)

jbelkins (approved):
Approved with the understanding that unit tests for middlewares will be added later:
https://github.com/awslabs/aws-sdk-swift/issues/1148

dayaffe (approved):

PR #609, opened 2023-10-25
feat: codegen changes

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1146

Companion PR in `aws-sdk-swift`: https://github.com/awslabs/aws-sdk-swift/pull/1190

## Description of changes

- Add constructor with default empty values for `AuthOption`
- Add storage / retrieval methods to middleware context object for `AuthScheme`, `AuthSchemeResolver`, and add methods moved over from `aws-sdk-swift`
- Add codegen for service specific auth scheme resolver protocol, service specific default auth scheme resolver struct, and service specific auth scheme resolver parameters struct.
- Add codegen for usage of `AuthSchemeMiddleware` and `SignerMiddleware` (these don't generate the middlewares. they simply generate the _usage_ of the middlewares)
- Hook up middleware codegen into existing codegen flow in `HttpBindingProtocolGenerator`
- Misc. changes on method signatures, etc.
- Add codegen test for auth scheme resolver generation.
- Update existing codegen tests to reflect changes in middleware context construction & middleware usage.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,251 @@
...
+import software.amazon.smithy.model.traits.AuthTrait
+import software.amazon.smithy.model.traits.OptionalAuthTrait
+import software.amazon.smithy.model.traits.Trait
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.integration.ServiceTypes
+import software.amazon.smithy.swift.codegen.utils.clientName
+import software.amazon.smithy.swift.codegen.utils.toLowerCamelCase
+
+const val AUTH_SCHEME_RESOLVER = "AuthSchemeResolver"
+const val AUTH_RESOLVER_PARAMS = "AuthSchemeResolverParameters"
```
dayaffe, line 18: should be private unless its used outside this class or planned to be used outside. If it is planned to be used outside this class maybe we should have an AuthSchemeResolveConstants file

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private fun renderResolverProtocol(ctx: ProtocolGenerator.GenerationContext, writer: SwiftWriter) {
+        writer.apply {
+            openBlock(
+                "public protocol ${getSdkId(ctx)}$AUTH_SCHEME_RESOLVER: \$L {",
+                "}",
+                ServiceTypes.AuthSchemeResolver
+            ) {
+                write("// Intentionally empty.")
```
dayaffe, line 66: what's the intention of this? Want to make it clear for those reading this code why it is intentionally empty

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,251 @@
...
+    private fun renderOperationSwitchCase(sdkId: String, opShape: OperationShape, opName: String, schemes: Map<ShapeId, Trait>, writer: SwiftWriter) {
+        writer.apply {
+            write("case \"$opName\":")
+            indent()
+            schemes.forEach {
+                if (it.key == SigV4Trait.ID) {
+                    write("var sigV4Option = AuthOption(schemeID: \"${it.key}\")")
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.signingName, value: ${(it.value as SigV4Trait).name})")
+                    openBlock("guard let region = serviceParams.region else {", "}") {
+                        write("throw ClientError.authError(\"Missing region in auth scheme parameters for SigV4 auth scheme.\")")
```
dayaffe, line 175: Should consider making various error messages constants rather than in-line during the codegen string so that it is more readable + potentially reusable.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,251 @@
...
+            schemes.forEach {
+                if (it.key == SigV4Trait.ID) {
+                    write("var sigV4Option = AuthOption(schemeID: \"${it.key}\")")
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.signingName, value: \"${(it.value as SigV4Trait).name}\")")
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.signingRegion, value: serviceParams.region)")
+
+                    val signedBodyHeader = if (sdkId == "s3" || sdkId == "glacier") ".contentSha256" else ".none"
+                    // Set .unsignedBody to false
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.unsignedBody, value: false)")
+                    // Set .signedBodyHeader to .contentSha256 IFF service is S3 / Glacier, set to .none otherwise.
```
dayaffe, line 209: might be good to provide a link to where this logic is from?

dayaffe (changes requested):
Just a few nits/suggestions!

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: These are used in files added in customizations milestone down the line - moved these to `ClientRuntimeTypes`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: This is just a parent protocol that all auth scheme resolver implementations of a given service must conform to. Added the comment to clarify.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Changed as suggested.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: The logic for this comes from here: https://github.com/awslabs/aws-sdk-swift/blob/14f3c16d226a005a4c99a61acd33f10716396b52/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSigningParams.kt#L55

Basically I'm just re-locating existing logic into the new auth flow.

dayaffe (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,254 @@
...
+                    // Set .unsignedBody to true IFF operation has UnsignedPayloadTrait
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.unsignedBody, value: $unsignedBody)")
+                    // Set .signedBodyHeader to .contentSha256 IFF service is S3 / Glacier AND operation does not have UnsignedPayloadTrait.
+                    // Set to .none otherwise.
+                    write("sigV4Option.signingProperties.set(key: AttributeKeys.signedBodyHeader, value: $signedBodyHeader)")
+
+                    write("validAuthOptions.append(sigV4Option)")
+                } else {
+                    write("validAuthOptions.append(AuthOption(schemeID: \"${it.key}\"))")
+                }
```
dayaffe, line 191: It looks like there is a lot of code duplicated between `renderOperationSwitchCase` and `renderDefaultSwitchCase`, what do you think about writing a helper that is re-used in both?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,254 @@
...
+                        val opName = it.name.toLowerCamelCase()
+                        val sdkId = getSdkId(ctx)
+                        val validSchemesForOp = serviceIndex.getEffectiveAuthSchemes(
+                            ctx.service,
+                            it,
+                            ServiceIndex.AuthSchemeMode.NO_AUTH_AWARE
+                        )
+                        renderOperationSwitchCase(sdkId, opShape, opName, validSchemesForOp, writer)
+                    }
+                }
```
dayaffe, line 155:
```
val operations = ctx.service.operations
operations.filter { opShape ->
    opShape.hasTrait(AuthTrait::class.java) ||
    opShape.hasTrait(OptionalAuthTrait::class.java) ||
    opShape.hasTrait(UnsignedPayloadTrait::class.java)
}.forEach { opShape ->
    val opName = it.name.toLowerCamelCase()
    val sdkId = getSdkId(ctx)
    val validSchemesForOp = serviceIndex.getEffectiveAuthSchemes(
        ctx.service, it, ServiceIndex.AuthSchemeMode.NO_AUTH_AWARE
    )
    renderOperationSwitchCase(sdkId, opShape, opName, validSchemesForOp, writer)
}
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,254 @@
...
+        serviceIndex: ServiceIndex,
+        ctx: ProtocolGenerator.GenerationContext,
+        writer: SwiftWriter
+    ) {
+        val sdkId = getSdkId(ctx)
+        val defaultResolverName = "Default$sdkId${ClientRuntimeTypes.Core.AuthSchemeResolver.name}"
+        val serviceProtocolName = sdkId + ClientRuntimeTypes.Core.AuthSchemeResolver.name
+
+        writer.apply {
+            writer.openBlock(
```
dayaffe, line 81: since you did `writer.apply` you should be able to call `openBlock` directly

dayaffe (changes requested):
Once this PR is good to go I would recommend holding off on merging until PR that contains relevant tests is reviewed as well to ensure this implementation produces desired results in all cases (following Test-Driven Development)

Also more of a personal preference so up to you, but we could consider including more comments when it comes to code generation such as done [here](https://github.com/smithy-lang/smithy-swift/blob/a03eedc088a19bbab5fb72c0af83d4c6c1ff49fe/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt#L80)

Edit: Let's add codegen tests to validate the output and that's good enough for me with this PR

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Ok so a lot of the logic here is re-located to AWSClientRuntime (to reduce redundant codegen & to have ONE place that has all the signingProperties logic) in the next milestone (customizations milestone) and there this is refactored. I could refactor the method shown here, but that would be pointless since the logic should be changed in next set of PRs & refactor is done there.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Changed as pointed out.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
dayaffe: No need to refactor here if you have plans to clean this up in the next milestone, let's just make sure we do that

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Changed to reflect suggested syntax.

dayaffe (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -0,0 +1,256 @@
...
+
+        ctx.delegator.useFileWriter("./$rootNamespace/${ClientRuntimeTypes.Core.AuthSchemeResolver.name}.swift") {
+            renderResolverParams(serviceIndex, ctx, it)
+            it.write("")
+            renderResolverProtocol(ctx, it)
+            it.write("")
+            renderDefaultResolver(serviceIndex, ctx, it)
+            it.write("")
+            it.addImport(SwiftDependency.CLIENT_RUNTIME.target)
+            it.addImport("AWSClientRuntime")
```
jbelkins, line 30: `smithy-swift` cannot depend on `aws-sdk-swift` or its library `AWSClientRuntime`.

Even without explicitly requiring `aws-sdk-swift` in this project's `Package.swift`, adding an import of that module in codegen is creating a "backdoor" dependency (i.e. the compiler doesn't require that `aws-sdk-swift` be required as a dependency but it is still a dependency since generated code will not function without being linked against that module), and even worse this is a circular dependency, where `aws-sdk-swift` and `smithy-swift` depend on each other.

I'm not immediately sure what is the best way to break this requirement without greater research and/or discussion, but we cannot refer to `aws-sdk-swift` or any of its components from this project.

jbelkins (changes requested):

Sources/ClientRuntime/Middleware/Attribute.swift
```diff
@@ -39,4 +39,8 @@ public struct Attributes {
     public mutating func remove<T>(key: AttributeKey<T>) {
         attributes.removeValue(forKey: key.name)
     }
+
+    public func getSize() -> Int {
+        return attributes.count
+    }
```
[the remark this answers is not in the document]
sichanyoo: The use case is to make sure there's at least one identity resolver configured on the service client. Checking non-nil isn't enough since default value is empty `Attributes`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Only dependency here was for saving values to middleware context, and luckily in following milestone (customizations milestone) I refactored it to handle it in auth scheme structs rather than in auth scheme resolver. This way auth scheme resolver (which is generic, for both smithy and AWS SDK) won't be dependent on AWSClientRuntime.

Sources/ClientRuntime/Middleware/Attribute.swift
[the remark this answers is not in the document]
jbelkins: But size could also be greater than zero due to any other non-identity related Attribute being set, right?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
sichanyoo: Updated code here by deleting AWS-dependent bits, which will be moved to auth scheme in AWSClientRuntime. About re-locating entirety of SigV4 related things from sdk to smithy, I'd rather keep that out of scope for the project and have that be part of modularization project, since it will require a new round of analysis.

Sources/ClientRuntime/Middleware/Attribute.swift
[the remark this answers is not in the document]
jbelkins: Attributes is sort of intended to be a "grab bag" that can be used by unrelated parts of the codebase.  I wouldn't assume that the contents of Attributes will only be "x" at any time after its creation.

jbelkins (commented):
A question on our use of attributes count.

Sources/ClientRuntime/Middleware/Attribute.swift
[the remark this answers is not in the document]
sichanyoo: In this case, I'm using an `Attributes` object to store only identity resolvers, which is used by getters and setters `getIdentityResolvers` and `withIdentityResolver`. So it's guaranteed to only have identity resolvers.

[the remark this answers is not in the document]
jbelkins: What type is holding its own set of Attributes?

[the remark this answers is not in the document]
sichanyoo: So there's `HttpContext`, which has an `Attributes` object to keep track of all middleware context values. I'm storing identity resolvers in `HttpContext` by having the attribute key be of type `Attributes`, so nested `Attributes` basically.

jbelkins (approved):

jbelkins (approved):

PR #610, opened 2023-10-29
chore: Convert idempotency token middleware from closure to reusable type

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1202

## Description of changes
Changes the idempotency token middleware from an anonymous closure that is rendered inline at every point of use to a reusable Swift type that is placed under unit test.

The reusable middleware uses Swift key paths to specify the idempotency token member on the input structure; the input structure type and the key path are passed as initialization parameters to the reusable middleware.

We currently have ~710 operations that use idempotency tokens throughout the SDK, and this change reduces generated SDK size by about 5000 lines by not duplicating the idempotency token code at every point of use.

## Scope


sichanyoo (approved):
lgtm

PR #613, opened 2023-10-31
feat: add utility method for converting SdkHttpRequest to URLRequest.

sichanyoo:
## Issue \#

Related Github issue: https://github.com/awslabs/aws-sdk-swift/issues/837
Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1198

## Description of changes

Adds constructor to `URLRequest` for converting `SdkHttpRequest` to `URLRequest`, and a test case.
Adds preprocessor conditional import functionality to `SwiftWriter`.

## Scope


Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -83,6 +89,28 @@ extension SdkHttpRequest {
...
+    public func toURLRequest() async throws -> URLRequest {
+        // Set URL
+        guard let url = self.endpoint.url else {
+            throw ClientError.dataNotFound("Failed to construct URLRequest due to missing URL.")
+        }
+        var request = URLRequest(url: url)
+        // Set method type
+        request.httpMethod = self.method.rawValue
+        // Set body
+        guard let body = try await self.body.readData() else {
```
jbelkins, line 102: `URLRequest` does have a `httpBodyStream` property for streaming the request body.  But for now, reading the entire body into memory should be sufficient.  If customers want to use presigned requests with bodies to large for RAM, we can revisit this.

Tests/ClientRuntimeTests/NetworkingTests/HttpRequestTests.swift
```diff
@@ -51,6 +57,31 @@ class HttpRequestTests: NetworkingTestUtils {
...
+        let mockHttpRequest = SdkHttpRequest(method: .get, endpoint: endpoint, body: httpBody)
+        let urlRequest = try await mockHttpRequest.toURLRequest()
+
+        XCTAssertNotNil(urlRequest)
+        guard let headersFromRequest = urlRequest.allHTTPHeaderFields else {
+            XCTFail("Headers in SdkHttpRequest were not successfully converted to headers in URLRequest.")
+            // Compiler doesn't recognize XCTFail as return / exception thrown
+            return
+        }
+        XCTAssertNotNil(headersFromRequest)
```
jbelkins, line 74: Is this check necessary?  `headersFromRequest` is necessarily non-nil here because of the `guard` statement above

Tests/ClientRuntimeTests/NetworkingTests/HttpRequestTests.swift
```diff
@@ -51,6 +57,31 @@ class HttpRequestTests: NetworkingTestUtils {
...
+
+        // Check URLRequest fields
+        XCTAssertTrue(headersFromRequest.contains { $0.key == "Testname-1" && $0.value == "testvalue-1" })
+        XCTAssertTrue(headersFromRequest.contains { $0.key == "Testname-2" && $0.value == "testvalue-2" })
+        let expectedBody = try await httpBody.readData()
+        XCTAssertTrue(urlRequest.httpBody == expectedBody)
```
jbelkins, line 80: Prefer `XCTAssertEqual` over comparing inside `XCTAssertTrue`

jbelkins (changes requested):

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
sichanyoo: Gotcha

Tests/ClientRuntimeTests/NetworkingTests/HttpRequestTests.swift
sichanyoo: That's true, changed as suggested.

Tests/ClientRuntimeTests/NetworkingTests/HttpRequestTests.swift
sichanyoo: Changed as suggested.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -83,6 +89,28 @@ extension SdkHttpRequest {
...
+            throw ClientError.serializationFailed("Failed to construct URLRequest due to HTTP body conversion failure.")
+        }
+        request.httpBody = body
```
dayaffe, line 105:
```
// Set body, handling any serialization errors
do {
    request.httpBody = try await self.body.readData()
} catch {
    throw ClientError.serializationFailed("Failed to construct URLRequest due to HTTP body conversion failure.")
}
```

In this case do / catch is more idiomatic than guard. From my understanding guard is meant to be used when we have a clear & simple condition we need to in order to proceed such as "X things is true" or "Y exists". In this case we are executing a method that we know may throw an error so its more of a complex statement. Therefore it is easier to read using do/catch as we're saying "we're gonna execute this block of code **knowing** that it may fail" vs saying "we don't expect this statement to fail but we need to **ensure** it never does to continue on a happy path"

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -83,6 +89,28 @@ extension SdkHttpRequest {
...
+        // Set headers
+        for header in self.headers.headers {
+            for value in header.value {
+                request.addValue(value, forHTTPHeaderField: header.name)
+            }
+        }
```
dayaffe, line 111:
```
self.headers.headers.forEach { header in 
    header.value.forEach { value in
        request.addValue(value, forHTTPHeaderField: header.name)
    }
}
```

forEach is more idiomatic

dayaffe (changes requested):
Requesting some code syntax changes

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
sichanyoo: Changed as suggested.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
sichanyoo: Changed as suggested.

dayaffe (approved):

jbelkins (approved):

PR #614, opened 2023-10-31
fix: Endpoint url should be nil if host or scheme is missing

dayaffe:
URL extension was set to optional to account for nil properties, but nil was never returned. As a result there was a TODO in the test wondering when endpoint.url should be nil.

## Issue \#
[#710](https://github.com/awslabs/aws-sdk-swift/issues/710)

## Description of changes
- Update code logic to set url to nil if host or scheme is nil (invalid URL)
- Update test to check for nil URL
- Remove TODO

## Scope


jbelkins (approved):
I sort of wonder if the `url` property should throw if the value is `nil` but this at least addresses the TODO

sichanyoo (approved):

PR #615, opened 2023-11-01
fix: Pool HTTP connections based on scheme, host, and port

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1035

## Description of changes
When pooling HTTP connections, the SDK currently shares connections among requests that are unique to that request's `Endpoint` structure.  Because `Endpoint` contains a lot of fields that are unrelated to where the connection is made (i.e. headers, path, query string, etc.), pooling by `Endpoint` results in almost every request being in its own pool.  This results in existing connections almost never being reused.

Per conversation with CRT, connections should be pooled together if they use the same scheme, host, and port.  A custom type is introduced that is used to store & reuse existing connection pool managers based on those fields alone.

Increased connection reuse seems to result in substantial performance improvements: integration tests are running about 2x faster with this change in place.  Improvements for customers who repeatedly access the same AWS service may see even greater gains.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
@@ -12,11 +12,28 @@ import Darwin
 
 public class CRTClientEngine: HttpClientEngine {
     actor SerialExecutor {
+
+        struct ConnectionID: Equatable, Hashable {
+            let endpoint: Endpoint
+
+            static func ==(lhs: ConnectionID, rhs: ConnectionID) -> Bool {
+                lhs.endpoint.port == rhs.endpoint.port &&
+                lhs.endpoint.host == rhs.endpoint.host &&
+                lhs.endpoint.protocolType == rhs.endpoint.protocolType
+            }
+
+            func hash(into hasher: inout Hasher) {
+                hasher.combine(endpoint.port)
+                hasher.combine(endpoint.host)
+                hasher.combine(endpoint.protocolType)
+            }
+        }
+
         private var logger: LogAgent
 
         private let windowSize: Int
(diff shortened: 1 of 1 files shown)


Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -29,22 +48,22 @@ public class CRTClientEngine: HttpClientEngine {
...
-            guard let connectionPool = http2ConnectionPools[endpoint] else {
+            let poolID = ConnectionPoolID(endpoint: endpoint)
+            guard let connectionPool = http2ConnectionPools[poolID] else {
                 let newConnectionPool = try createHTTP2ConnectionPool(endpoint: endpoint)
-                http2ConnectionPools[endpoint] = newConnectionPool // save in dictionary
+                http2ConnectionPools[poolID] = newConnectionPool // save in dictionary
                 return newConnectionPool
             }
-
             return connectionPool
```
dayaffe, line 67: `getOrCreateConnectionPool` and `getOrCreateHTTP2ConnectionPool` have a lot of duplicate code. Could we create a helper function that can be re-used in both that contains the majority of the logic?

dayaffe (changes requested):

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
dayaffe: Discussed offline -- better to not refactor this part of the code and prioritize getting out the bug fix

dayaffe (approved):

sichanyoo (approved):
Lgtm

PR #616, opened 2023-11-04
add default log level to initialize method

CyprienRicque:
Make it possible to set specific log configuration to some items and then a default one for non-set ones.

## Description of changes
Adding default value as parameter to initialize method

## Scope

Edit: Just noticed the commit should have a special [format](https://github.com/smithy-lang/smithy-swift/blob/main/CONTRIBUTING.md) sorry.. 

### DIFF
--- a/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
+++ b/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
@@ -15,13 +15,13 @@ public class SDKLoggingSystem {
         factories[label] = logHandlerFactory
     }
 
-    public class func initialize() {
+    public class func initialize(defaultLogLevel: SDKLogLevel = .info) {
         LoggingSystem.bootstrap { label in
             if let factory = factories[label] {
                 return factory.construct(label: label)
             }
             var handler = StreamLogHandler.standardOutput(label: label)
-            handler.logLevel = .info
+            handler.logLevel = defaultLogLevel.toLoggerType()
             return handler
         }
     }


Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
```diff
@@ -15,13 +15,13 @@ public class SDKLoggingSystem {
...
-    public class func initialize() {
+    public class func initialize(defaultLogLevel: SDKLogLevel = .info) {
         LoggingSystem.bootstrap { label in
             if let factory = factories[label] {
                 return factory.construct(label: label)
             }
             var handler = StreamLogHandler.standardOutput(label: label)
-            handler.logLevel = .info
+            handler.logLevel = defaultLogLevel.toLoggerType()
             return handler
```
sichanyoo, line 25: What's the reason for this addition when there already exists an initializer that takes in a `SDKLogLevel` in lines 29 to 35?

CyprienRicque: Because the other initializer ignores the manually set values. 
And the documentation of the bootstrap method says that bootstrap must be call only once otherwise it will likely crash

sichanyoo (approved):

Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
sichanyoo: Gotcha - just ran CIs, will merge now.

PR #619, opened 2023-11-13
feat!: Replace the XML encoder with a custom Smithy implementation

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/964

(this is the XML _encoder_ part of the project.  This is free-standing and can merge by itself without the decoder.)

Note that only the RestXML Smithy protocol uses XML _encoding_, the other protocols either use JSON or form URL encoding for their requests.  So this new encoder affects only about six services that use RestXML as their Smithy protocol.)

The third-party `XMLCoder` library remains in the project because it is still used for decoding.  The PR for the new XML decoder will follow once this PR merges, and will be built with similar patterns.  At that time, the `XMLCoder` library will be removed once and for all.

## Description of changes
Eliminates the use of the third-party `XMLCoder` library for encoding XML, replacing it with a first-party XML encoding solution.  As a result, the codegen for XML encoding is greatly simplified, and the generated code for these types is much more compact.
- A new `SmithyReadWrite` public library is added to this project, with standard interfaces for serialization based on the use of modular closures.
- A new `SmithyXML` public library is added to this project, which contains the new XML encoder.
- A new `SmithyTimestamps` non-public library is added, which allows `ClientRuntime` and `SmithyXML` to share the timestamp-related code.
- On Linux systems, `libxml2` is packaged as a Swift module for use by `SmithyXML`.  (Apple platforms come with `libxml2` already modularized and do not need this module.)
- `Encodable` conformance for restXML model types is replaced by per-type static methods that perform the encoding.  This allows for hiding the encoding implementation from the model's public interface, and for static dispatch during encoding for greater performance.  Type-safety is still achieved through the use of generics where needed.
- Because `Foundation`'s best tool for handling XML (the `XMLDocument` class) is not available on Apple platforms other than Mac, `libxml2` is used directly to parse XML.  `libxml2` is available standard on all Apple platforms, and is part of the standard Swift install for Linux, so no special dependency management should be required.
- Code generation has been simplified greatly by moving much of what was previously code-generated encoding logic into `SmithyXML`.

Other changes:
- Universal encoding interfaces are introduced in the middleware that performs encoding of both regular requests and event streams.  The JSON and Form URL encoders that are currently based on Swift Codable are adapted to this interface as well.
- The `HTMLBodyMiddleware` that was previously code generated with complex alternate code paths has been split into several single purpose middlewares in ClientRuntime that can be used as needed.
- Replaces the `Box<T>` class that was previously used to hold self-referencing properties with the `@Indirect` Swift property wrapper, which holds the member in a reference like `Box<T>` but requires less custom codegen to implement.

## Scope


Package.swift
jbelkins, line 1: Adding the XML feature as modularized, to save us work down the line

Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
jbelkins, line 1: Just fixing a spelling error

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/BlobBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is a blob

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/BlobStreamBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is a streaming blob

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/BodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is a Smithy document (I.e. JSON)

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/EnumBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is an enum

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/EventStreamBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is an event stream

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/PayloadBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is a member with a HTTPPayload trait

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/StringBodyMiddleware.swift
jbelkins, line 1: This type replaces the old code-generated HttpBodyMiddleware when the body is a string

Sources/ClientRuntime/PrimitiveTypeExtensions/BoxType.swift
jbelkins, line 1: Box is no longer used. Was used to allow recursive members in structure, now we have the @Indirect property wrapper instead

Sources/ClientRuntime/PrimitiveTypeExtensions/Indirect.swift
jbelkins, line 1: This property wrapper allows recursive struct members by storing its value in a reference.  Used because it’s less obtrusive than an enclosing reference type like Box above

Sources/ClientRuntime/Serialization/Encoder/DynamicNodeEncoding.swift
jbelkins, line 1: Deleted because XMLCoder is no longer used for encoding to XML

Sources/ClientRuntime/Serialization/Encoder/RequestEncoder.swift
jbelkins, line 1: Superficial edits

Sources/ClientRuntime/Serialization/Encoder/XMLEncoder+Extensions.swift
jbelkins, line 1: A RequestEncoder is no longer used for XML.  The closure based encoding system is used instead.

Sources/ClientRuntime/Serialization/FormURL/Encoder/FormURLReadWrite.swift
jbelkins, line 1: This type bridges the existing FormURL encoder to the new closure based encoding system.

Sources/ClientRuntime/Serialization/SerializationUtils/JSONReadWrite.swift
jbelkins, line 1: This type bridges the Swift JSON encoder to the new closure based encoding system.

Sources/SmithyXML/Writer/Writer.swift
```diff
@@ -0,0 +1,223 @@
...
+        try write(value?.base64EncodedString())
+    }
+
+    public func detach() {
+        element.detach()
+    }
+
+    private func record(string: String?) {
+        guard let string, let key = nodeInfoPath.last else { detach(); return }
+        let xmlKey = nodeInfoPath.last
```
dayaffe, line 208: `xmlKey` looks equivalent to `key`, we can remove line 208 and re-use the `key` we defined in the guard statement

Sources/ClientRuntime/Serialization/SerializationUtils/TimestampFormatter.swift
jbelkins, line 1: Rather than make the SmithyTimestamps module public, a couple methods & types are re-exported publicly by ClientRuntime.

Sources/ClientRuntime/Serialization/SerializationUtils/XML/MapEntry.swift
jbelkins, line 1: Encodable conformance is removed since it is no longer used

Sources/ClientRuntime/Serialization/SerializationUtils/XML/MapKeyValue.swift
jbelkins, line 1: Encodable conformance is removed since it is no longer used

Sources/SmithyXML/Writer/Writer.swift
```diff
@@ -0,0 +1,223 @@
...
+        switch xmlKey?.kind ?? .element {
```
dayaffe, line 209: given the above, if we re-use key from the guard statement when we do not need to check for nullability or provide a default case and do `switch key.kind { ... }`

Sources/ClientRuntime/Serialization/SerializationUtils/XML/ValueWriters.swift
jbelkins, line 1: Adds SmithyXML writer support to ByteStream.  This is located in ClientRuntime because ByteStream is there as well

Sources/SmithyXML/Writer/Writer.swift
```diff
@@ -0,0 +1,223 @@
...
+        case .attribute:
+            guard let parent = element.parent as? XMLElement else { break }
+            let attribute = XMLNode(kind: .attribute)
+            attribute.name = key.name
+            attribute.stringValue = string
+            parent.addAttribute(attribute)
+            detach()
```
dayaffe, line 216: could consider making this a helper function `private func addAttribute(name: String, value: String)`

Sources/SmithyReadWrite/DocumentWritingClosure.swift
jbelkins, line 1: The type def for a document writer closure.  A document writer closure’s job is to turn a data type into serialized data.

Sources/SmithyReadWrite/WritingClosure.swift
jbelkins, line 1: The type def for a value writing closure.  Each type has or must provide one if it is to be serialized.  The value writing closure’s job is to allow a writer to serialize a type.

Sources/SmithyTestUtil/RequestTestUtil/HttpRequestTestBase.swift
jbelkins, line 1: Refactored to not rely on an encoder.

dayaffe (changes requested):

indirect:
hi, I am a person, and when you type `@indirect` you subscribe me to your tickets. you may not want to do that in the future, unless you are actually looking for feedback from me. 😅

Sources/SmithyXML/Writer/Writer.swift
jbelkins: Done, though I actually made this into a custom initializer for `XMLElement`.

Sources/SmithyXML/Writer/Writer.swift
jbelkins: Done.  Much simpler now.

Sources/SmithyXML/Writer/Writer.swift
jbelkins: Done.

Sources/SmithyXML/Writer/Writer+LibXML2.swift
```diff
@@ -0,0 +1,115 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#if os(Linux)
+    import LibXML2
+#else
+    import libxml2
+#endif
```
waahm7, line 12: This will probably break if we ever add support for Windows in future. Not sure if it matters.

Package.swift
```diff
@@ -21,22 +23,51 @@ let package = Package(
...
-        .testTarget(
-            name: "ClientRuntimeTests",
-            dependencies: ["ClientRuntime", "SmithyTestUtil"]
+        .target(name: "SmithyReadWrite"),
+        .target(
+            name: "SmithyXML",
+            dependencies: [
+                "SmithyReadWrite",
+                "SmithyTimestamps",
+                .target(name: "LibXML2", condition: .when(platforms: [.linux]))
```
waahm7, line 38: Is it possible to have the same name as the macOS counterpart so that we don't have to use #if else in the code? I don't know, but it's worth checking out.

jbelkins: Done.  For some reason I was thinking it was not permitted, but it worked fine.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/OperationInputBodyMiddleware.kt
```diff
@@ -25,43 +40,155 @@ class OperationInputBodyMiddleware(
...
+        var defaultBody = "\"{}\"".takeIf { ctx.service.hasTrait<AwsJson1_0Trait>() || ctx.service.hasTrait<AwsJson1_1Trait>() || ctx.service.hasTrait<RestJson1Trait>() } ?: "nil"
+        var payloadMember = inputShape.members().find { it.hasTrait<HttpPayloadTrait>() }
+        if (payloadMember != null) {
+            payloadShape = ctx.model.expectShape(payloadMember.target)
+            val memberName = ctx.symbolProvider.toMemberName(payloadMember)
+            keyPath = writer.format("\\.\$L", memberName)
+            payloadWritingClosure = writingClosureUtils.writingClosure(payloadMember)
+            documentWritingClosure = documentWritingClosureUtils.closure(payloadMember)
+            isPayloadMember = true
+        }
```
dayaffe, line 69: In kotlin its best to avoid explicit null checks and instead use a safe operator `?.` combined with `let`

```
payloadMember?.let { member ->
    payloadShape = ctx.model.expectShape(member.target)
    val memberName = ctx.symbolProvider.toMemberName(member)
    keyPath = writer.format("\\.\$L", memberName)
    payloadWritingClosure = writingClosureUtils.writingClosure(member)
    documentWritingClosure = documentWritingClosureUtils.closure(member)
    isPayloadMember = true
}
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/readwrite/WritingClosureUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+                    TimestampUtils.timestampFormat(memberTimestampFormatTrait, shape)
+                )
+            }
+
+            else -> {
+                return writer.format("\$N.writingClosure(_:to:)", ctx.symbolProvider.toSymbol(shape))
+            }
+        }
+    }
+}
```
dayaffe, line 84: Kotlin best practice is to lift return statements out such that you do `return when { ... }` and then the individual `return` in each block can be removed completely as the last statement in a code block is always returned.

you can also replace the first `if ... else if ... else` with a when statement to make it more idiomatic.

In all, I suggest refactoring this whole function to the following:

```
private fun writingClosure(shape: Shape, memberTimestampFormatTrait: TimestampFormatTrait? = null): String {
    when {
        ctx.service.hasTrait<AwsJson1_0Trait>() || 
        ctx.service.hasTrait<AwsJson1_1Trait>() || 
        ctx.service.hasTrait<RestJson1Trait>() -> return "JSONReadWrite.writingClosure()"

        ctx.service.hasTrait<AwsQueryTrait>() || 
        ctx.service.hasTrait<Ec2QueryTrait>() -> return "FormURLReadWrite.writingClosure()"
    }

    return when (shape) {
        is MapShape -> {
            val keyNodeInfo = nodeInfoUtils.nodeInfo(shape.key)
            val valueNodeInfo = nodeInfoUtils.nodeInfo(shape.value)
            val valueWriter = writingClosure(shape.value)
            val isFlattened = shape.hasTrait<XmlFlattenedTrait>()
            writer.format(
                "SmithyXML.mapWritingClosure(valueWritingClosure: \$L, keyNodeInfo: \$L, valueNodeInfo: \$L, isFlattened: \$L)",
                valueWriter, keyNodeInfo, valueNodeInfo, isFlattened
            )
        }

        is ListShape -> {
            val memberNodeInfo = nodeInfoUtils.nodeInfo(shape.member)
            val memberWriter = writingClosure(shape.member)
            val isFlattened = shape.hasTrait<XmlFlattenedTrait>()
            writer.format(
                "SmithyXML.listWritingClosure(memberWritingClosure: \$L, memberNodeInfo: \$L, isFlattened: \$L)",
                memberWriter, memberNodeInfo, isFlattened
            )
        }

        is TimestampShape -> writer.format(
            "SmithyXML.timestampWritingClosure(format: \$L)",
            TimestampUtils.timestampFormat(memberTimestampFormatTrait, shape)
        )

        else -> writer.format("\$N.writingClosure(_:to:)", ctx.symbolProvider.toSymbol(shape))
    }
}
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/NodeInfoUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+    private fun namespaceParam(xmlNamespaceTrait: XmlNamespaceTrait?): String {
+        if (xmlNamespaceTrait != null) {
+            return writer.format(
+                ", namespace: .init(prefix: \$S, uri: \$S)",
+                xmlNamespaceTrait.prefix,
+                xmlNamespaceTrait.uri
+            )
+        } else {
+            return ""
+        }
```
dayaffe, line 70: return should be lifted out of `if` statement but we can also make it even more idiomatic by using safe calls

```
return xmlNamespaceTrait?.let {
    writer.format(
        ", namespace: .init(prefix: \$S, uri: \$S)",
        it.prefix,
        it.uri
    )
} ?: ""
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/OperationInputBodyMiddleware.kt
```diff
@@ -25,43 +40,155 @@ class OperationInputBodyMiddleware(
     override val position = MiddlewarePosition.AFTER
 
     override fun render(
+        ctx: ProtocolGenerator.GenerationContext,
         writer: SwiftWriter,
         op: OperationShape,
         operationStackName: String,
     ) {
+        if (!alwaysSendBody && !MiddlewareShapeUtils.hasHttpBody(ctx.model, op)) { return }
```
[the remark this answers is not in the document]
jbelkins: Fixed

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/OperationInputBodyMiddleware.kt
jbelkins: Fixed

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/readwrite/WritingClosureUtils.kt
jbelkins: Fixed as suggested

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/NodeInfoUtils.kt
jbelkins: Fixed

dayaffe (approved):

PR #624, opened 2023-11-29
feat!: Use closures for processing HTTP response

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/964

## Description of changes
Provides closure types and concrete closures for the processing of HTTP responses into models or errors.  These closures are provided to `DeserializeMiddleware` at operation stack creation and used to process the response.

Because the closures handle all responsibility for turning a response into an output value or error, all references to binding protocols are now removed from the operation stack.

This improves / simplifies SDK design by decoupling the operation stack and middlewares unrelated to the response from the manner in which the response is processed.  It also lays the groundwork for incorporating new XML decoding, which will not be based on protocols.

https://github.com/awslabs/aws-sdk-swift/pull/1242 makes corresponding changes in the AWS SDK for Swift.

Also:
- `OperationStackInput` & `OperationStackOutput` generic type names have been changed to be consistent across all operation stack components.
- Remove the `builder.json` file which is now unused.

## Scope


smithy-swift-codegen/src/test/kotlin/HttpProtocolUnitTestRequestGeneratorTests.kt
```diff
@@ -784,7 +780,8 @@ class HttpProtocolUnitTestRequestGeneratorTests {
                 throw serviceError
             })
         }
- """
```
jbelkins, line 787: May the single space someone left in front of these three quotes die a horrible death.

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -14,12 +25,10 @@ public struct DeserializeMiddleware<Output: HttpResponseBinding,
             Self.MOutput == H.Output,
             Self.Context == H.Context {
 
-            let decoder = context.getDecoder()
             let response = try await next.handle(context: context, input: input) // call handler to get http response
             var copiedResponse = response
             if (200..<300).contains(response.httpResponse.statusCode.rawValue) {
-                let output = try await Output(httpResponse: copiedResponse.httpResponse,
-                                        decoder: decoder)
+                let output = try await httpResponseClosure(response.httpResponse)
```
jbelkins, line 31: Response closure is called in place of creating an `Output` directly.  (The closure initializes an `Output` value the exact same way though.)

jbelkins: This also prevents the middleware from needing to know that `Output` is `HttpResponseBinding` conformant, and allows us to take reference to that protocol out of the entire operation stack.

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -29,9 +38,7 @@ public struct DeserializeMiddleware<Output: HttpResponseBinding,
                 // and then the service error eg. [AccountNotFoundException](https://github.com/awslabs/aws-sdk-swift/blob/d1d18eefb7457ed27d416b372573a1f815004eb1/Sources/Services/AWSCloudTrail/models/Models.swift#L62)
                 let bodyData = try await copiedResponse.httpResponse.body.readData()
                 copiedResponse.httpResponse.body = .data(bodyData)
-                let error = try await OutputError.makeError(httpResponse: copiedResponse.httpResponse,
-                                            decoder: decoder)
-                throw error
+                throw try await httpResponseErrorClosure(copiedResponse.httpResponse)
```
jbelkins, line 41: Same thing for response error as for response output above - the closure does the work of calling the `makeError` method, and the middleware needs to know nothing about the error response type.  So we can clear `HttpResponseErrorBinding` out of every middleware where it is referenced.

Sources/ClientRuntime/Middleware/PresignerShimHandler.swift
```diff
@@ -7,20 +7,20 @@
...
-struct PresignerShim<OperationStackOutput: HttpResponseBinding,
-                     OperationStackError: HttpResponseErrorBinding>: Middleware {
+struct PresignerShim<OperationStackOutput>: Middleware {
     public let id: String = "PresignerShim"
 
     private let handler: PresignerShimHandler
+    private let output: OperationStackOutput
 
-    init(handler: @escaping PresignerShimHandler) {
+    init(handler: @escaping PresignerShimHandler, output: OperationStackOutput) {
```
jbelkins, line 16: Since the removal of the `HttpResponseBinding` protocol means that this shim can no longer create an `Output`, one is passed into this middleware at creation.

PR #625, opened 2023-11-30
feat: add custom trait PaginationTruncationMember

dayaffe:
## Issue \#
[#1021](https://github.com/awslabs/aws-sdk-swift/issues/1021)

## Description of changes
- Added a custom trait PaginationTruncationMember
- Updated PaginatorGenerator to check for PaginationTruncationMember trait and codegen accordingly
- Updated PaginatorSequence to check for isTruncated if it is passed into the paginated method
- added codegen test to make sure isTruncatedKey gets added when custom trait @paginationTruncationMember is used

## Scope


jbelkins:
Resolve conflicts & I'll approve

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PaginatorGenerator.kt
```diff
@@ -122,22 +126,41 @@ class PaginatorGenerator : SwiftIntegration {
...
-                    inputSymbol,
-                    outputSymbol,
-                    operationShape.toLowerCamelCase()
-                )
+                val isTruncatedFlag = outputShape
+                    .members()
+                    .firstOrNull { it.hasTrait(PaginationTruncationMember.ID) }
+                    ?.defaultName()
+
+                if (isTruncatedFlag != null) {
```
jbelkins, line 142: Rather than an if-else for the entire return statement, can we construct a smaller string for the `isTruncatedKey:` param that is different whether or not the `isTruncated` flag is set?

This would save a lot of duplication

jbelkins (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PaginatorGenerator.kt
dayaffe: oh yeah I tried that approach at first with StringBuilder and then reverted when it caused errors, Ill try again

jbelkins (approved):
Approved & can merge once this finishes green:
https://github.com/awslabs/aws-sdk-swift/actions/runs/7063450922/job/19232267355

PR #627, opened 2023-12-05
chore: refactor HttpBody and ByteStream to be a single class ByteStream

dayaffe:
## Scope


Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -7,45 +7,12 @@
...
-        return .data(data)
-    }
-
-    /// Returns ByteStream from a FileHandle object.
-    /// - Parameter fileHandle: FileHandle object to be converted to ByteStream.
-    /// - Returns: ByteStream representation of the FileHandle object.
-    public static func from(fileHandle: FileHandle) -> ByteStream {
-        return .stream(FileStream(fileHandle: fileHandle))
-    }
+    case none
```
jbelkins, line 13: Naming a case `.none` can lead to confusion because a Swift optional is actually an `enum` with a case also named `.none`:

```
enum Optional<T> {
    case some(T)
    case none
}
```

Consider another name such as `noStream` to prevent hard to find bugs down the line.

jbelkins (changes requested):
Just one comment

PR #637, opened 2024-01-05
fix!: Add no-op behavior for initialize methods of logging system.

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/903

## Description of changes

Calling `bootstrap` method of `swift-log`'s `LoggingSystem` more than once per program execution results in unexpected behavior ([swift-log spec](https://github.com/apple/swift-log/blob/e97a6fcb1ab07462881ac165fdbb37f067e205d5/Sources/Logging/Logging.swift#L672-L674)), most likely a crash.

To avoid this crash, no-op behavior is added to `SDKLoggingSystem` which wraps `LoggingSystem`. Whether the `initialize` methods have been already called or not is internally recorded to `isInitialized` variable. If yes, methods return with no-op done.

To make it thread-safe, `SDKLoggingSystem` is made an actor, and methods made async. This change is a breaking change. 

## Scope

- [] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
+++ b/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
@@ -8,6 +8,7 @@
 import Logging
 
 public class SDKLoggingSystem {
+    private static var isInitialized = false
     private static var factories: [String: SDKLogHandlerFactory] = [:]
 
     public class func add(logHandlerFactory: SDKLogHandlerFactory) {
@@ -16,6 +17,7 @@ public class SDKLoggingSystem {
     }
 
     public class func initialize(defaultLogLevel: SDKLogLevel = .info) {
+        if isInitialized { return } else { isInitialized = true }
         LoggingSystem.bootstrap { label in
             if let factory = factories[label] {
                 return factory.construct(label: label)
@@ -27,6 +29,7 @@ public class SDKLoggingSystem {
     }
 
     public class func initialize(logLevel: SDKLogLevel) {
+        if isInitialized { return } else { isInitialized = true }
         LoggingSystem.bootstrap { label in
             var handler = StreamLogHandler.standardOutput(label: label)
             handler.logLevel = logLevel.toLoggerType()
--- a/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
+++ b/Sources/ClientRuntime/Logging/SDKLoggingSystem.swift
@@ -7,16 +7,16 @@
 
 import Logging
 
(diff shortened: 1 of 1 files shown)


jbelkins (changes requested):
This approach is not thread-safe because this may still result in multiple calls to `bootstrap` if it is simultaneously called from different threads.

Two possible approaches to thread-safe initialization:
- Use `NSRecursiveLock` to enforce exclusive access to `LoggingSystem` and the `isInitialized` variable.  You can find a recursive lock example in our codebase here:
https://github.com/smithy-lang/smithy-swift/blob/f11e2e816b8b7e578872bde38b56578caa69fd90/Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift#L71
- Use a Swift `actor` to enforce exclusive access to initialization.  Look up Swift actors if you have not used them before.

The first approach is technically not async-safe since it blocks the calling thread, though the effect is minimal if the lock is released quickly.  The second approach would require marking the initialize methods `async` which, depending on where they're called from, may be a breaking API change.

Look at either of these methods & let me know if you have further questions about how they work.

dayaffe:
> This approach is not thread-safe because this may still result in multiple calls to `bootstrap` if it is simultaneously called from different threads.
> 
> Two possible approaches to thread-safe initialization:
> 
> * Use `NSRecursiveLock` to enforce exclusive access to `LoggingSystem` and the `isInitialized` variable.  You can find a recursive lock example in our codebase here:
>   https://github.com/smithy-lang/smithy-swift/blob/f11e2e816b8b7e578872bde38b56578caa69fd90/Sources/ClientRuntime/Networking/Streaming/BufferedStream.swift#L71
> * Use a Swift `actor` to enforce exclusive access to initialization.  Look up Swift actors if you have not used them before.
> 
> The first approach is technically not async-safe since it blocks the calling thread, though the effect is minimal if the lock is released quickly. The second approach would require marking the initialize methods `async` which, depending on where they're called from, may be a breaking API change.
> 
> Look at either of these methods & let me know if you have further questions about how they work.

Super interesting! I just looked up these approaches and using `actor` seems like a good approach to me since it is part of Swift's concurrency features which we already utilize elsewhere (async/await). 

Based on your latest commit seems like you agree @sichanyoo! From my limited understanding it looks like you want to declare the `add` and `initialize` with async to make it work with actor.

jbelkins (approved):
Approved, with a change of title to match Conventional Commits format.

dayaffe (approved):

PR #642, opened 2024-01-11
feat: add wrapper for checksums + unit tests

dayaffe:
Create a wrapper class to utilize hash/checksum functions. We will read desired checksum from an enum and want to go from string -> hash function -> computed value -> hexadecimal string easily based on whatever input algorithm is chosen.

## Issue \#
[#1287](https://github.com/awslabs/aws-sdk-swift/issues/1287)

## Description of changes
- Add a HashFunction class that wraps hash/checksum implementations in CRT swift
- Add ability to go from string -> hash function and use .computeHash() to get the returned value which is of type `HashResult` and be either a .data(Data) or a .integer(UInt32)
- Add ability to go from Data -> hexadecimal string using %02 or UInt32 -> hexadecimal string using %08
- Unit tests for class

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/HashFunction.swift
+++ b/Sources/ClientRuntime/Networking/HashFunction.swift
@@ -10,6 +10,11 @@ enum HashResult {
     case integer(UInt32)
 }
 
+enum HashError: Error {
+    case invalidInput
+    case hashingFailed(reason: String)
+}
+
 enum HashFunction {
     case crc32, crc32c, sha1, sha256, md5
 
@@ -46,21 +51,21 @@ enum HashFunction {
                 let hashed = try data.computeSHA1()
                 return .data(hashed)
             } catch {
-                throw ClientRuntime.ClientError.unknownError("Error computing SHA1: \(error)")
+                throw HashError.hashingFailed(reason: "Error computing SHA1: \(error)")
             }
         case .sha256:
             do {
                 let hashed = try data.computeSHA256()
                 return .data(hashed)
             } catch {
-                throw ClientRuntime.ClientError.unknownError("Error computing SHA256: \(error)")
+                throw HashError.hashingFailed(reason: "Error computing SHA256: \(error)")
             }
         case .md5:
             do {
                 let hashed = try data.computeMD5()
                 return .data(hashed)
             } catch {
-                throw ClientRuntime.ClientError.unknownError("Error computing MD5: \(error)")
+                throw HashError.hashingFailed(reason: "Error computing MD5: \(error)")
             }
         }
     }
--- a/Sources/ClientRuntime/Networking/HashFunction.swift
+++ b/Sources/ClientRuntime/Networking/HashFunction.swift
@@ -5,6 +5,11 @@
 
 import AwsCommonRuntimeKit
 
+enum HashResult {
+    case data(Data)
+    case integer(UInt32)
+}
+
 enum HashFunction {
     case crc32, crc32c, sha1, sha256, md5
 
@@ -28,54 +33,36 @@ enum HashFunction {
         }
     }
 
-    func computeHash(of data: Data) throws -> Data? {
+    func computeHash(of data: Data) throws -> HashResult? {
         switch self {
         case .crc32:
             // Should turn this result back into a UInt32 using .toUInt32()
-            return Data(fromUInt32: data.computeCRC32())
+            return .integer(data.computeCRC32())
         case .crc32c:
             // Should turn this result back into a UInt32 using .toUInt32()
-            return Data(fromUInt32: data.computeCRC32C())
+            return .integer(data.computeCRC32C())
         case .sha1:
             do {
                 let hashed = try data.computeSHA1()
-                return Data(hashed)
+                return .data(hashed)
             } catch {
                 throw ClientRuntime.ClientError.unknownError("Error computing SHA1: \(error)")
             }
         case .sha256:
             do {
                 let hashed = try data.computeSHA256()
-                return Data(hashed)
+                return .data(hashed)
             } catch {
                 throw ClientRuntime.ClientError.unknownError("Error computing SHA256: \(error)")
             }
         case .md5:
             do {
                 let hashed = try data.computeMD5()
-                return Data(hashed)
+                return .data(hashed)
             } catch {
                 throw ClientRuntime.ClientError.unknownError("Error computing MD5: \(error)")
             }
         }
-    }
-}
-
-extension Data {
-
-    // Create a Data type from a UInt32 value
-    init(fromUInt32 value: UInt32) {
-        var val = value // Create a mutable UInt32
-        self = Swift.withUnsafeBytes(of: &val) { Data($0) }
-    }
-
-    // Convert a Data type to UInt32
-    func toUInt32() -> UInt32? {
-        guard self.count == MemoryLayout<UInt32>.size else {
-            return nil // Ensures the data is of the correct size
-        }
-
-        return self.withUnsafeBytes { $0.load(as: UInt32.self) }
     }
 }
 
@@ -96,5 +83,18 @@ extension UInt32: HexStringable {
     // Convert a UInt32 type to a hexademical String
     func toHexString() -> String {
         return String(format: "%08x", self)
+    }
+}
+
+extension HashResult: HexStringable {
+
+    // Convert a HashResult to a hexadecimal String
+    func toHexString() -> String {
+        switch self {
+        case .data(let data):
+            return data.toHexString()
+        case .integer(let integer):
+            return integer.toHexString()
+        }
     }
 }
--- a/Sources/ClientRuntime/Networking/HashFunction.swift
+++ b/Sources/ClientRuntime/Networking/HashFunction.swift
@@ -1,0 +1,100 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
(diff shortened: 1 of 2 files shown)


Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,100 @@
...
+    // Create a Data type from a UInt32 value
+    init(fromUInt32 value: UInt32) {
+        var val = value // Create a mutable UInt32
+        self = Swift.withUnsafeBytes(of: &val) { Data($0) }
+    }
+
+    // Convert a Data type to UInt32
+    func toUInt32() -> UInt32? {
+        guard self.count == MemoryLayout<UInt32>.size else {
+            return nil // Ensures the data is of the correct size
```
waahm7, line 75: Discussed offline, I think it would be better if we could avoid converting back and forth between `UInt32` and `Data`. Depending on the context of where this is required, perhaps we can use algebraic data types or just convert it once to whatever data type is needed. Please feel free to let me know if there are any changes required in the CRT API to make it better.

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
```diff
@@ -0,0 +1,150 @@
...
+import AwsCommonRuntimeKit
+
+class HashFunctionTests: XCTestCase {
+
+    override func setUp() {
+        // Initialize function needs to be called before interacting with CRT
+        CommonRuntimeKit.initialize()
+    }
+
+    func testCRC32() {
```
waahm7, line 19: debatable: I prefer explicit test names like `testCRC32NonUTF8Bytes()`

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            XCTFail("CRC32 not found")
+            return
+        }
+
+        // Create test data
+        let testBytes: [UInt8] = [0xFF, 0xFE, 0xFD, 0xFC] // Bytes not valid in UTF-8
+        let testData = Data(testBytes)
+
+        let result = try? hashFunction.computeHash(of: testData)?.toUInt32()
+        let expected = testData.computeCRC32()
```
waahm7, line 30: Would be better to test on hard coded values. You can use something like https://crccalc.com/?crc=0xFFFEFDFC&method=crc32&datatype=hex&outtype=0 to get the values.

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,100 @@
...
+            return Data(fromUInt32: data.computeCRC32())
+        case .crc32c:
+            // Should turn this result back into a UInt32 using .toUInt32()
+            return Data(fromUInt32: data.computeCRC32C())
+        case .sha1:
+            do {
+                let hashed = try data.computeSHA1()
+                return Data(hashed)
+            } catch {
+                throw ClientRuntime.ClientError.unknownError("Error computing SHA1: \(error)")
```
waahm7, line 44: Would it possible to throw a more specific error than `unknownError`?

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: Changed as suggested -- HashResult has a .data and a .integer

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
dayaffe: Changed as suggested

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
dayaffe: Changed as suggested

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: Changed as suggested -- now throws HashError

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,105 @@
...
+            return true
+        default:
+            return false
+        }
+    }
+
+    func computeHash(of data: Data) throws -> HashResult? {
+        switch self {
+        case .crc32:
+            // Should turn this result back into a UInt32 using .toUInt32()
```
waahm7, line 44: outdated comment

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+extension UInt32: HexStringable {
+
+    // Convert a UInt32 type to a hexademical String
+    func toHexString() -> String {
+        return String(format: "%08x", self)
+    }
+}
```
waahm7, line 92: If these two extensions are only used for converting `HashResult` to a `HexString`, would it be better to just move this one line to that function?

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,105 @@
...
+
+extension HashResult: HexStringable {
+
+    // Convert a HashResult to a hexadecimal String
+    func toHexString() -> String {
+        switch self {
+        case .data(let data):
+            return data.toHexString()
+        case .integer(let integer):
+            return integer.toHexString()
```
waahm7, line 102: Continuing on my previous comment, we can do something like 
```suggestion
        case .data(let data):
            return map { String(format: "%02x", $0) }.joined()
        case .integer(let integer):
            return String(format: "%08x", self)
```

Sources/ClientRuntime/Networking/HashFunction.swift
waahm7: Debatable: We could also get rid of the protocol unless it will be used in multiple places.

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
```diff
@@ -0,0 +1,163 @@
...
+    override func setUp() {
+        // Initialize function needs to be called before interacting with CRT
+        CommonRuntimeKit.initialize()
+    }
+
+    func testCRC32NonUTF8Bytes() {
+        guard let hashFunction = HashFunction.from(string: "crc32") else {
+            XCTFail("CRC32 not found")
+            return
+        }
```
waahm7, line 23: Debatable: Feel free to ignore. 
I find it better and more concise to use `let hashFunction = HashFunction.from(string: "crc32")!` instead of a guard in tests. The test will fail anyway if we don't find it. Same comment for all the guards, `.?`, and `try?` in these tests.

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    func computeHash(of data: Data) throws -> HashResult? {
+        switch self {
+        case .crc32:
+            // Should turn this result back into a UInt32 using .toUInt32()
+            return .integer(data.computeCRC32())
+        case .crc32c:
+            // Should turn this result back into a UInt32 using .toUInt32()
```
waahm7, line 47: outdated comment

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -0,0 +1,105 @@
...
+    var isSupported: Bool {
+        switch self {
+        case .crc32, .crc32c, .sha256, .sha1:
+            return true
+        default:
+            return false
+        }
+    }
+
+    func computeHash(of data: Data) throws -> HashResult? {
```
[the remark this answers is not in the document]
dayaffe: changed to HashResult

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: changed as suggested

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: changed as suggested

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
dayaffe: removing `try?` causes error unhandled errors, the guard statements I prefer to keep in for precise error messaging should it ever fail

waahm7: Okay, by the way, you can replace `try?` with `try!` and `.?` with `.!` to avoid causing any errors.

jbelkins: Generally agreed, I am less of a stickler about force-unwrap in tests, though it can be annoying for the test suite to crash when it could have just failed a test & moved on.

If the `!` is expected to fail under any circumstance, I'd prefer to throw & fail versus crash the test suite.

jbelkins (approved):

Tests/ClientRuntimeTests/NetworkingTests/HashFunctionTests.swift
dayaffe: Good to know! For now I am going to continue with the guard so that it is abundantly clear what the issue is if we get a failure here

PR #647, opened 2024-01-18
feat: add support for flexible checksums on Normal payloads

dayaffe:
## Issue \#
[#1287](https://github.com/awslabs/aws-sdk-swift/issues/1287)

## Scope


Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+        guard let logger = context.getLogger() else {
+            return try await next.handle(context: context, input: input)
+        }
+
+        // Get supported list of checksums in priority order
+        let validationList = HashFunction.fromList(checksumAlgorithms).getPriorityOrderValidationList()
+        
+        // Skip flexible checksums workflow if no valid checksum algorithms are provided
+        guard let checksumAlgorithm: HashFunction = validationList.first else {
+            logger.info("Found no supported checksums! Skipping flexible checksums workflow...")
```
jbelkins, line 32: Not sure, should this be logged as an error?

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+                logger.debug("Calculating checksum")
+            }
+            
+            let checksum = try checksumAlgorithm.computeHash(of: data).toBase64String()
+            
+            request.updateHeader(name: headerName, value: [checksum])
+        }
+
+        func handleStreamPayload(_ stream: Stream) throws {
+            logger.error("Stream payloads are not yet supported with flexible checksums!")
```
jbelkins, line 66: Just checking, we do have a ticket for flexible checksums on streams, right?

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+        
+        // Handle body vs handle stream
+        switch request.body {
+        case .data(let data):
+            try handleNormalPayload(data)
+        case .stream(let stream):
+            try handleStreamPayload(stream)
+        case .noStream:
+            throw ClientError.dataNotFound("Cannot calculate the checksum of an empty body!")
```
jbelkins, line 77: Isn't there an "empty data checksum" for this case?  Or is that only for the case where body is `.data` and the data is 0 bytes?

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsResponseMiddleware.swift
```diff
@@ -0,0 +1,108 @@
...
+            logger.info("Checksum validation should not be performed! Skipping workflow...")
+            return try await next.handle(context: context, input: input)
+        }
+
+        // Get the response
+        let response = try await next.handle(context: context, input: input)
+        
+        // Determine if any checksum headers are present
+        logger.debug("HEADERS: \(response.httpResponse.headers)")
+        let _checksumHeader = CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST.first { response.httpResponse.headers.value(for: "x-amz-checksum-\($0)") != nil }
```
jbelkins, line 49: Underscores are usually reserved for private variables in Swift.  I think you can just drop the underscore

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
```diff
@@ -66,6 +66,8 @@ object ClientRuntimeTypes {
         val LoggerMiddleware = runtimeSymbol("LoggerMiddleware")
         val ContentLengthMiddleware = runtimeSymbol("ContentLengthMiddleware")
         val ContentMD5Middleware = runtimeSymbol("ContentMD5Middleware")
+        val FlexibleChecksumsRequestMiddleware = runtimeSymbol("FlexibleChecksumsRequestMiddleware")
+        val FlexibleChecksumsResponseMiddleware = runtimeSymbol("FlexibleChecksumsResponseMiddleware")
```
jbelkins, line 70: I don't see where these middlewares are ever added to the stack

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+                          input: SerializeStepInput<OperationStackInput>,
+                          next: H) async throws -> OperationOutput<OperationStackOutput>
+    where H: Handler,
+    Self.MInput == H.Input,
+    Self.MOutput == H.Output,
+    Self.Context == H.Context {
+        
+        // Initialize logger
+        guard let logger = context.getLogger() else {
+            return try await next.handle(context: context, input: input)
```
sichanyoo, line 24: Q: So if a logger wasn't set on the middleware context, we skip the flexible checksum request middleware? What's the reason?

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
sichanyoo, line 32: Q: Should an error be thrown instead if an operation is supposed to support checksum algorithm to use? Or is providing checksum algorithm always optional?

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+            
+            let checksum = try checksumAlgorithm.computeHash(of: data).toBase64String()
+            
+            request.updateHeader(name: headerName, value: [checksum])
+        }
+
+        func handleStreamPayload(_ stream: Stream) throws {
+            logger.error("Stream payloads are not yet supported with flexible checksums!")
+            return
+        }
```
sichanyoo, line 68: I think having these functions defined outside the handle() would be better, as that would allow shorter and more concise / readable handle()  - same thing for response middleware.

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsResponseMiddleware.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+        
+        // Exit if validation should not be performed
+        if (!validationMode) {
+            logger.info("Checksum validation should not be performed! Skipping workflow...")
+            return try await next.handle(context: context, input: input)
+        }
+
+        // Get the response
+        let response = try await next.handle(context: context, input: input)
```
sichanyoo, line 45: Wouldn't the response already be received & deserialized by the time we reach this point of this middleware, since we're placing the middleware _after_ the DeserializeMiddleware?

sichanyoo (changes requested):
Just some questions and suggestions!

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: Yeah let's log this as an error

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: Yeah I am working on aws-chunked encoding now which will then let me do the streams portion!

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: for which case? ByteStream is defined as:

```
case data(Data?)
case stream(Stream)
case noStream
```

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsResponseMiddleware.swift
dayaffe: I can make this variable private since the variable that is actually utilized elsewhere is the guard statement

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
dayaffe: Middleware are added in aws-sdk-swift PR

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: Good catch I meant to change this as I did in response to `throw ClientError.unknownError("No logger found!")`

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: Will change to logging an error. No need to stop the request at this point we can just "skip the workflow" as stated in the SEP

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: I defined them inside the `handle()` so that they would have access to all the variables defined there and avoid having to pass many variables around

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsResponseMiddleware.swift
sichanyoo: Note: Ongoing offline discussion on whether calling `next.handle` here and one more time later means executing request twice & if not, how calling `next.handle` here conveniently gives a copy of the response both times

dayaffe: After doing some research, and perhaps someone else can confirm, I do not believe calling next.handle() issues the request multiple times.

Evidence for this:
- The deserializeMiddleware is actually being added to the stack `before` subsequent middlewares such as FlexibleChecksumsResponseMiddleware. The line `let response = try await next.handle(...)` is a recursive call saying "come back to me with the response we get after the entire middleware stack is done". The next call in the stack that happens is going to FlexibleChecksumsResponseMiddleware where we again say `let response = try await next.handle(...)` and ask that the result of calling subsequent middlewares is given back to me here. Then the response should be issued and goes back to FlexibleChecksumsResponseMiddleware. And once we're done there then we go back to DeserializeMiddleware where eventually we return either a `throw try await httpResponseErrorClosure(copiedResponse.httpResponse)` or `return copiedResponse`. While this is **not** the last middleware added to the stack, it ends up being the last middleware since it is the first in the stack to ask for the response back **and** does not end up returning a `try await next.handle(...)` at the end.
- Inspecting the response in both places this is called I can see that the memory address of the stream in httpResponse is same
- Looking at the logs we would expect to see a completion handler execute more than once with the bytes streamed down before it if the same call to pull down a file was executed more than once, but this is not the case

**My Conclusion:** It is ok from a networking perspective to have response held in more than one middleware as long as it is after the DeserializeMiddleware in the stack. However, we are holding this thing in memory so it should be a stream with a known content length (not event streams).

sichanyoo (approved):
Lgtm, just need swiftlint warnings taken care of.

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -69,6 +91,43 @@ enum HashFunction {
...
+    /*
+     * Priority-order for validating checksum = [ CRC32C, CRC32, SHA1, SHA256 ]
+     * Order is determined by speed of the algorithm's implementation
+     * MD5 is not supported by list ordering
+     */
+    public static func < (lhs: HashFunction, rhs: HashFunction) -> Bool {
+        let order: [HashFunction] = [.crc32c, .crc32, .sha1, .sha256]
+
+        let lhsIndex = order.firstIndex(of: lhs) ?? Int.max
+        let rhsIndex = order.firstIndex(of: rhs) ?? Int.max
```
jbelkins, line 104: So `.md5` is always last when compared to an algorithm listed on L101?

Sources/ClientRuntime/Networking/HashFunction.swift
```diff
@@ -69,6 +91,43 @@ enum HashFunction {
...
+}
+
+extension UInt32 {
+    func toBase64EncodedString() -> String {
+        // Create a Data instance from the UInt32 value
+        var value = self
+        let data = Data(bytes: &value, count: MemoryLayout<UInt32>.size)
+
+        // Reverse bytes if on a little-endian architecture
+        let endianAdjustedData = Data(data.reversed())
```
jbelkins, line 124: How does this behave differently if on a big- or little-endian system?  From what I can tell, `reversed()` does the same on either.

jbelkins: I think you want to use `UInt32.littleEndian` or `UInt32.bigEndian` to get the byte order you want, no matter the platform.

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
jbelkins: Was this addressed?  Still looks like `logger.info()` to me

jbelkins (changes requested):
See my comments plus fix the lint issues.

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: md5 is not a valid flexible checksum as per the SEP so it will never appear as part of the requestAlgorithmMember. If flexible checksums workflow is skipped it should be the default used

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: Ah great point, Ill change this

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: Oops I think I might've reverted this change back accidentally, will change this

Sources/ClientRuntime/Networking/HashFunction.swift
jbelkins: From what I gather, Apple platforms (Apple Silicon & x86_64) are little-endian.  But when we start talking about proprietary ARM on Linux, it could be either... IDK

Sources/ClientRuntime/Networking/Http/Middlewares/FlexibleChecksumsRequestMiddleware.swift
dayaffe: @jbelkins Thinking about it some more, I believe this should be a logger.info as it is not an unexpected case. The FlexibleChecksumsRequestMiddleware should be added to any method that has the httpChecksum trait with a requestAlgorithmMember enum. But the user does not have to use checksums, it is an opt-in feature.

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: Adding an update to skip Content-MD5 calculation when using checksums, as per the SEP

Sources/ClientRuntime/Networking/HashFunction.swift
dayaffe: forcing one or the other using littleEndian vs bigEndian regardless seems to change the base64encoded output causing the request to fail against S3's servers. A solution I thought of was to reverse only if the system is littleEndian:

```
// Reverse bytes if on a little-endian architecture
if self == self.littleEndian {
    data = Data(data.reversed())
}
```

jbelkins: Calling `.littleEndian` should have no effect on the value on a little-endian system. Vice-versa for calling `.bigEndian` on a big-endian system.

Assuming you want the bytes in big-endian order:
```swift
let value: UInt32 = 123456  // or whatever value
let bigEndianValue = value.bigEndian
let data = Data(bytes: &bigEndianValue, count: MemoryLayout<UInt32>.size) 
```
`data` should be your integer's 4 bytes in big-endian order.

jbelkins (approved):

PR #656, opened 2024-01-30
fix: Make HttpResponse.statusCode getter & setter thread-safe

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1324
https://github.com/aws-amplify/amplify-swift/issues/3708

Addresses SIM ticket `/V1219649867`.

## Description of changes

A customer raised a SIM ticket with the following screenshot of data race warning on IDE:

image

To address this data race warning, use serial queue and hidden property `_statusCode` to make getter & setter thread-safe. Observed that the thread sanitization warning goes away when running CRT HTTP client with this change.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/HttpResponse.swift
+++ b/Sources/ClientRuntime/Networking/Http/HttpResponse.swift
@@ -7,22 +7,37 @@
 
 import protocol SmithyReadWrite.WireDataProviding
 import AwsCommonRuntimeKit
+import class Foundation.DispatchQueue
 
 public class HttpResponse: HttpUrlResponse, ResponseMessage {
 
     public var headers: Headers
     public var body: ByteStream
-    public var statusCode: HttpStatusCode
+
+    private var _statusCode: HttpStatusCode
+    private let statusCodeQueue = DispatchQueue(label: "statusCodeSerialQueue")
+    public var statusCode: HttpStatusCode {
+        get {
+            statusCodeQueue.sync {
+                return _statusCode
+            }
+        }
+        set {
+            statusCodeQueue.sync {
+                self._statusCode = newValue
+            }
+        }
+    }
 
     public init(headers: Headers = .init(), statusCode: HttpStatusCode = .processing, body: ByteStream = .noStream) {
         self.headers = headers
-        self.statusCode = statusCode
+        self._statusCode = statusCode
         self.body = body
     }
 
     public init(headers: Headers = .init(), body: ByteStream, statusCode: HttpStatusCode) {
         self.body = body
-        self.statusCode = statusCode
+        self._statusCode = statusCode
         self.headers = headers
     }
 }
--- a/Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
+++ b/Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
@@ -86,7 +86,7 @@ public extension DefaultSDKRuntimeConfiguration {
     static func makeClient(
         httpClientConfiguration: HttpClientConfiguration = defaultHttpClientConfiguration
     ) -> HTTPClient {
-        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS) //|| os(macOS)
+        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS) || os(macOS)
         return URLSessionHTTPClient(httpClientConfiguration: httpClientConfiguration)
         #else
         let connectTimeoutMs = httpClientConfiguration.connectTimeout.map { UInt32($0 * 1000) }
--- a/Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
+++ b/Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
@@ -86,7 +86,7 @@ public extension DefaultSDKRuntimeConfiguration {
     static func makeClient(
         httpClientConfiguration: HttpClientConfiguration = defaultHttpClientConfiguration
     ) -> HTTPClient {
-        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS) || os(macOS)
+        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS) //|| os(macOS)
         return URLSessionHTTPClient(httpClientConfiguration: httpClientConfiguration)
         #else
         let connectTimeoutMs = httpClientConfiguration.connectTimeout.map { UInt32($0 * 1000) }


jbelkins (changes requested):
Two things:
- It is not enough to just do the writes on the serial queue.  Writing on the serial queue ensures that writes will not happen at the same time, but does not guarantee that reads will not happen at the same time as writes.  To fully ensure no simultaneous access, reads would have to be performed on the queue as well.
- You create the serial queue as a local variable each time that the private function `safelySetStatusCode` is called.  That means that every write to status code will happen on its own queue, and nothing will actually be deconflicted.  To ensure that all writes go onto the same queue, the queue should be a property of either the request or the client engine.

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
```diff
@@ -4,12 +4,24 @@
...
 
     public var headers: Headers
     public var body: ByteStream
     public var statusCode: HttpStatusCode
 
+    public func addHeaders(additionalHeaders: Headers) {
+        self.headers.addAll(headers: additionalHeaders)
+    }
+
+    public func setBody(newBody: ByteStream) {
```
jbelkins, line 17: Does this function not need to be marked `async`?  We call it with `async`...

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
```diff
@@ -23,7 +35,7 @@ public class HttpResponse: HttpUrlResponse {
     }
 }
 
-extension HttpResponse: CustomDebugStringConvertible {
+extension HttpResponse {
```
jbelkins, line 38: Presuming this protocol was removed because it requires a non-async method or var.

Is the rest of this extension of any use anymore?

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
```diff
@@ -27,7 +27,11 @@ public struct DeserializeMiddleware<OperationStackOutput>: Middleware {
 
             let response = try await next.handle(context: context, input: input) // call handler to get http response
             var copiedResponse = response
-            if (200..<300).contains(response.httpResponse.statusCode.rawValue) {
+            while (100..<200).contains(await response.httpResponse.statusCode.rawValue) {
+                // Wait until http response status code gets set to either success or error
+                try await Task.sleep(nanoseconds: 10)
```
jbelkins, line 32: Is this the change that was required for tests to pass on Linux?

jbelkins (commented):
A couple questions inline.  Most of PR is good - just adding `await` where needed

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
sichanyoo: Unless I'm mistaken, I don't think the functions in an actor have to be marked async unless they call async functions themselves. In this case the setter functions don't call anything async themselves (they just set the properties of the actor instance), so the async keyword can be left out. Though when we call these methods outside of the actor, we need to prepend `await` keyword since the action gets put into actor's queue of tasks that gets completed in async manner.

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
sichanyoo: Yep the protocol conformance was removed because the computed property `debugDescription` is now actor-isolated and can't fulfill the protocol requirement. But the properties are still used in other places so I left the extension as is.

Sources/ClientRuntime/Networking/Http/Middlewares/DeserializeMiddleware.swift
sichanyoo: Yeah, in Linux the integration tests failed sporadically because while the status code is 102 (which indicates to client that a full request has been received and the server is working on it), the code progresses to deserialization which caused it to throw errors. By making sure the status code is one of the following: success, fail, error, this sporadic failure was fixed.

jbelkins: Looping is not the preferred way to handle a case like this, because it will result in extra CPU being consumed and make the rest of the app less responsive, even when an async wait is performed as you have here (10ns is way too short BTW.)

A better way is to provide a method on `HttpResponse` that will suspend execution until the status code changes to the desired value.

You would do that with a callback like this (code is over-simplified and maybe not syntactically correct, but gives you the feel for it):
```swift
    var continuation = CheckedContinuation<Void>?

    func waitForFinalStatus() async {
        return await withCheckedContinuation { continuation in
            self.continuation = continuation
        }
    }

    func setStatusCode(_ statusCode: Int) {
        self.statusCode = statusCode
        if code >= 200 {
            continuation?.resume()
        }
    }
```

jbelkins: (I fixed a similar polling solution that we used on data streams last summer in this manner)

jbelkins (changes requested):
See my suggestion on a more efficient way to resume execution when status code changes to a final value

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
```diff
@@ -19,7 +20,20 @@ public actor HttpResponse: HttpUrlResponse {
     }
 
     public func setStatusCode(newStatusCode: HttpStatusCode) {
+        let codeBeforeUpdate = self.statusCode.rawValue
         self.statusCode = newStatusCode
+        if newStatusCode.rawValue >= 200 && codeBeforeUpdate < 200 {
+            self.continuation?.resume()
```
jbelkins, line 26: Since a continuation must be called exactly once, it is a good idea to set it to `nil` after calling `resume` on it.

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
```diff
@@ -19,7 +20,20 @@ public actor HttpResponse: HttpUrlResponse {
...
+        }
+    }
+
+    public func waitForFinalStatusCode() async {
```
jbelkins, line 30: Would it be useful to actually return the status code once a code 200 or higher is received?

jbelkins (changes requested):
See comments.  Also, remember that when a continuation is used, whether the result is success or failure, it must be resumed exactly once.

So make sure that every possible path that this request could take (success, failure, cancellation before completion, etc.) results in the continuation being called.  As-is, I don't see any resumption of the continuation except on the success path.

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
sichanyoo: Changed as suggested.

Sources/ClientRuntime/Networking/Http/HttpResponse.swift
sichanyoo: Changed as suggested.

sichanyoo:
@jbelkins RE: making sure continuation gets resumed exactly once for all code paths: 
For any situation where HTTP response is returned from the server (with any status code >= 200), it's resumed exactly once.
And for any situation that involves non-HTTP error such as the connection being dropped due to a connection error, those errors are caught by the `RetryMiddleware`, which is then thrown back to the user without the code execution progressing into `DeserializeMiddleware` (so the continuation doesn't even get instantiated).

jbelkins (approved):
This is correct as far as thread-safety.  Not sure how I feel about using `DispatchQueue` though - CRT generally does not use it, and our projects only use it where needed to interact with Foundation APIs that are based on it.

Since this fixes the problem though, I'm approving and we can revisit it later.

PR #657, opened 2024-02-05
feat: Add setter for CRT logger to allow users to customize its log level

sichanyoo:
Adds a setter for the logger used by CRT IO.

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1131

## Description of changes

Currently the log level of the logger for CRT HTTP client engine is `.none`, but there are instances where we need CRT log from the user to investigate failure cause(s). One such instance, which prompted the linked issue, is https://github.com/awslabs/aws-crt-swift/issues/197. 

## Scope


Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -94,7 +94,10 @@ public extension DefaultSDKRuntimeConfiguration {
         return URLSessionHTTPClient(httpClientConfiguration: httpClientConfiguration)
         #else
         let connectTimeoutMs = httpClientConfiguration.connectTimeout.map { UInt32($0 * 1_000_000) }
-        let config = CRTClientEngineConfig(connectTimeoutMs: connectTimeoutMs)
+        let config = CRTClientEngineConfig(
```
jbelkins, line 97: The config object to which the CRT `logLevel` field was added is specifically to configure the CRT HTTP client.

Unless CRT only logs from the HTTP client (which I don't think is the case) then the CRT log level should be set elsewhere.

SRA has a Logging section, does it talk about logger configuration?

jbelkins (changes requested):

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
sichanyoo: The config exposed via HTTP client configuration gets used in CRT client engine initialization, where it re-assigns a logger to have the config log level value if the user-provided config value is anything other than `.none`. Given our SDK statically initializes `SDKDefaultIO` and shares it until program closes, we'd be missing CRT log from CRT resource initializations, but any CRT log after initialization would be logged properly.

jbelkins: Remember that the CRT client engine is not always used.  In that case, the logger would not get configured at all, right?

sichanyoo: Hm... come to think of it - if the user doesn't use the CRT http client then log level won't be set to user-provided value, so for users on mac that use the URLSession client, CRT log would still be .none.... I think I do need to find a earlier place in program execution to allow user to configure log level for CRT.

dayaffe (approved):

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
sichanyoo: Changed the approach from exposing a config value to adding a setter.

jbelkins (approved):

PR #672, opened 2024-02-26
feat: add support to CRT HTTP Client for providing custom trust and key store

dayaffe:
## Issue \#
[#1110](https://github.com/awslabs/aws-sdk-swift/issues/1110)
## Description of changes

Example Usage with STSClient on Apple devices:
```
let urlSessionTLSOptions = URLSessionTLSOptions(
    certificateFile: "localhost", // .cer
    keyStoreName: "certificate", // .p12
    keyStorePassword: "test"
)

let client = try await STSClient(
    config: .init(
        httpClientConfiguration: .init(
            tlsOptions: .init(urlSessionTLSOptions: urlSessionTLSOptions)
        )
    )
)
```

Example Usage with STSClient on Linux:
```
let crtTLSOptions = CRTClientTLSOptions(
    certificatePath: "/path/",
    certificateFilename: "trustStoreFile",
    keyStoreFilepath: "/path/file",
    keyStorePassword: "password"
)

let client = try await STSClient(
    config: .init(
        httpClientConfiguration: .init(
            tlsOptions: .init(crtTLSOptions: crtTLSOptions)
        )
    )
)
```

### Directions for testing URLSession changes:

Step 1: Downgrade Openssl on mac from 3.x to 1.1. Mac's Keychain Access and XCode are not compatible with the latest openssl encryption algorithms

Step 2. Generate a key (.pem) and a certificate (.cer) and then include both in a keystore (.p12)

Step 3. Include both the certificate file (.cer) and keystore (.p12) in the project's bundle

Step 4. Create a local test server. You can use mine as an exmaple and run it using `node server.js`:
```
const express = require('express');
const https = require('https');
const fs = require('fs');

const app = express();

app.get('/', (req, res) => {
  res.send('Hello, Secure World!');
});

https.createServer({
  key: fs.readFileSync('key.pem'),
  cert: fs.readFileSync('certificate.pem'),
  passphrase: "test"
}, app)
.listen(3000, () => {
  console.log('Listening on https://localhost:3000');
});
```

Step 5. Run XCode proj that hits the server. Here is my example code:
```
let urlSessionTLSOptions = URLSessionTLSOptions(
    certificateFile: "localhost", // .cer
    keyStoreName: "certificate", // .p12
    keyStorePassword: "test"
)

// Create an instance of our URLSessionHTTPClient
let urlSessionClient = URLSessionHTTPClient(httpClientConfiguration: .init(urlSessionTLSOptions: urlSessionTLSOptions))

// Create an SdkHttpRequest
let sdkHttpRequest = SdkHttpRequest(
    method: .get,
    endpoint: Endpoint(host: "localhost", path: "/", port: 3000, protocolType: .https),
    body: .noStream
)
// Send the request
do {
    let response = try await urlSessionClient.send(request: sdkHttpRequest)
    print("Response received: \(response)")
} catch {
    print("Failed to make request: \(error)")
}
```

## Scope


Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
```diff
@@ -121,8 +127,81 @@ public final class URLSessionHTTPClient: HTTPClient {
...
+            completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void
+        ) {
+            guard let tlsOptions, tlsOptions.useProvidedKeystore,
+                  let keystoreName = tlsOptions.keyStoreName,
+                  let keystorePasword = tlsOptions.keyStorePassword,
+                  let identity = Bundle.main.identity(named: keystoreName, password: keystorePasword) else {
+                logger.debug(
+                    "Either TLSOptions not set or missing values or certificate is not found! " +
+                    "Using default key store."
+                )
```
sichanyoo, line 180: Same comment as above. May be good to factor the guard statements out to a `private func ... -> Bool`.

sichanyoo (changes requested):
Minor suggestions

Sources/ClientRuntime/Networking/Http/HttpClientConfiguration.swift
```diff
@@ -34,6 +35,18 @@ public class HttpClientConfiguration {
...
+    ///
+    /// Enables specifying client certificates and trust stores for secure communication.
+    /// Defaults to system's TLS settings if `nil`.
+    public var tlsContext: TLSContext?
+
+    /// Custom TLS configuration for HTTPS connections with URLSession client.
+    ///
+    /// Enables specifying client certificates and trust stores for secure communication.
+    /// Defaults to system's TLS settings if `nil`.
+    public var urlSessionTLSOptions: URLSessionTLSOptions?
```
waahm7, line 48: If this is the public API, Is it possible to have one TLSContext class and create either a URLSessionTLSOptions or a TLSContext internally from it?

waahm7: If this is a public API, it might be better to just expose a wrapper to TLSContextOptions and create a TLSContext from that. No one should be passing server mode to this TLSContext.init function.

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionTLSOptions.swift
```diff
@@ -0,0 +1,37 @@
...
+    public var keyStoreName: String?
+
+    /// Password for the key store if required.
+    public var keyStorePassword: String?
+
+    /// Information is provided to use custom trust store
+    public var useSelfSignedCertificate: Bool
+
+    /// Information is provided to use custom key store
+    public var useProvidedKeystore: Bool
```
waahm7, line 23: These are derived variables, but someone can modify them after the init function. Is that intended?

Sources/ClientRuntime/Networking/Http/HttpClientConfiguration.swift
```diff
@@ -34,6 +35,18 @@ public class HttpClientConfiguration {
...
+
```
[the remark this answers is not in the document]
dayaffe: Will create a wrapper class as suggested by @waahm7

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionTLSOptions.swift
dayaffe: Will change so they cant modify

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
```diff
@@ -330,4 +413,71 @@ public enum URLSessionHTTPClientError: Error {
...
+
+    func identity(named name: String, password: String) -> SecIdentity? {
+        guard let p12URL = self.url(forResource: name, withExtension: "p12"),
+              let p12Data = try? Data(contentsOf: p12URL) else {
+            return nil
+        }
+
+        let options = [kSecImportExportPassphrase as String: password] as CFDictionary
+        var items: CFArray?
+        let status = SecPKCS12Import(p12Data as CFData, options, &items)
```
xiazhvera, line 433: `SecPKCS12Import`: The latest `aws-crt-swift` disabled the API for iOS, tvOS, and watchOS, as we dont really have a way to test the API on those platforms. Would you have a test plan to the feature with pkcs12 key on the platforms? Maybe I could add the tests to CRT as well and enable the API there. 

https://github.com/awslabs/aws-crt-swift/pull/246 <= PR that disabled `SecPKCS12Import`

jbelkins (changes requested):
See individual comments.  Also it seems that we have some HTTP client config that is specific to only one client.  Short of clearly documenting what affects what, is there a way that we can unify settings so that one setting affects both clients?

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
dayaffe: I agree there's no easy way to test on those platforms natively (would work via simulator for dev testing), I manually generated the p12 file using a combination of keychain access and openssl 1.1

sichanyoo (approved):

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
```diff
@@ -330,4 +425,71 @@ public enum URLSessionHTTPClientError: Error {
     case unresumedConnection
 }
 
+extension Bundle {
```
jbelkins, line 428: Let's move these extensions to a separate source file

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
```diff
@@ -121,8 +127,93 @@ public final class URLSessionHTTPClient: HTTPClient {
...
+
+        /// The URLSession delegate method where authentication challenges are handled.
+        func urlSession(
+            _ session: URLSession,
+            task: URLSessionTask,
+            didReceive challenge: URLAuthenticationChallenge,
+            completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void
+        ) {
+            switch challenge.protectionSpace.authenticationMethod {
+            case NSURLAuthenticationMethodServerTrust:
```
jbelkins, line 210: nit: These cases do the same thing, merge them into one

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
```diff
@@ -330,4 +425,71 @@ public enum URLSessionHTTPClientError: Error {
...
+              let identity = firstItem[kSecImportItemIdentity as String] else {
+            return nil
+        }
+
+        // Explanation of cross-platform behavior differences:
+        // - Linux and macOS treat Core Foundation types differently.
+        // - The `as? SecIdentity` casting causes a compiler error on Apple platforms as the cast is guaranteed.
+        // - Directly returning `identity` works on Linux but not on macOS due to strict type expectations.
+        // SwiftLint is temporarily disabled for the next line to allow a force cast, acknowledging the platform-specific behavior.
+        // swiftlint:disable:next force_cast
```
jbelkins, line 458: Eventually Swift Foundation will eliminate these differences, but for now we're stuck with `swift-corelibs-foundation` on Linux which is not always faithful to the original.

jbelkins (changes requested):

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionHTTPClient.swift
dayaffe: oops this is actually an issue I must've introduced recently, thanks for catching -- one should be serverTrustChallenge and one should be clientIdentityChallenge

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -39,13 +39,15 @@ public class CRTClientEngine: HTTPClient {
         private var http2ConnectionPools: [ConnectionPoolID: HTTP2StreamManager] = [:]
         private let sharedDefaultIO = SDKDefaultIO.shared
         private let connectTimeoutMs: UInt32?
+        private let tlsContext: TLSContext?
```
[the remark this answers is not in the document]
dayaffe: Will change this to CRTClientTLSOptions type that I created and provide a resolveContext() method on it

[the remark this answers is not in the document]
dayaffe: Can also rename that if we want to get rid of CRT from it

waahm7 (approved):

Sources/ClientRuntime/Networking/Http/CRT/CRTClientTLSOptions.swift
```diff
@@ -0,0 +1,91 @@
...
+                #if os(tvOS) || os(iOS) || os(watchOS) || os(macOS)
+                if let path = keyStoreFilepath, let password = keyStorePassword {
+                    tlsOptions = try .makeMTLS(pkcs12Path: path, password: password)
+                }
+                #else
+                if let certPath = certificatePath,
+                    let certFilename = certificateFilename,
+                    let privateKeyPath = privateKeyFilepath {
+                    let certFilepath = "\(certPath)/\(certFilename)"
+                    tlsOptions = try .makeMTLS(certificatePath: certFilepath, privateKeyPath: privateKeyPath)
```
xiazhvera, line 72: `makeMTLS(certificatePath, privateKeyPath)` is also available on MacOS. The current implementation would fall back into default `tlsOptions` if the user tries to setup `certFilepath` and `privateKeyPath` on MacOS. 
I'm not sure if that's on purpose or not. If so, we probably wanna add comments to document it.

dayaffe: Which should be used for visionOS?

dayaffe: Ran latest version by @xiazhvera who says it looks good!

PR #680, opened 2024-03-13
feat: Add eventstream input support for protocols that use XML for requests

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1379

Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1404

## Description of changes

Currently the Swift SDK does not support event stream input for protocols that use XML as request wire protocol (e.g., restXml).
Fortunately, currently there is no AWS service that uses restXml AND uses event stream input for an operation in their API either.
However, if any service that uses restXml adds an operation that takes in an event stream input in the future, our SDK would break the customers.

This PR and the companion PR in `aws-sdk-swift` linked above do the groundwork for supporting event stream input for protocols that use XML as request wire protocol (e.g., restXml).

Detailed descriptions of each change will be put as comments on the Github diff.

## Scope


Sources/ClientRuntime/EventStream/DefaultMessageEncoderStream.swift
```diff
@@ -13,42 +13,42 @@ extension EventStream {
         let stream: AsyncThrowingStream<Event, Error>
```
sichanyoo, line 13: Here, the usage of `requestEncoder` is completely replaced by `marshalClosure`. This has to be done because XML serialization does not rely on Swift's Codable interface. For JSON serialization, the closure passed here will use `JSONEncoder` behind the scenes. For XML serialization, the closure passed here will use `XMLWriter` behind the scenes.

Sources/ClientRuntime/EventStream/MessageMarshallable.swift
```diff
@@ -16,3 +16,16 @@ public protocol MessageMarshallable {
     /// - Returns: The marshalled `Message`.
```
sichanyoo, line 16: Defines a new typealias for closure that takes a stream struct, determines the specific instance of `Event`, and returns an `EventStream.Message` that has values "serialized" into it from the original `Event` for generic handling and encoding into Data.

A utility method `jsonMarshalClosure` is defined here that returns a closure which simply uses the old `marshall` method - that gets code-generated for each streaming union shape within the `MessageMarshallable` protocol conformance - to return an instance of `EventStream.Message`.

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/EventStreamBodyMiddleware.swift
```diff
@@ -17,13 +17,16 @@ public struct EventStreamBodyMiddleware<OperationStackInput,
```
sichanyoo, line 17: Instead of getting the `requestEncoder` from middleware context and passing it along to `DefaultMessageEncoderStream`, the middleware now takes in a `marshalClosure` at initialization and passes it along to `DefaultMessageEncoderStream`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/OperationInputBodyMiddleware.kt
```diff
@@ -21,7 +21,9 @@ import software.amazon.smithy.swift.codegen.SwiftWriter
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
```
sichanyoo, line 21: Depending on the request wire protocol, either pass in the utility method defined for JSON, or pass in the newly-added code-generated `marshal` static function variable of the streaming shape.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/xml/StructDecodeXMLGenerator.kt
```diff
@@ -38,7 +38,7 @@ open class StructDecodeXMLGenerator(
             SmithyXMLTypes.Reader
```
sichanyoo, line 38: This additional condition fixes the error thrown when handling `EndEvent` that's returned as part of the event stream output of the call to `S3:SelectObjectContent`. The [S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html) say the payload of this event is empty (not an empty XML object, just literally empty), which causes an issue with deserializing in SDK side. The additional condition checks if the struct being deserialized has any properties, and if it doesn't, returns a new instance rather than returning nil.

jbelkins (approved):

PR #683, opened 2024-03-14
feat: Add socketTimeout to HTTP client config for closing stale connections

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1349

Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1413

## Description of changes

- Adds a new config `socketTimeout`, which gets used to construct `HTTPMonitoringOptions` for `CRTClientEngine`. Sets the maximum time to wait between data packets, which is used to close stale connections.
- Fixes `URLSessionHTTPClient` to use `socketTimeout` for its `timeoutIntervalForRequest` config instead of `connectTimeout`. Refer to this [official doc form Apple](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408259-timeoutintervalforrequest), which confirms that `timeoutIntervalForRequest` maps to the new `socketTimeout` rather than `connectTimeout`, which is a connection timeout config.
- Adds missing `continuation.resume()` that makes the application hang when socket timeout error is thrown.
- Misc. comment change for clarification on what `connectTimeout` actually is for. It's not a socket timeout for limiting stale connections. It's used to limit time it takes to obtain a live connection to the server.

## Prerequisite
- Blocked on the aws-crt-swift PR: https://github.com/awslabs/aws-crt-swift/pull/242, which makes initializer for `HTTPMonitoringOptions` public for usage in smithy-swift.
- Blocked on bumping aws-crt-swift version to the new release that contains changes from above PR.

## Testing
Tested against localhost that mocks DynamoDB & does not return a response. HTTP client connection now closes and recovers from no server response as it should, instead of hanging for eternity.

## Scope


Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -338,6 +346,7 @@ public class CRTClientEngine: HTTPClient {
                 response.statusCode = makeStatusCode(statusCode)
             case .failure(let error):
                 self.logger.error("Response encountered an error: \(error)")
+                continuation.resume(throwing: error)
```
jbelkins, line 349: good catch

Sources/ClientRuntime/Networking/Http/HttpClientConfiguration.swift
```diff
@@ -9,11 +9,18 @@ import struct Foundation.TimeInterval
...
     ///
     /// If none is provided, the client will use default values based on the platform.
     public var connectTimeout: TimeInterval?
 
+    /// The timeout for socket, in seconds.
+    /// Sets maximum time to wait between two data packets.
+    /// Used to close stale connections that have no activity.
+    ///
+    /// If no value is provided, the defaut client won't have a socket timeout.
+    public var socketTimeout: TimeInterval?
```
jbelkins, line 22: URLSession does not have separate settings for timeout on initial connection, and for idle time between data transmissions on an existing connection.

https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1408259-timeoutintervalforrequest

seems to be the setting for both and the two cannot be set individually.
Our doc comments should make clear which settings apply to which HTTP client and which are ignored.

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionConfiguration+HTTPClientConfiguration.swift
```diff
@@ -13,8 +13,8 @@ extension URLSessionConfiguration {
 
     public static func from(httpClientConfiguration: HttpClientConfiguration) -> URLSessionConfiguration {
         let config = URLSessionConfiguration.default
-        if let connectTimeout = httpClientConfiguration.connectTimeout {
-            config.timeoutIntervalForRequest = connectTimeout
+        if let socketTimeout = httpClientConfiguration.socketTimeout {
+            config.timeoutIntervalForRequest = socketTimeout
```
jbelkins, line 17: See my note above.  How we transfer the HTTP config fields to specific clients should be made clear in the doc comments.

jbelkins (changes requested):
Technically this looks good.  But we need to make sure through doc comments on the related config properties that customers understand what fields have what effect on the two supported HTTP clients.

Sources/ClientRuntime/Networking/Http/HttpClientConfiguration.swift
sichanyoo: Found a separate setting for initial connection timeout in URLRequest: https://developer.apple.com/documentation/foundation/urlrequest/2011509-timeoutinterval/

connectTimeout config on HttpClientConiguration is now set for URLRequest:timeoutInterval.

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionConfiguration+HTTPClientConfiguration.swift
sichanyoo: Both config fields are now used for same purposes in both clients, so I think existing comments do their job without exposing arbitrary details.

jbelkins (approved):

PR #685, opened 2024-03-22
feat: add interceptors interfaces

milesziemer:
Add interceptors interfaces to client libraries.

Interceptors defines hooks that can be used to inject functionality into different phases of the operation execution pipeline.

## Description of changes

- Adds the `Interceptor` protocol, which has methods defining the hooks into the operation execution pipeline. Protocol implementations can specify concrete types for the transport request/response types (i.e. `SdkHttpRequest`/`HttpResponse`), as well as a type that provides `Attributes` through the new `HasAttributes` protocol. This allows 'less specific' interceptors that don't care about the transport request/response types, or the available attributes, to be used with more specific interceptors. See an example in [InterceptorTests](Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift)
- Adds `HttpInterceptor`, which inherits from `Interceptor` and specializes the request, response, and attributes types to be http specific. Interceptor implementations can use this protocol so they don't have to define `RequestType`, `ResponseType`, and `AttributesType` if they need all http specific stuff.
- Adds the `InterceptorContext` protocol and various child protocols that define which properties of the interceptor context can be accessed at different points in the execution pipeline. Each interceptor hook accepts one of these child protocols as input. The context protocols have the same associated types as `Interceptor`, so implementations can define the types they will have in the context.
- Adds the `HasAttributes` protocol, which is basically a protocol version of the `Attributes` struct, and implements it for `HttpContext`. This lets us wrap `HttpContext` into interceptor context objects, and use it in interceptor implementations, much like how it is used in existing middleware.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
+++ b/Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
@@ -7,11 +7,13 @@
 
 /// Type-erased, concrete interceptor.
 /// 
-/// This structure stores 
-public struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttributes> {
-    public typealias InterceptorContextType = DefaultInterceptorContext<RequestType, ResponseType, AttributesType>
-    public typealias InterceptorFn = (InterceptorContextType) async throws -> Void
-    public typealias MutatingInterceptorFn = (inout InterceptorContextType) async throws -> Void
+/// In order to have multiple interceptors hooked into a single operation, we
+/// need a concrete type, not a protocol. This stores references to the closures
+/// of interceptor implementations and delegates to them for each interceptor hook.
+internal struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttributes> {
+    internal typealias InterceptorContextType = DefaultInterceptorContext<RequestType, ResponseType, AttributesType>
+    internal typealias InterceptorFn = (InterceptorContextType) async throws -> Void
+    internal typealias MutatingInterceptorFn = (inout InterceptorContextType) async throws -> Void
 
     private var readBeforeExecution: InterceptorFn
     private var modifyBeforeSerialization: MutatingInterceptorFn
@@ -33,7 +35,7 @@ public struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttri
     private var modifyBeforeCompletion: MutatingInterceptorFn
     private var readAfterExecution: InterceptorFn
 
-    public init<I: Interceptor>(interceptor: I)
+    internal init<I: Interceptor>(interceptor: I)
     where
         I.RequestType == RequestType,
         I.ResponseType == ResponseType,
@@ -59,61 +61,79 @@ public struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttri
         self.readAfterExecution = interceptor.readAfterExecution(context:)
     }
 
-    public func readBeforeExecution(context: InterceptorContextType) async throws {
+    internal func readBeforeExecution(context: InterceptorContextType) async throws {
         try await self.readBeforeExecution(context)
     }
-    public func modifyBeforeSerialization(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeSerialization(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeSerialization(&context)
     }
-    public func readBeforeSerialization(context: InterceptorContextType) async throws {
+
+    internal func readBeforeSerialization(context: InterceptorContextType) async throws {
         try await self.readBeforeSerialization(context)
     }
-    public func readAfterSerialization(context: InterceptorContextType) async throws {
+
+    internal func readAfterSerialization(context: InterceptorContextType) async throws {
         try await self.readAfterSerialization(context)
     }
-    public func modifyBeforeRetryLoop(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeRetryLoop(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeRetryLoop(&context)
     }
-    public func readBeforeAttempt(context: InterceptorContextType) async throws {
+
+    internal func readBeforeAttempt(context: InterceptorContextType) async throws {
         try await self.readBeforeAttempt(context)
     }
-    public func modifyBeforeSigning(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeSigning(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeSigning(&context)
     }
-    public func readBeforeSigning(context: InterceptorContextType) async throws {
+
+    internal func readBeforeSigning(context: InterceptorContextType) async throws {
         try await self.readBeforeSigning(context)
     }
-    public func readAfterSigning(context: InterceptorContextType) async throws {
+
+    internal func readAfterSigning(context: InterceptorContextType) async throws {
         try await self.readAfterSigning(context)
     }
-    public func modifyBeforeTransmit(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeTransmit(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeTransmit(&context)
     }
-    public func readBeforeTransmit(context: InterceptorContextType) async throws {
+
+    internal func readBeforeTransmit(context: InterceptorContextType) async throws {
         try await self.readBeforeTransmit(context)
     }
-    public func readAfterTransmit(context: InterceptorContextType) async throws {
+
+    internal func readAfterTransmit(context: InterceptorContextType) async throws {
         try await self.readAfterTransmit(context)
     }
-    public func modifyBeforeDeserialization(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeDeserialization(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeDeserialization(&context)
     }
-    public func readBeforeDeserialization(context: InterceptorContextType) async throws {
+
+    internal func readBeforeDeserialization(context: InterceptorContextType) async throws {
         try await self.readBeforeDeserialization(context)
     }
-    public func readAfterDeserialization(context: InterceptorContextType) async throws {
+
+    internal func readAfterDeserialization(context: InterceptorContextType) async throws {
         try await self.readAfterDeserialization(context)
     }
-    public func modifyBeforeAttemptCompletion(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeAttemptCompletion(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeAttemptCompletion(&context)
     }
-    public func readAfterAttempt(context: InterceptorContextType) async throws {
+
+    internal func readAfterAttempt(context: InterceptorContextType) async throws {
         try await self.readAfterAttempt(context)
     }
-    public func modifyBeforeCompletion(context: inout InterceptorContextType) async throws {
+
+    internal func modifyBeforeCompletion(context: inout InterceptorContextType) async throws {
         try await self.modifyBeforeCompletion(&context)
     }
-    public func readAfterExecution(context: InterceptorContextType) async throws {
+
+    internal func readAfterExecution(context: InterceptorContextType) async throws {
         try await self.readAfterExecution(context)
     }
 }
--- a/Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
+++ b/Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
@@ -13,26 +13,25 @@
 internal struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttributes> {
     internal typealias InterceptorContextType = DefaultInterceptorContext<RequestType, ResponseType, AttributesType>
     internal typealias InterceptorFn = (InterceptorContextType) async throws -> Void
-    internal typealias MutatingInterceptorFn = (inout InterceptorContextType) async throws -> Void
 
     private var readBeforeExecution: InterceptorFn
-    private var modifyBeforeSerialization: MutatingInterceptorFn
+    private var modifyBeforeSerialization: InterceptorFn
     private var readBeforeSerialization: InterceptorFn
     private var readAfterSerialization: InterceptorFn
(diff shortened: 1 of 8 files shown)


Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public func modifyBeforeSerialization(context: inout some MutableInput<Self.AttributesType>) async throws {
+            var input: InputType = context.getInput()!
+            input[keyPath: keyPath] = value
+            context.updateInput(updated: input)
+        }
+    }
+
+    struct AddHeaderInterceptor : HttpInterceptor {
```
syall, line 44: nit: Add a test interceptor that implements multiple interceptor hooks, e.g. both `modifyBeforeSerialization` and `modifyBeforeTransmit` in one definition

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
```diff
@@ -0,0 +1,119 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Type-erased, concrete interceptor.
+/// 
+/// This structure stores
```
syall, line 10: Documentation is incomplete.

Sources/ClientRuntime/Middleware/Attribute.swift
```diff
@@ -41,3 +41,22 @@ public struct Attributes {
         attributes.removeValue(forKey: key.name)
     }
 }
+
+/// A type that can be used as a type-safe property bag.
+public protocol HasAttributes: AnyObject {
```
syall, line 46: Is there a reason why we don't use `Attributes` directly? Either way, `Attributes` should be marked with `: HasAttributes` since it does implement the protocol.

Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift
```diff
@@ -0,0 +1,86 @@
...
+        var interceptorContext = DefaultInterceptorContext<SdkHttpRequest, HttpResponse, HttpContext>(input: input, attributes: httpContext)
+        let addAttributeInterceptor = AddAttributeInterceptor<String, SdkHttpRequest, HttpResponse, HttpContext>(key: AttributeKey(name: "foo"), value: "bar")
+        let modifyInputInterceptor = ModifyInputInterceptor<TestInput, SdkHttpRequest, HttpResponse, HttpContext>(keyPath: \.property, value: "bar")
+        let addHeaderInterceptor = AddHeaderInterceptor(headerName: "foo", headerValue: "bar")
+
+        let interceptors: [AnyInterceptor<SdkHttpRequest, HttpResponse, HttpContext>] = [
+            addAttributeInterceptor.erase(), 
+            modifyInputInterceptor.erase(), 
+            addHeaderInterceptor.erase()
+        ]
```
syall, line 72: Will customers implementing interceptors need to call `erase()`?

syall (changes requested):

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
milesziemer: Oops. Thanks

Sources/ClientRuntime/Middleware/Attribute.swift
milesziemer: Yea, so the reason is that I want interceptor implementations to be able to specify a more specific thing that 'has attributes', which may have helper functions for getting specific attributes. For example, `HttpContext` is now a `HasAttributes`, so in `HttpInterceptor`s, you can say 
```
let authSchemeResolver = context.getAttributes().getAuthSchemeResolver()
```
instead of
```
let authSchemeResolver = context.getAttributes().get(key: AttributeKeys.authSchemeResolver)
```
This makes it easier to adapt middleware that uses `HttpContext` to interceptors.

milesziemer: Also, `Attributes` doesn't implement `HasAttributes` because `HasAttributes` can only be implemented by a reference type, i.e. a class, which is why it has the `: AnyObject`. The reason I did this was because interceptors need to mutate the attributes, and if attributes isn't a reference type, you'd need to do
```
var attrs = context.getAttributes()
attrs.set(...)
context.updateAttributes(attrs)
```

syall (approved):

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
```diff
@@ -0,0 +1,139 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Type-erased, concrete interceptor.
+/// 
+/// In order to have multiple interceptors hooked into a single operation, we
+/// need a concrete type, not a protocol. This stores references to the closures
+/// of interceptor implementations and delegates to them for each interceptor hook.
+internal struct AnyInterceptor<RequestType, ResponseType, AttributesType: HasAttributes> {
```
jbelkins, line 13: Should `AnyInterceptor` conform to `Interceptor`?

jbelkins, line 13: Should `AnyInterceptor` conform to `Interceptor`?

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
```diff
@@ -0,0 +1,89 @@
...
+        self.attributes = attributes
+    }
+
+    public func getInput() -> Any {
+        self.input
+    }
+
+    public func getAttributes() -> AttributesType {
+        return self.attributes
+    }
```
jbelkins, line 30: Typical form for a Swift getter (and optionally setter) is a read-only or read-write property.
```
    public var attributes: AttributesType {
        return self._attributes
    }
```
(when the name of a public & private property conflict, preface the private property with an underscore.)

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
jbelkins: also, nit: Swift types are `internal` by default, and the `internal` keyword is usually only applied where the compiler infers otherwise (i.e. in a `public` or `private` extension.)

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
```diff
@@ -0,0 +1,172 @@
...
+    associatedtype ResponseType
+
+    /// The type of the attributes that will be available to all interceptors.
+    associatedtype AttributesType: HasAttributes
+
+    /// - Returns: The input for the operation being invoked.
+    func getInput() -> Any
+
+    /// - Returns: The attributes available in this interceptor context.
+    func getAttributes() -> AttributesType
```
jbelkins, line 24: See my comment above on Swift getter conventions

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
```diff
@@ -0,0 +1,172 @@
...
+    associatedtype RequestType
+
+    /// The type of the transport message that will be received by the operation being invoked.
+    associatedtype ResponseType
+
+    /// The type of the attributes that will be available to all interceptors.
+    associatedtype AttributesType: HasAttributes
+
+    /// - Returns: The input for the operation being invoked.
+    func getInput() -> Any
```
jbelkins, line 21: Isn't the type of the input `RequestType`?  Not sure why this return value is untyped.

Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public func modifyBeforeSerialization(context: inout some MutableInput<Self.AttributesType>) async throws {
+            var input: TestInput = context.getInput()!
+            input.otherProperty = newInputValue
+            context.updateInput(updated: input)
+        }
+
+        public func modifyBeforeTransmit(context: inout some MutableRequest<Self.RequestType, Self.AttributesType>) async throws {
+            let input: TestInput = context.getInput()!
```
jbelkins, line 82: Force unwrapping during a test will crash the entire test suite if the unwrapped value is nil.

Instead, use `XCTUnwrap` which will throw a Swift error and just fail the test being run:
```swift
let input: TestInput = try XCTUnwrap(context.getInput())
```

Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        let interceptors: [AnyInterceptor<SdkHttpRequest, HttpResponse, HttpContext>] = [
+            addAttributeInterceptor.erase(), 
+            modifyInputInterceptor.erase(), 
+            addHeaderInterceptor.erase(),
+            modifyMultipleInterceptor.erase()
+        ]
+        for i in interceptors {
+            try await i.modifyBeforeSerialization(context: &interceptorContext)
+        }
```
jbelkins, line 106: nit: loop inline with `forEach`:
```swift
try await interceptors.forEach { try await $0.modifyBeforeSerialization(context: &interceptorContext) }
```

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
```diff
@@ -0,0 +1,139 @@
...
+        self.readAfterAttempt = interceptor.readAfterAttempt(context:)
+        self.modifyBeforeCompletion = interceptor.modifyBeforeCompletion(context:)
+        self.readAfterExecution = interceptor.readAfterExecution(context:)
+    }
+
+    internal func readBeforeExecution(context: InterceptorContextType) async throws {
+        try await self.readBeforeExecution(context)
+    }
+
+    internal func modifyBeforeSerialization(context: inout InterceptorContextType) async throws {
```
jbelkins, line 68: What is the purpose for making the `context` param `inout` in this & the other interceptor hooks?

`InterceptorContextType` is a typealias for `DefaultInterceptorContext`.  `DefaultInterceptorContext` is a class, which is a Swift reference type.  

The `inout` modifier means that the reference to `context` can be modified by the called method to point to a different `context`, which isn't what I think you want.  (`inout` would be necessary if the passed context was a Swift value type such as `struct` and you wanted to change the struct's properties, but that is not the case here.)

jbelkins (changes requested):

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
milesziemer: Good question - `AnyInterceptor` can't because it must operation on `DefaultInterceptorContext` with concrete `ResponseType` etc, which can't be guaranteed by the `Interceptor` protocol.

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
milesziemer: `input` and `output` are the types of the modeled operation input and output, respectively. `RequestType` and `ResponseType` are the transport message types, like `SdkHttpRequest`, `HttpResponse`

Sources/ClientRuntime/Interceptor/AnyInterceptor.swift
milesziemer: No purpose. Working on the [orchestrator PR](https://github.com/smithy-lang/smithy-swift/pull/694) I realized it and just made `InterceptorContext` `AnyObject`, so there's no more `inout`s anywhere. Need to update this PR

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
milesziemer: I think the reason I did it this way is because the presence of certain properties on the context object differs between Interceptor hooks. For example, in `readAfterAttempt`, the service response (`response`) _may_ not be available, if we were unable to send the request, so the property is optional. Whereas in `readBeforeDeserialization`, we definitely have received a service response, so the property is non-optional. The method getters can be overloaded, so we have `getResponse() -> ResponseType?` and `getResponse() -> ResponseType` on `DefaultInterceptorContext`, but I don't know if I can do that with properties. I could add more concrete `InterceptorContext` implementations though

milesziemer: Also, I realize this comment and the one below are about `input` and `attributes`, which don't have the issue I mentioned (neither does `output`), so I could also make those regular properties and leave `getRequest` + `getResponse` as methods

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
jbelkins: Should the modeled input type be an associated type of this protocol like the RequestType & ResponseType then?  That way you can return the actual modeled input type, and you don't need the `func getInput<T>() -> T?` method below.

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    /// - Returns: The attributes available in this interceptor context.
+    func getAttributes() -> AttributesType
+}
+
+public extension InterceptorContext {
```
jbelkins, line 27: See my comment above, we should make the modeled input type an associated type of InterceptorContext so this conditional cast is unneeded

Tests/ClientRuntimeTests/InterceptorTests/InterceptorTests.swift
jbelkins: lol thought there was an async override of forEach

jbelkins (commented):
See my comment on using an associated type on interceptor context.

Also fix all the lint errors (mostly lines need to be broken up.)

To run swiftlint locally, in the root of the smithy-swift project on your Mac:
```
$ brew install swiftlint
$ swiftlint
```

Sources/ClientRuntime/Interceptor/InterceptorContext.swift
milesziemer: Done

syall (approved):

PR #686, opened 2024-03-22
feat: add checksum algorithm to runtime plugin

AndrewFossAWS:
## Description of changes
* Add checksum algorithm to runtime plugin

## Scope


Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
```diff
@@ -29,5 +29,8 @@ public protocol DefaultClientConfiguration: ClientConfiguration {
     /// A token generator to ensure idempotency of requests.
     var idempotencyTokenGenerator: IdempotencyTokenGenerator { get set }
 
-    /// TODO(plugins): Add Checksum, Traceprobes, etc.
+    /// A checksum algorithm used for requests.
+    var checksumAlgorithm: ChecksumAlgorithm? { get set }
```
syall, line 33: Just double checking, do we need to support multiple checksum algorithms on a client configuration at once? e.g. [like smithy-typescript](https://github.com/smithy-lang/smithy-typescript/blob/e03a10ac1f34ac52e0b03509c13499e459bde3bc/packages/smithy-client/src/extensions/checksum.ts#L9-L15)

If not, then LGTM

syall (approved):

Sources/ClientRuntime/Networking/ChecksumAlgorithm.swift
```diff
@@ -40,7 +40,7 @@ public enum ChecksumAlgorithm {
         return hashFunctions
     }
 
-    func toString() -> String {
+    public func toString() -> String {
```
dayaffe, line 43: Clarification: It looks like the config has ChecksumAlgorithm enum and then the specified value is converted to a string to pass into FlexibleChecksumRequestMiddleware, is that correct?

dayaffe (commented):
some small comments but overall LGTM (pending swiftlint fix)

Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
dayaffe: Question: Should the plugin use ChecksumAlgorithm enum (all possible supported checksum algorithms) or specifically the ones the particular client supports by looking in XClientTypes?

PR #694, opened 2024-04-09
feat: add operation orchestrator

milesziemer:
Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1435

### Client libraries

Adds Orchestrator which implements the operation execution pipeline.
It uses the following new "runtime components" SelectAuthScheme, ApplyEndpoint,
ApplySigner, and ExecuteRequest, in addition to the existing RetryStrategy and
Interceptor components. Orchestrator supports executing a full request, in
addition to creating a presigned request - same as OperationStack.

I say "runtime components" because these new interfaces are essentially proxies
for the actual components. For example, SelectAuthScheme is just a convenient
way of using identity and auth within Orchestrator. This makes it easier to
migrate to Interceptors & Orchestrator, because the middlewares that currently
implement these components can just conform to the new interfaces. Again take
as an example SelectAuthScheme - AuthSchemeMiddleware currently does all the
work, so we just adapt it to the SelectAuthScheme interface and we can pass it
directly to Orchestrator. In the future, we may not need these interfaces at
all because there will be no middleware. But for now they help with transition.

All existing middleware (except for RetryMiddleware) have been extended to
either implement one of the new runtime component interfaces, or Interceptor.
RetryMiddleware's implementation was just copied (roughly) into Orchestrator because
it ended up being simpler than trying to abstract it into a pseudo component.
This does have the unfortunate consequence that there are essentially two
implementations of the core retry functionality for the time being. I also
adapted SdkHttpClient to be useable as ExecuteRequest.

OrchestratorBuilder is used in generated operation method bodies to build
the operation Orchestrator with the required runtime components. This is
analagous to how an OperationStack is currently used to tie together the
middleware needed for an operation.

There are a few other new interfaces too, under Message. These are the starting
points for Requests and Responses, and will be a unifying interface for all
'transport' message types. They were added here because Orchestrator actually
needs to use some pieces of the request, like the host for example, and it also
helps with the transition from middleware. Currently, serialization is done in
multiple steps, operation on a request builder. In order to replicate this
in Orchestrator, I've added RequestMessageSerializer which is an interface that
takes an operation input, and a RequestMessageBuilder, and modifies the builder
with whatever serialized data it pulled from the input. Because Orchestrator
needs to store the serializers, we also need a type for the builder, which is
where RequestMessageBuilder comes in. These interfaces are minimal right now,
but will be expanded in the future so we can move more middleware/interceptors
to being protocol-agnostic (right now they're all http specific).
but will be expanded in the future so we can move more middleware/interceptors
to being protocol-agnostic (right now they're all http specific).

A comparison of a current generated operation body with what it will look like
using Orchestrator (I've omitted some details for readability):
```
public func getFoo(input: GetFooInput) async throws GetFooOutput {
        ...
        .build()
    var operation = OperationStack(id: "getFoo")
    operation.initializeStep.intercept(position: .after, middleware: URLPathMiddleware(...))
    operation.initializeStep.intercept(position: .after, middleware: URLHostMiddleware())
    operation.serializeStep.intercept(position: .after, middleware: QueryItemMiddleware(...))
    operation.serializeStep.intercept(position: .after, middleware: HeaderMiddleware(...))
    operation.buildStep.intercept(position: .before, middleware: EndpointResolverMiddleware(...))
    operation.buildStep.intercept(position: .before, middleware: AuthSchemeMiddleware())
    operation.finalizeStep.intercept(position: .after, middleware: RetryMiddleware(...))
    operation.finalizeStep.intercept(position: .before, middleware: SignerMiddleware())
    operation.deserializeStep.intercept(position: .after, middleware: DeserializeMiddleware(...))
    operation.deserializeStep.intercept(position: .after, middleware: LoggerMiddleware(...))
    return try await operation.handleMiddleware(context: context, input: input, next: client.getHandler())
}
```
```
public func getFoo(input: GetFooInput) async throws GetFooOutput {
        ...
        .build()
    let builder = OrchestratorBuilder()
    builder.interceptors.add(URLPathMiddleware(...))
    builder.interceptors.add(URLHostMiddleware())
    builder.serialize(QueryItemMiddleware(...))
    builder.serialize(HeaderMiddleware(...))
    builder.applyEndpoint(EndpointResolverMiddleware(...))
    builder.selectAuthScheme(AuthSchemeMiddleware())
    // These used to be passed to retry middleware
    builder.retryStrategy(...)
    builder.retryErrorInfoProvider(...)
    builder.applySigner(SignerMiddleware())
    builder.deserialize(DeserializeMiddleware(...))
    builder.interceptors.add(LoggerMiddleware(...))
    let op = builder.attributes(context) // context can be changed in the function body
        .executeRequest(client)
        .build()
    return try await op.execute(input: input)
}
```

Some other changes were made to the existing interceptors interfaces:
- Made InterceptorContext a reference type so we don't need to use `inout` everywhere
- Switch InterceptorContext 'output' -> 'result' which is a Result type containing
  the operation output. This allows us to capture error responses and other errors
  that occur during execution, so we don't immediately throw from wherever the error
  occurs.

### Codegen

Updates the codegenerator to be able to generate operations that use
Orchestrator and Interceptors instead of OperationStack. This is
configurable using the `useInterceptors` codegen setting. Because most
middleware was extended to be useable by Orchestrator, codegen only
needed to be able to render middleware a little differently based on
whether `useInterceptors` was set.

Ultimately we won't need `useInterceptors`, and we can fully refactor
the codegenerator + libraries, but for now this makes it easier and
safer to migrate.

Some test cases had to be updated to expect generic params where they
were previously inferred (they can't be inferred with interceptors,
but the middleware initializer is the same for both code paths)

## Scope


Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
```diff
@@ -0,0 +1,1251 @@
...
+            ]
+        )
+    }
+
+    func test_stepErrorHandlingReadAfterExecution() async {
+        let readAfterExecutionTrace = Trace()
+        let readAfterExecution = await asyncResult {
+            let b = self.traceOrchestrator(trace: readAfterExecutionTrace)
+            b.interceptors.addReadAfterExecution({ _ in throw TestError(value: "1") })
+            b.interceptors.addReadAfterExecution({ _ in throw TestError(value: "2") })
```
syall, line 1202: For these tests (there are a number of them), what exactly are these testing? I see `"2"` is being asserted, but not `"1"`.

syall (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSettings.kt
```diff
@@ -78,8 +81,9 @@ class SwiftSettings(
             val gitRepo = config.expectStringMember(GIT_REPO).value
             val swiftVersion = config.expectStringMember(SWIFT_VERSION).value
             val sdkId = sanitizeSdkId(config.getStringMemberOrDefault(SDK_ID, serviceId.name))
+            val useInterceptors = config.getBooleanMemberOrDefault(USE_INTERCEPTORS)
```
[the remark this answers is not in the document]
milesziemer: `false` right now, so we can control the release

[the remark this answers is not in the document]
milesziemer: That's just how `getBooleanMemberOrDefault` works

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/MiddlewareGenerator.kt
```diff
@@ -60,5 +61,7 @@ class MiddlewareGenerator(
             writer.write("public typealias MOutput = \$L", middleware.outputType)
             writer.write("public typealias Context = \$N", middleware.contextType)
         }
+
+        middleware.renderExtensions()
```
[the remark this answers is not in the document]
milesziemer: Yea, I'm pretty sure its used in the aws sdk for some of the generated middleware

Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
milesziemer: These are making sure the most recent error thrown from an interceptor is what is returned

Sources/ClientRuntime/Interceptor/Interceptors.swift
```diff
@@ -0,0 +1,448 @@
...
+        self.interceptors.append(AnyInterceptor(readAfterExecution: interceptorFn))
+    }
+}
+
+/// This extension is for the convenience of Orchestrator to call the same hook on all interceptors.
+/// Note that it doesn't conform to Interceptor because it is backed by AnyInterceptor, which also
+/// doesn't conform to Interceptor.
+///
+/// These methods throw the last error that occurred when executing a given hook, so if multiple
+/// interceptors fail, only the last error is thrown. The rest are logged (TODO)
```
dayaffe, line 181: We have a goal to remove all TODOs in our SDK, if we plan to log the additional errors let's either include that logic or create a tracking ticket for that work and pull it into sprint

Sources/ClientRuntime/Interceptor/Interceptors.swift
```diff
@@ -0,0 +1,448 @@
...
+        for i in interceptors {
+            do {
+                try await i.readBeforeExecution(context: context)
+            } catch let e {
+                error = e
+            }
+        }
+        if let error = error {
+            throw error
+        }
```
dayaffe, line 194: Lines 184-194 are repeated in each of these funcs with only one line differing, can we create a higher-order function that takes the relevant function as input? Should reduce code duplication and shorten this file by a lot

Ex. I didn't test this but maybe something like the following (also note potential suggestion on accomplishing logging)
```
internal func executeInterceptors<T>( // can make the name better
    context: InterceptorType.InterceptorContextType, 
    operation: @escaping (InterceptorType) async throws -> T //
) async throws {
    var error: Error? // suggestion: you could change this to a list 
    for i in interceptors {
        do {
            try await operation(i)
        } catch let e {
            error = e // append to list
        }
    }
   if let error = error {
        throw error // loop through list of errors and log the errors then throw the last one
   }
}
```

then would modify these functions like so:

```
internal func readBeforeExecution(context: InterceptorType.InterceptorContextType) async throws {
    try await executeInterceptors(context: context) { interceptor in
        try await interceptor.readBeforeExecution(context: context)
    }
}
```

Sources/ClientRuntime/Networking/Http/Middlewares/ContentLengthMiddleware.swift
```diff
@@ -28,37 +28,53 @@ public struct ContentLengthMiddleware<OperationStackOutput>: Middleware {
...
+        try addHeaders(builder: input, attributes: context)
+        return try await next.handle(context: context, input: input)
+    }
 
-        switch input.body {
+    private func addHeaders(builder: SdkHttpRequestBuilder, attributes: HttpContext) throws {
+        switch builder.body {
         case .data(let data):
-            input.headers.update(name: "Content-Length", value: String(data?.count ?? 0))
+            builder.headers.update(name: "Content-Length", value: String(data?.count ?? 0))
```
dayaffe, line 38: nit: could move `data?.count ?? 0` but I'll leave that up to you

```
let contentLength = data?.count ?? 0
builder.headers.update(name: "Content-Length", value: String(contentLength))
```

Sources/ClientRuntime/Networking/Http/Middlewares/RequestBody/EnumBodyMiddleware.swift
```diff
@@ -25,13 +25,22 @@ public struct EnumBodyMiddleware<OperationStackInput,
...
 }
+
+extension EnumBodyMiddleware: RequestMessageSerializer {
+    public typealias InputType = OperationStackInput
+    public typealias RequestType = SdkHttpRequest
+    public typealias AttributesType = HttpContext
+
+    public func apply(input: OperationStackInput, builder: SdkHttpRequestBuilder, attributes: HttpContext) throws {
+        let bodyString = input[keyPath: keyPath]?.rawValue ?? ""
+        builder.withBody(.data(Data(bodyString.utf8)))
```
dayaffe, line 44: nit: can use a variable for readability

```
// Convert to utf8
let bodyData = Data(bodyContent.utf8)

builder.withBody(.data(bodyData))
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/RequestTestEndpointResolverMiddleware.kt
```diff
@@ -19,15 +19,29 @@ class RequestTestEndpointResolverMiddleware(private val model: Model, private va
...
+            writer.openBlock(
+                "$operationStackName.${middlewareStep.stringValue()}.intercept(position: ${position.stringValue()}, id: \"${name}\") { (context, input, next) -> \$N<$outputShapeName> in", "}",
+                ClientRuntimeTypes.Middleware.OperationOutput
+            ) {
+                writer.write("input.withMethod(context.getMethod())")
+                writer.write("input.withPath(context.getPath())")
+                writer.write("let host = \"\\(context.getHostPrefix() ?? \"\")\\(context.getHost() ?? \"\")\"")
+                writer.write("input.withHost(host)")
+                writer.write("return try await next.handle(context: context, input: input)")
+            }
```
dayaffe, line 44: Very basic Q: Because I am unfamiliar with middlewares vs interceptors, is interceptors something that wont always be used (hence the ctx.settings.useInterceptors)? In which case, do customizations need to follow this same code logic of if interecptors then x else y for middleware?

dayaffe (changes requested):

Sources/ClientRuntime/Interceptor/Interceptors.swift
syall, line 194: This is something I didn't catch, but is there a reason why we don't fail on the first error encountered?

milesziemer: I think its a safer contract than failing on the first error encountered, in case you have interceptors dependent on something occurring in a previous interceptor, like setting up some resources. There's still issues, but at least you know your interceptor will run.

Sources/ClientRuntime/Interceptor/Interceptors.swift
milesziemer: Nice catch, forgot about this one.

Sources/ClientRuntime/Interceptor/Interceptors.swift
milesziemer: Done

Sources/ClientRuntime/Interceptor/Interceptors.swift
milesziemer: Done

dayaffe (approved):

PR #700, opened 2024-04-24
feat!: Remove Equatable protocol conformance unless necessary

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/898

## Description of changes

Removing protocol conformance where possible brings great benefits to the SDK in many ways, including compile time reduction and binary size reduction.

This PR removes codegen for `Equatable` protocol conformance for Smithy structure shapes and Smithy union shapes, **except** when they are nested under member shapes used as pagination token. Also, `Equatable` protocol conformance is left intact for Smithy enum and intEnum shapes, as equivalence operators are to be frequently used on those enum types.
- Adds a new custom trait `@equatableConformance`, used for codegen
- Modifies Smithy structure & Smithy union codegen to generate `: Equatable` only if custom trait is found on the shape
- Update existing codegen tests and remove `: Equatable` from expected outputs where needed
- Enhance existing pagination codegen test to check for `Equatable` protocol conformance where needed

#### The background on pagination token & why it's relevant
DynamoDB uses Smithy map shape for their pagination token (every other service uses a simple String, as [recommended in Smithy docs](https://smithy.io/2.0/spec/behavior-traits.html#paginated-trait:~:text=input%20member%20MUST%20NOT%20be%20marked%20as%20required%20and-,SHOULD%20target%20a%20string%20shape.%20It%20can%2C%20but%20SHOULD%20NOT%20target%20a%20map%20shape.,-When%20contained%20within%20a%20service%2C%20a%20paginated%20operation%20MUST)). There can be any number of structure shapes or union shapes nested in the map shape DynamoDB uses for pagination token, and because [AWS Swift SDK uses equivalence operations on pagination token](https://github.com/smithy-lang/smithy-swift/blob/main/Sources/ClientRuntime/Pagination/PaginatorSequence.swift#L48) to detect infinite loop, `Equatable` protocol conformance must be generated for structure shapes and union shapes IF they're nested under a pagination token.

#### The approach
- Create a new custom trait called `@equatableConformance`.
- During model pre-processing, attach this trait to all structure shapes and union shapes nested under pagination token member shapes.
- During codegen, generate `: Equatable` for structure shapes and union shapes only if the custom trait is found on the shape.

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
@@ -123,6 +123,12 @@ fun Model.getNestedShapes(serviceShape: ServiceShape): Set<Shape> {
         .select(this)
 }
 
+fun Model.getNestedShapes(memberShape: MemberShape): Set<Shape> {
+    return Selector
+        .parse("member [id=${memberShape.id}] ~> *")
+        .select(this)
+}
+
 /**
  * Test if an operation input is an event stream
  */
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
@@ -125,7 +125,7 @@ fun Model.getNestedShapes(serviceShape: ServiceShape): Set<Shape> {
 
 fun Model.getNestedShapes(memberShape: MemberShape): Set<Shape> {
     return Selector
-        .parse("member [id=${memberShape.id}] ~> *")
+        .parse("member [id='${memberShape.id}'] ~> *")
         .select(this)
 }
 
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt
@@ -112,8 +112,8 @@ class StructureGenerator(
     private fun generateStruct() {
         writer.writeShapeDocs(shape)
         writer.writeAvailableAttribute(model, shape)
-        val equatableConformance = (": " + SwiftTypes.Protocols.Equatable).takeIf { shape.hasTrait<EquatableConformanceTrait>() } ?: ""
-        writer.openBlock("public struct \$struct.name:L ${equatableConformance} {")
+        val equatableConformance = (": " + SwiftTypes.Protocols.Equatable + " ").takeIf { shape.hasTrait<EquatableConformanceTrait>() } ?: ""
+        writer.openBlock("public struct \$struct.name:L ${equatableConformance}{")
             .call { generateStructMembers() }
             .write("")
             .call { generateInitializerForStructure(false) }
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt
@@ -13,6 +13,7 @@ import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.StructureShape
 import software.amazon.smithy.model.traits.ErrorTrait
 import software.amazon.smithy.model.traits.RetryableTrait
+import software.amazon.smithy.swift.codegen.customtraits.EquatableConformanceTrait
 import software.amazon.smithy.swift.codegen.customtraits.NestedTrait
 import software.amazon.smithy.swift.codegen.customtraits.SwiftBoxTrait
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
@@ -111,7 +112,8 @@ class StructureGenerator(
     private fun generateStruct() {
         writer.writeShapeDocs(shape)
         writer.writeAvailableAttribute(model, shape)
-        writer.openBlock("public struct \$struct.name:L: \$N {", SwiftTypes.Protocols.Equatable)
+        val equatableConformance = (": " + SwiftTypes.Protocols.Equatable).takeIf { shape.hasTrait<EquatableConformanceTrait>() } ?: ""
+        writer.openBlock("public struct \$struct.name:L ${equatableConformance} {")
             .call { generateStructMembers() }
             .write("")
             .call { generateInitializerForStructure(false) }
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftCodegenPlugin.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftCodegenPlugin.kt
@@ -14,6 +14,7 @@ import software.amazon.smithy.model.transform.ModelTransformer
 import software.amazon.smithy.swift.codegen.core.GenerationContext
 import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
 import software.amazon.smithy.swift.codegen.model.AddOperationShapes
+import software.amazon.smithy.swift.codegen.model.EquatableConformanceTransformer
 import software.amazon.smithy.swift.codegen.model.NestedShapeTransformer
 import software.amazon.smithy.swift.codegen.model.RecursiveShapeBoxer
 import software.amazon.smithy.swift.codegen.model.UnionIndirectivizer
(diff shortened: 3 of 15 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/ShapeExt.kt
```diff
@@ -123,6 +123,12 @@ fun Model.getNestedShapes(serviceShape: ServiceShape): Set<Shape> {
         .select(this)
 }
 
+fun Model.getNestedShapes(memberShape: MemberShape): Set<Shape> {
+    return Selector
+        .parse("member [id='${memberShape.id}'] ~> *")
+        .select(this)
+}
```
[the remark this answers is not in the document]
sichanyoo: Oh, that's a good callout. I'll update it to make it more concise.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestResponseGenerator.kt
```diff
@@ -139,44 +138,97 @@ open class HttpProtocolUnitTestResponseGenerator protected constructor(builder:
...
+                        is StructureShape -> {
+                            shape.members().filter { !it.hasTrait<HttpQueryTrait>() }.forEach { member ->
+                                val propertyName = ctx.symbolProvider.toMemberName(member)
+                                val path = "properties.".takeIf { shape.hasTrait<ErrorTrait>() } ?: ""
+                                val propertyAccessor = "$path$propertyName"
+                                val target = ctx.model.expectShape(member.target)
+                                when (target) {
+                                    is FloatShape, is DoubleShape -> {
+                                        writer.write(
+                                            "if (lhs.\$L?.isNaN ?? false && rhs.\$L?.isNaN ?? false) { return true }",
```
jbelkins, line 196: Just realized that if you `return true` after this check, it will short-circuit whatever checks remain on the structure's fields.  So either put it back the way it was, or find another way to get the required behavior.

jbelkins (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestResponseGenerator.kt
sichanyoo: Fixed via helper function.

Sources/ClientRuntime/Networking/Streaming/ByteStream.swift
```diff
@@ -34,7 +34,9 @@ extension ByteStream: Equatable {
         case (.data(let lhsData), .data(let rhsData)):
             return lhsData == rhsData
         case (.stream(let lhsStream), .stream(let rhsStream)):
-            return lhsStream === rhsStream
+            // swiftlint:disable force_try
+            return try! lhsStream.readToEnd() == rhsStream.readToEnd()
+            // swiftlint:enable force_try
```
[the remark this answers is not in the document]
sichanyoo: Moved and took care of error that came up after offline discussion.

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
```diff
@@ -0,0 +1,24 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public func floatingPointValuesMatch(
+    lhs: (any FloatingPoint)?,
+    rhs: (any FloatingPoint)?
+) -> Bool {
```
jbelkins, line 11: Rather than using `any`, I would make the signature generic.
```
public func floatingPointValuesMatch<FP: FloatingPoint>(lhs: FP?, rhs: FP?) -> Bool
```
This does two things:
1) It allows the compiler to resolve the types at compile time rather than run time.  This allows for better performance through generic specialization.
2) It requires that `lhs` and `rhs` are of the same type, which simplifies your logic within.  (Since you're using this to compare the same member in the same struct, that is always a valid condition.

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
```diff
@@ -0,0 +1,24 @@
...
+    // Only one of them is nil => not equal
+    if (lhs == nil || rhs == nil) { return false }
+    // Both are NaN => equal
+    if (lhs!.isNaN && rhs!.isNaN) { return true }
+    // Only one of them is NaN => not equal
+    if (lhs!.isNaN || rhs!.isNaN) { return false }
+    // Both are non-nil & non-NaN => equal IFF values equal
+    let doubleMatch = (lhs as? Double ?? .nan) == (rhs as? Double ?? .nan)
+    let floatMatch = (lhs as? Float ?? .nan) == (rhs as? Float ?? .nan)
+    return doubleMatch || floatMatch
```
jbelkins, line 23: Logic should break down to:
```
if let lhs, let rhs, lhs.isNaN, rhs.isNaN { return true }
return lhs == rhs
```

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
```diff
@@ -0,0 +1,24 @@
...
+public func floatingPointValuesMatch(
+    lhs: (any FloatingPoint)?,
+    rhs: (any FloatingPoint)?
+) -> Bool {
+    // Both are nil => equal
+    if (lhs == nil && rhs == nil) { return true }
+    // Only one of them is nil => not equal
+    if (lhs == nil || rhs == nil) { return false }
+    // Both are NaN => equal
+    if (lhs!.isNaN && rhs!.isNaN) { return true }
```
jbelkins, line 17: I know you already verified that both lhs & rhs are non-nil, but still beware of force-unwrap.

If someone comes along & refactors or borrows this code later & isn't careful, the `!` could bite them.

jbelkins (changes requested):
See comments.  Some simplification is definitely possible.  Since you have a very thorough set of tests for your helper, you can aggressively refactor with a lot of confidence.

I also pulled this branch down & verified this helper method passes the tests:
```swift
public func floatingPointValuesMatch<FP: FloatingPoint>(
    lhs: FP?,
    rhs: FP?
) -> Bool {
    if let lhs, let rhs, lhs.isNaN, rhs.isNaN { return true }
    return lhs == rhs
}
```

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
jbelkins: I almost forgot the most important part: with `lhs` and `rhs` constrained to be the same type, you can directly use Equatable to compare the two values rather than casting as you did below.

This enables the very simplified logic I posted below.

jbelkins: (Equatable contains `Self` constraints, so you can only equate two values of the same type, even if they are both `FloatingPoint`s.)

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
jbelkins: (presuming you adopt the generic method sig above)

Tests/SmithyTestUtilTests/ResponseTestUtilTests/FloatingPointEquivalenceCheckerTests.swift
```diff
@@ -0,0 +1,87 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import SmithyTestUtil
+import XCTest
+
+class FloatingPointEquivalenceCheckerTests : XCTestCase {
```
jbelkins, line 11: These tests are great!  They look like they cover all our cases pretty exhaustively.

With the generic param types, doing both Float & Double is probably overkill, but might as well keep both.

Sources/SmithyTestUtil/ResponseTestUtil/FloatingPointEquivalenceChecker.swift
sichanyoo: Changed as suggested 👍

jbelkins (approved):

PR #703, opened 2024-04-28
feat!: Remove model namespace

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1458

## Description of changes
Removes the namespace around nested model types to reduce boilerplate and simplify service clients.  Results in the following changes to a service's models:
- Input and output structures are renamed to end in `OperationInput` and `OperationOutput` (to resolve name collision with a couple services that have names that clash with the current `Input` and `Output` names.)
- Model types that are not top-level (i.e. in a property or sub-property of an input, output, or error) are no longer namespaced to `ServiceClientTypes` and are created in the root namespace of the service.
- Error structures are unchanged.

Specific changes made to support:
- The `NestedTrait` and the transformer that applies it are removed, since no models are nested anymore.
- Removes code that renders models and their extensions in a namespace.
- Fixes many references to runtime types so that they use the fully qualified name, to avoid unresolvable ambiguity with modeled types of the same name.
- Changes the name of input / output type name suffix from `Input` and `Output` to `OperationInput` and `OperationOutput`.  This was needed to avoid name collisions after renaming.

_Note: Many codegen tests are broken due to changed names.  Will fix once we settle on naming structure._

## Scope


jbelkins:
Will have to investigate this further before deciding on an approach.  Closing this PR but preserving the branch for potential future use.

PR #704, opened 2024-04-30
feat!: add interceptor client config

milesziemer:
## Description of changes

Allows configuration of interceptors on a client level by adding interceptor
providers to client config, allowing Plugins to add interceptors.

The primary addition is `InterceptorProvider`, an interface that creates
generic interceptors which can operate on any transport - http or otherwise.
When an operation is executed, interceptor providers are called to create
new instances of service-level interceptors. Creating new instances also
means we don't need to synchronize on the shared interceptors. Transport
specific config, have their own methods for adding interceptor providers, so
you can add `HttpInterceptorProvider`s to http config. Operations know which
transport they operate on, and can choose which transport-specific interceptor
providers to use.

If/when we have operation-level configuration, it might make more sense to
allow plugins to configure more generic 'operation customizations' or
something, rather than just the interceptors. Operations would then call
the customizations before executing.

A few other minor changes were made to the client libraries:
- Added actual builder methods to RequestMessageBuilder, which we would need
eventually, so I could use them in testing
- Made SdkHttpRequestBuilder final

Codegen was also updated to generate the new config methods, and to call
interceptor providers in operations to add configured interceptors.

## Scope

- [x] Technically a breaking change because it adds methods to config protocols, but it is a new feature


Sources/ClientRuntime/Message/RequestMessageBuilder.swift
```diff
@@ -13,6 +13,10 @@ public protocol RequestMessageBuilder<RequestType>: AnyObject {
 
     init()
 
+    func withHost(_ host: String) -> Self
```
AndrewFossAWS, line 16: `Host` should be provided from `URI`?

milesziemer: Oh yea, this is just here so I could use it in a test. I think this can changed to something like `withURI`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -105,7 +105,20 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
             if (!ctx.settings.useInterceptors) {
                 writer.write("var $operationStack = OperationStack<$inputSymbol, $outputSymbol>(id: \"${test.id}\")")
             } else {
+                val serviceSymbol = ctx.symbolProvider.toSymbol(ctx.service)
+                writer.write("let config = try await \$L.builder().build().config", serviceSymbol.name)
                 writer.write("let builder = OrchestratorBuilder<$inputSymbol, $outputSymbol, SdkHttpRequest, HttpResponse, HttpContext>()")
+                writer.write("config.interceptorProviders.forEach { builder.interceptors.add($$0.create()) }")
```
dayaffe, line 111: nit: more readable if you did 
```
config.interceptorProviders.forEach { provider in
     builder.interceptors.add(provider.create())
 }
```

dayaffe (approved):
Just one nit

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/middleware/MiddlewareExecutionGenerator.kt
```diff
@@ -53,12 +53,25 @@ class MiddlewareExecutionGenerator(
...
+                ClientRuntimeTypes.Http.HttpContext,
+            )
+            writer.write("config.interceptorProviders.forEach { builder.interceptors.add($$0.create()) }")
+            // Swift can't infer the generic arguments to `create` for some reason
+            writer.write(
+                """
+                config.httpInterceptorProviders.forEach {
+                    let i: any HttpInterceptor<${'$'}N, ${'$'}N> = $$0.create()
+                    builder.interceptors.add(i)
+                }
```
dayaffe, line 71: same nit here might be better to use a name instead of `$0.create` for readability

Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
```diff
@@ -31,5 +31,10 @@ public protocol DefaultClientConfiguration: ClientConfiguration {
     /// If none is provided, only a default logger provider will be used.
     var telemetryProvider: TelemetryProvider { get set }
 
+    /// Add an `InterceptorProvider` that will be used to provide interceptors for all operations.
```
dayaffe, line 34: Q: Interceptor providers apply to Http operations + all other operations vs Http Interceptor providers apply to only http operations?

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
```diff
@@ -85,8 +93,8 @@ extension DefaultInterceptorContext: AfterAttempt {
 }
 
 extension DefaultInterceptorContext: MutableOutputAfterAttempt {
-    public func updateResult(updated: Result<OutputType, Error>) {
-        self.result = updated
+    public func updateOutput(updated: OutputType) {
+        self.result = .success(updated)
```
dayaffe, line 97: before updated could be an `Error`, will this always be a success?

Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
milesziemer: That is correct

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
```diff
@@ -35,9 +34,13 @@ public class DefaultInterceptorContext<
         self.input
     }
 
-    public func getAttributes() -> AttributesType {
+    public func getAttributes() -> Context {
         return self.attributes
     }
+
```
dayaffe, line 40: please take care of the lint issues

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
milesziemer: The orchestrator uses [setResult](https://github.com/smithy-lang/smithy-swift/pull/704/files/157b12fcffaacb0145c94d499554e626f9f291f9#diff-8184a6ba179295649afb02c7f9473e695195634dfcc8b1ab425ca8d4c916b3c1R41) to set errors. So internally DefaultInterceptorContext keeps track of failures - if you look at `getOutput`, you'll see it now throws.

This change was based on discussion with @jbelkins about mixing and matching `Result` with `throws` in Swift.

Sources/ClientRuntime/Interceptor/DefaultInterceptorContext.swift
milesziemer: Maybe we should make these lint issues errors instead of warnings so they fail local `swiftlint` runs and CI

dayaffe (approved):

PR #707, opened 2024-05-01
fix: Move FoundationStreamBridge from Thread to GCD for stream callbacks

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1467

## Description of changes

I discovered the `CFWriteStreamSetDispatchQueue` which configures a Foundation stream to deliver callbacks on a GCD dispatch queue instead of on a RunLoop.  This PR makes the switch, and eliminates the Thread and Runloop previously used for interacting with the stream.

Apple docs for `perform(_:on:with:waitUntilDone:modes:)` warn of incorrect behavior when performing on a different thread from a GCD queue:
https://developer.apple.com/documentation/objectivec/nsobject/1417922-perform#

Moving stream interactions to GCD should avoid this, and should also interact better with Swift concurrency.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
+++ b/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
@@ -5,6 +5,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#if os(iOS) || os(macOS) || os(watchOS) || os(tvOS) || os(visionOS)
+
 import func Foundation.CFWriteStreamSetDispatchQueue
 import class Foundation.DispatchQueue
 import class Foundation.NSObject
@@ -204,3 +206,5 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
         }
     }
 }
+
+#endif
--- a/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
+++ b/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
@@ -7,8 +7,8 @@
 
 #if os(iOS) || os(macOS) || os(watchOS) || os(tvOS) || os(visionOS)
 
+import func Foundation.CFWriteStreamSetDispatchQueue
 import class Foundation.DispatchQueue
-import func Foundation.autoreleasepool
 import class Foundation.NSObject
 import class Foundation.Stream
 import class Foundation.InputStream
@@ -73,25 +73,9 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
     /// Actor used to enforce the order of multiple concurrent stream writes.
     private let writeCoordinator = WriteCoordinator()
 
-    /// A shared serial DispatchQueue to run the `perform`-on-thread operations.
-    /// Performing thread operations on an async queue allows Swift concurrency tasks to not block.
-    private static let queue = DispatchQueue(label: "AWSFoundationStreamBridge")
-
-    /// Foundation Streams require a run loop on which to post callbacks for their delegates.
-    /// All stream operations should be performed on the same thread as the delegate callbacks.
-    /// A single shared `Thread` is started and is used to host the RunLoop for all Foundation Stream callbacks.
-    private static let thread: Thread = {
-        let thread = Thread {
-            autoreleasepool {
-                let timer = Timer(timeInterval: TimeInterval.greatestFiniteMagnitude, repeats: true, block: { _ in })
-                RunLoop.current.add(timer, forMode: .default)
-                RunLoop.current.run(until: Date.distantFuture)
-            }
-        }
-        thread.name = "AWSFoundationStreamBridge"
-        thread.start()
-        return thread
-    }()
+    /// A shared serial DispatchQueue to run the stream operations.
+    /// Performing operations on an async queue allows Swift concurrency tasks to not block.
+    private let queue = DispatchQueue(label: "AWSFoundationStreamBridge")
 
     // MARK: - init & deinit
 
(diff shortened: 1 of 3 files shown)


Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -73,25 +73,9 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
-                RunLoop.current.run(until: Date.distantFuture)
-            }
-        }
-        thread.name = "AWSFoundationStreamBridge"
-        thread.start()
-        return thread
-    }()
+    /// A shared serial DispatchQueue to run the stream operations.
+    /// Performing operations on an async queue allows Swift concurrency tasks to not block.
+    private let queue = DispatchQueue(label: "AWSFoundationStreamBridge")
```
jbelkins, line 78: This queue is now a member of a single `FoundationStreamBridge` instead of a shared, static queue.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -123,47 +107,38 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
         }
     }
 
-    /// Configure the output stream to make StreamDelegate callback to this bridge using the special thread / run loop, and open the output stream.
-    /// The input stream is not included here.  It will be configured by `URLSession` when the HTTP request is initiated.
-    @objc private func openOnThread() {
-        outputStream.delegate = self
-        outputStream.schedule(in: RunLoop.current, forMode: .default)
-        outputStream.open()
-    }
```
jbelkins, line 147: The `openOnThread()` method was collapsed into the `open()` method above, since I no longer need to separate the specific steps that are performed on the thread.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -123,47 +107,38 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
-            continuation.resume()
         }
     }
 
-    /// Close the output stream and remove it from the thread / run loop.
-    @objc private func closeOnThread() {
-        outputStream.close()
-        outputStream.remove(from: RunLoop.current, forMode: .default)
-        outputStream.delegate = nil
-    }
```
jbelkins, line 165: Like with `openOnThread()` & `open()`, the `closeOnThread()` method was collapsed into the `close()` method above.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -204,25 +181,9 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
-        buffer.withUnsafeBytes { bufferPtr in
-            let bytePtr = bufferPtr.bindMemory(to: UInt8.self).baseAddress!
-            writeCount = outputStream.write(bytePtr, maxLength: buffer.count)
-        }
-        if writeCount > 0 {
-            logger.info("FoundationStreamBridge: wrote \(writeCount) bytes to request body")
-            buffer.removeFirst(writeCount)
-        }
-        result.error = outputStream.streamError
-    }
```
jbelkins, line 221: Same with `writeToOutputStreamOnThread()` & `writeToOutputStream()` as with open & close.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
jbelkins: The static thread is deleted because stream callbacks will now be delivered to `queue`, and the thread is unused.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -182,20 +157,22 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
     private func writeToOutputStream(data: Data) async throws {
         try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
-            Self.queue.async {
-                let result = WriteToOutputStreamResult()
-                result.data = data
-                let selector = #selector(self.writeToOutputStreamOnThread)
-                self.perform(selector, on: Self.thread, with: result, waitUntilDone: true)
-                if let error = result.error {
+            queue.async { [self] in
+                guard !buffer.isEmpty || !data.isEmpty else { continuation.resume(); return }
```
dayaffe, line 164: When I looking at this code during checksums, I was wondering should we check if `outputStream.streamStatus == .open` ?

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -182,20 +157,22 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
+                buffer.append(data)
+                var writeCount = 0
+                buffer.withUnsafeBytes { bufferPtr in
+                    let bytePtr = bufferPtr.bindMemory(to: UInt8.self).baseAddress!
```
dayaffe, line 168: Could use `if let bytePtr = bufferPtr.bindMemory(to: UInt8.self).baseAddress` to remove bang `!`

dayaffe (commented):
Looks great! Just one part maybe we could clean up a little. I agree that GCD should help a lot

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -182,20 +157,22 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
+                    writeCount = outputStream.write(bytePtr, maxLength: buffer.count)
+                }
+                if writeCount > 0 {
+                    logger.info("FoundationStreamBridge: wrote \(writeCount) bytes to request body")
+                    buffer.removeFirst(writeCount)
+                }
+                if let error = outputStream.streamError {
```
dayaffe, line 175: Could move if statement on 171 to inside of the scope of withUnsafeBytes and delete line 166 that initializes writeCount. line 169 should set writeCount to 0 if it doesnt write anything. Overall this could be changed to:
```
buffer.withUnsafeBytes { bufferPtr in
        if let bytePtr = bufferPtr.bindMemory(to: UInt8.self).baseAddress {
            let writeCount = outputStream.write(bytePtr, maxLength: buffer.count)
            if writeCount > 0 {
                logger.info("FoundationStreamBridge: wrote \(writeCount) bytes to request body")
                buffer.removeFirst(writeCount)
            }
        }
    }
```

jbelkins: I can't restructure the `if` into the buffer access because I can't modify `buffer` inside the block where I access its contents.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
jbelkins: The stream will error below if we write to it while not open, I think that is sufficient.  

(This method is only called as a result of a stream delegate callback, and that callback only happens when the stream is open, and the `writeToOutput()` method above enforces concurrent access with a check that the stream has not already closed, so as a practical matter, a write will never happen when the stream is closed.)

dayaffe (approved):

dayaffe:
LGTM pending CI passes

PR #710, opened 2024-05-03
feat!: Create retry modules

jbelkins:
## Scope


.swiftlint.yml
```diff
@@ -2,11 +2,7 @@ excluded:
   - .build
   - Sources/SmithyTestUtil/*
   - Sources/WeatherSDK/*
-  - Tests/ClientRuntimeTests/*
-  - Tests/SmithyTestUtilTests/*
-  - Tests/SmithyXMLTests/*
-  - Tests/SmithyTimestampsTests/*
-  - Tests/WeatherSDKTests/*
+  - Tests/*
```
jbelkins, line 5: No need to list these test exceptions to lint rules 1 by 1

Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
```diff
@@ -5,6 +5,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import struct SmithyRetriesAPI.RetryStrategyOptions
+
```
jbelkins, line 9: Separate imports have to be added in many files for the retry types extracted into modules.
I won't comment in these files every time those are added.

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -119,7 +124,9 @@ public extension DefaultSDKRuntimeConfiguration {
     /// The retry strategy options to use when none is provided.
     ///
     /// Defaults to options with the defaults defined in `RetryStrategyOptions`.
-    static var defaultRetryStrategyOptions: RetryStrategyOptions { RetryStrategyOptions() }
+    static var defaultRetryStrategyOptions: RetryStrategyOptions {
+        RetryStrategyOptions(backoffStrategy: ExponentialBackoffStrategy())
+    }
```
jbelkins, line 129: Since `RetryStrategyOptions` (in SmithyRetriesAPI) has been split from `ExponentialBackoffStrategy` (an implementation in SmithyRetries) the backoff strategy must be provided to the retry strategy options when it is created by the caller.

Sources/SmithyRetries/DefaultRetryStrategy/DefaultRetryStrategy.swift
```diff
@@ -55,8 +59,3 @@ public struct DefaultRetryStrategy: RetryStrategy {
         await token.quota.retryQuotaRelease(isSuccess: true, capacityAmount: token.capacityAmount)
     }
 }
-
-enum RetryError: Error {
-    case maxAttemptsReached
-    case insufficientQuota
-}
```
jbelkins, line 62: This error was made public and moved to its own file in `SmithyRetriesAPI`.

Sources/SmithyRetries/RetryErrorType+CRT.swift
```diff
@@ -0,0 +1,21 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import enum SmithyRetriesAPI.RetryErrorType
+import AwsCommonRuntimeKit
+
+public extension RetryErrorType {
```
jbelkins, line 11: This extension was split from the declaration of `RetryErrorType`, which was placed in SmithyRetriesAPI.

This extension goes into the implementation module SmithyRetries since it is dependent on CRT.

Sources/SmithyRetriesAPI/RetryError.swift
```diff
@@ -0,0 +1,11 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public enum RetryError: Error {
+    case maxAttemptsReached
+    case insufficientQuota
+}
```
jbelkins, line 11: This type was originally declared along with the `DefaultRetryStrategy` and was internal.

Now, it's public and in the SmithyRetriesAPI module.

Sources/SmithyRetriesAPI/RetryErrorType.swift
```diff
@@ -22,15 +21,3 @@ public enum RetryErrorType: Equatable {
...
-
-    func toCRTType() -> AwsCommonRuntimeKit.RetryError {
-        switch self {
-        case .transient: return .transient
-        case .throttling: return .throttling
-        case .serverError: return .serverError
-        case .clientError: return .clientError
-        }
-    }
-}
```
jbelkins, line 36: This CRT-dependent extension was split off to its own file in SmithyRetries (see above.)

Sources/SmithyRetriesAPI/RetryStrategyOptions.swift
```diff
@@ -50,7 +50,7 @@ public struct RetryStrategyOptions {
     ///   - availableCapacity: The number of available tokens in a retry quota.  Defaults to 500.
     ///   - maxCapacity: The max number of tokens in a retry quota.  Defaults to 500.
     public init(
-        backoffStrategy: RetryBackoffStrategy = ExponentialBackoffStrategy(),
+        backoffStrategy: RetryBackoffStrategy,
```
jbelkins, line 53: `ExponentialBackoffStrategy` was moved off to the implementation module, so this default value can't be set anymore - backoff strategy must be passed in when `RetryStrategyOptions` is created.

Sources/WeatherSDK/models/Baz.swift
```diff
@@ -3,7 +3,7 @@
 import ClientRuntime
 
 extension WeatherClientTypes {
-    public struct Baz: Swift.Equatable {
+    public struct Baz {
```
jbelkins, line 6: Not commenting on individual `WeatherSDK` files.

Some of the changes seen in `WeatherSDK` generated code are from features prior to this PR.

Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
```diff
@@ -8,6 +8,8 @@
 import XCTest
 
 @testable import ClientRuntime
+import SmithyRetriesAPI
+import SmithyRetries
 
 class OrchestratorTests: XCTestCase {
```
jbelkins, line 14: Just a few changes related to the relocation of retry types.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
```diff
@@ -120,8 +120,6 @@ object ClientRuntimeTypes {
         val SDKLogLevel = runtimeSymbol("SDKLogLevel")
         val ClientLogMode = runtimeSymbol("ClientLogMode")
         val IdempotencyTokenGenerator = runtimeSymbol("IdempotencyTokenGenerator")
-        val DefaultRetryStrategy = runtimeSymbol("DefaultRetryStrategy")
-        val RetryStrategyOptions = runtimeSymbol("RetryStrategyOptions")
```
jbelkins, line 124: Symbols moved to a separate file for retry module types

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftDependency.kt
```diff
@@ -26,6 +26,22 @@ enum class SwiftDependency(
...
+        "smithy-swift"
+    ),
+    SMITHY_RETRIES(
+        "SmithyRetries",
+        "main",
+        "0.1.0",
+        "https://github.com/smithy-lang/smithy-swift",
+        Resources.computeAbsolutePath("smithy-swift", "", "SMITHY_SWIFT_CI_DIR"),
+        "smithy-swift"
+    ),
```
jbelkins, line 44: Our two new modules defined as Swift dependencies

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolServiceClient.kt
```diff
@@ -124,6 +124,7 @@ open class HttpProtocolServiceClient(
             val properties: List<ConfigProperty> = ctx.integrations
                 .flatMap { it.clientConfigurations(ctx).flatMap { it.getProperties(ctx) } }
                 .let { overrideConfigProperties(it) }
+            properties.forEach { writer.addImport(it.type) }
```
jbelkins, line 127: Making sure imports are added as needed

smithy-swift-codegen/src/test/kotlin/ContentMd5MiddlewareTests.kt
```diff
@@ -7,35 +7,7 @@ class ContentMd5MiddlewareTests {
         val context = setupTests("Isolated/contentmd5checksum.smithy", "aws.protocoltests.restxml#RestXml")
         val contents = getFileContents(context.manifest, "/RestXml/RestXmlProtocolClient.swift")
         val expectedContents = """
-    public func idempotencyTokenWithStructure(input: IdempotencyTokenWithStructureInput) async throws -> IdempotencyTokenWithStructureOutput {
```
jbelkins, line 10: This test is ridiculously overbroad for "generates ContentMD5 middleware" so I cut out the expected Swift code unrelated to that test assertion.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/RetryMiddleware.kt
```diff
@@ -10,12 +10,14 @@ import software.amazon.smithy.codegen.core.SymbolProvider
...
+import software.amazon.smithy.swift.codegen.SwiftDependency
 import software.amazon.smithy.swift.codegen.SwiftWriter
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
 import software.amazon.smithy.swift.codegen.integration.middlewares.handlers.MiddlewareShapeUtils
 import software.amazon.smithy.swift.codegen.middleware.MiddlewarePosition
 import software.amazon.smithy.swift.codegen.middleware.MiddlewareRenderable
 import software.amazon.smithy.swift.codegen.middleware.MiddlewareStep
+import software.amazon.smithy.swift.codegen.swiftmodules.SmithyRetriesTypes
 
 class RetryMiddleware(
```
jbelkins, line 22: Adjust imports and a little code cleanup.

smithy-swift-codegen/src/test/kotlin/IdempotencyTokenTraitTests.kt
```diff
@@ -6,34 +6,7 @@ class IdempotencyTokenTraitTests {
         val context = setupTests("Isolated/idempotencyToken.smithy", "aws.protocoltests.restxml#RestXml")
         val contents = getFileContents(context.manifest, "/RestXml/RestXmlProtocolClient.swift")
         val expectedContents = """
-    public func idempotencyTokenWithStructure(input: IdempotencyTokenWithStructureInput) async throws -> IdempotencyTokenWithStructureOutput {
```
jbelkins, line 9: Like the test above, the expectation is way way broader than "generates idempotent middleware".

Sources/SmithyRetriesAPI/RetryError.swift
milesziemer: Is this actually used by/part of SmithyRetriesAPI? I don't see mention of the 'quota' concept anywhere in API, but it is here (unless I've missed something). I also don't see it used anywhere, presumably its meant to be the type of error that could be thrown by `refreshTokenForRetry`?

milesziemer: Also, might be a good time to add some more swift docs on what each of these types are, what they do, etc. (in general, not just for `RetryError`). Like RetryErrorInfo, RetryErrorType, RetryStrategy.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
```diff
@@ -120,8 +120,6 @@ object ClientRuntimeTypes {
...
         val DefaultRetryErrorInfoProvider = runtimeSymbol("DefaultRetryErrorInfoProvider")
```
milesziemer, line 123: Does this need to be moved too? I don't see `DefaultRetryErrorInfoProvider` in the new `SmithyRetries` module actually, is that intentional?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
milesziemer, line 1: Probably would add a lot of noise to this PR, but in the future maybe we consider moving ClientRuntimeTypes into swiftmodules

milesziemer (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
jbelkins: @milesziemer `DefaultRetryErrorInfoProvider` stayed in `ClientRuntime` because it has dependencies on things like service errors, HTTP requests, etc.

If I moved it, `SmithyRetries` would have to depend on `ClientRuntime`, and that would create a circular dependency (a bad practice in any language but a compile error in Swift.)  Perhaps the error info provider can be moved out later when its dependencies themselves are modularized.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ClientRuntimeTypes.kt
jbelkins: True.  I have many more of these modularization PRs to tackle, I can do it in a future one.

jbelkins:
@milesziemer 
> Is this actually used by/part of SmithyRetriesAPI? I don't see mention of the 'quota' concept anywhere in API, but it is here (unless I've missed something). I also don't see it used anywhere, presumably its meant to be the type of error that could be thrown by refreshTokenForRetry?

Agreed.  `.maxAttemptsReached` is implied by the retry configuration in `SmithyRetriesAPI` so it should stay here, but the `. insufficientQuota` error is really tied to the "adaptive" implementation.  I'll move that error back in with the default implementation.

Also no matter where these go, if these errors are public they should have doc comments.  I'll add them.

jbelkins:
@milesziemer Per discussion above:
- The general-case max retries error was split from the adaptive-related quota error, and the quota error was relocated to `SmithyRetries` with the retry strategy implementation
- Doc comments were added to both errors

Please re-review.

Sources/SmithyRetriesAPI/RetryErrorType.swift
milesziemer, line 1: Still a little confused on `RetryError` vs `RetryErrorType` - I see `RetryError` is for when you can no longer perform a retry, and `RetryErrorType` is more like what kind of error (caused?) the retry? Maybe a naming change would be helpful, or just docs on `RetryErrorType`.

milesziemer (changes requested):

Sources/SmithyRetriesAPI/RetryErrorType.swift
jbelkins: @milesziemer I added doc comments to `RetryErrorType` and `RetryErrorInfo` (of which, `RetryErrorType` is one of its fields) to clarify how these two types are used in the retry flow.

I don't want to rename `RetryError` because that is the best name for it, and `RetryErrorInfo` is named straight out of reference architecture.  But I think the comments will be enough to distinguish them from one another.

PR #711, opened 2024-05-03
fix: Fix CRT HTTP client continuation bug

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1474

## Description of changes

Uses a class that wraps the continuation to ensure the `StreamContinuation` that gets passed around in `CRTClientEngine.swift` gets resumed exactly once.

*Confirmed with Waqar that the callback functions do not get called concurrently in the underlying aws-crt-swift client engine, hence no need for special handling for the flag to make it concurrency safe.

## Scope


Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -361,4 +419,33 @@ public class CRTClientEngine: HTTPClient {
         response.body = .stream(stream)
         return requestOptions
     }
+
+    actor ContinuationFlag {
+        var continuationResumed = false
+        public func setFlag(val: Bool) {
```
dayaffe, line 425: nit: rename to something more descriptive like `setContinuationResumed` and since `set` is the name we can remove val: when using the function

ex. `public func setContinuationResumed(_ val: Bool)`
usage: `public func setContinuationResume(true)`

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -361,4 +419,33 @@ public class CRTClientEngine: HTTPClient {
         response.body = .stream(stream)
         return requestOptions
     }
+
+    actor ContinuationFlag {
+        var continuationResumed = false
```
dayaffe, line 424: suggestion: can make continuationResume `private` and add a getter to check if it has not been resumed like `setContinuationNotResumed`, that way access to continuationResumed is more explicitly controlled

dayaffe (changes requested):
Two small suggestions! Great work on fixing this

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
sichanyoo: Changed as suggested.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
sichanyoo: Changed as suggested.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -361,4 +433,46 @@ public class CRTClientEngine: HTTPClient {
...
+    }
+
+    private func safeResumeThrowContinuation (
+        continuationFlagSyncQueue: DispatchQueue,
+        continuationFlag: ContinuationFlag,
+        continuation: StreamContinuation,
+        error: Error
+    ) async {
+        continuationFlagSyncQueue.async {
+            Task {
```
jbelkins, line 454: Getting onto a DispatchQueue just to start a task has no effect whatsoever.  The async queue task will return immediately after the Task is created & before it finishes.

Might as well start the Task from the calling thread without the dispatch queue, the effect will be identical.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -217,57 +224,82 @@ public class CRTClientEngine: HTTPClient {
...
-                // it is now safe to write the body
-                // writing is done in a separate task to avoid blocking the continuation
-                Task { [logger] in
+                let requestOptions = makeHttpRequestStreamOptions(
+                    continuationFlag: continuationFlag,
+                    request: crtRequest,
+                    continuation: continuation,
+                    http2ManualDataWrites: true
+                )
+                Task {
```
jbelkins, line 255: What is the purpose of this Task?  It contains some code that doesn't appear to be async, then creates yet another async task from inside it.

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -80,7 +80,7 @@ public extension DefaultSDKRuntimeConfiguration {
     static func makeClient(
         httpClientConfiguration: HttpClientConfiguration = defaultHttpClientConfiguration
     ) -> HTTPClient {
-        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS) || os(macOS)
+        #if os(iOS) || os(tvOS) || os(watchOS) || os(visionOS)
```
jbelkins, line 83: Let's split this out to a separate PR since it is unrelated to the fix below and these are both significant changes.

jbelkins (changes requested):

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
sichanyoo: I must've forgot to clean it up; deleted as pointed out 👍

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
sichanyoo: Ok 👍

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -361,4 +410,36 @@ public class CRTClientEngine: HTTPClient {
         response.body = .stream(stream)
         return requestOptions
     }
+
+    class ContinuationFlag {
+        private var continuationResumed = false
```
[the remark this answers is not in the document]
sichanyoo: That's a good point; changed as suggested.

jbelkins (changes requested):
Just small stuff this time

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -267,36 +276,37 @@ public class CRTClientEngine: HTTPClient {
...
+                } catch {
+                    logger.error(error.localizedDescription)
                 }
             }
         }
     }
 
     /// Creates a `HTTPRequestOptions` object that can be used to make a HTTP request
     /// - Parameters:
+    ///   - continuationFlag: The actor that wraps the boolean flag used to track continuation resume status
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested

jbelkins (approved):

dayaffe (approved):

PR #714, opened 2024-05-07
chore: add codegen CI generated code check

syall:
## Issue \#

The Weather SDK codegen is out of sync, and there is no mechanism to check it.

## Description of changes

Add codegen CI generated code check.

## Scope

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -32,3 +32,8 @@ jobs:
 
       - name: Clean and Build
         run: ./gradlew clean build -Plog-tests
+
+      - name: Ensure there are no changes after building (e.g. the Weather SDK)
+        run: |
+          git diff
+          test -z "$(git diff)"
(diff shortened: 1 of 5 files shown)


AndrewFossAWS (approved):

jbelkins (commented):
Not sure I see the value of this.

Three additional jobs that run in series with our existing CI jobs and duplicate much of the setup we already perform doesn't solve any problem we have.  This will just make CI take longer than it already does, which IMO is too long.

If we want to check the diff is clean for generated code, we can do that as an added step in our existing CI steps and it will add very little complexity and very little build time.

syall:
> Not sure I see the value of this.
> 
> Three additional jobs that run in series with our existing CI jobs and duplicate much of the setup we already perform doesn't solve any problem we have. This will just make CI take longer than it already does, which IMO is too long.
> 
> If we want to check the diff is clean for generated code, we can do that as an added step in our existing CI steps and it will add very little complexity and very little build time.

I changed it back to only checking `git diff`.

PR #715, opened 2024-05-07
fix: Implement URLSessionTaskDelegate needNewBodyStream method

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1457

## Description of changes
- Implement the `urlSession(_:task:needNewBodyStream:)` delegate method.  This method must be implemented whenever `URLRequest` input body streaming is used.  When called, the `FoundationStreamBridge` replaces its bound stream pair and passes the new `InputStream` to the `URLSessionTask`.
- In practice, this method gets called when there are multiple concurrent HTTP requests to the same host, so without it we were seeing intermittent connection failures on Amplify customers, when they attempted to authenticate to the server at app startup (this involves multiple concurrent requests to AWS Cognito.)

Also:
- Configure the `URLSession` to never handle cookies, which are not set or used on AWS API endpoints.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
+++ b/Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
@@ -32,6 +32,10 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
     /// at any given time.
     let bridgeBufferSize: Int
 
+    /// The max number of bytes to buffer between the Foundation `OutputStream` and the Foundation `InputStream`
+    /// at any given time.
+    let boundStreamBufferSize: Int
+
     /// A buffer to hold data that has been read from the `ReadableStream` but not yet written to the
     /// Foundation `OutputStream`.  At most, it will contain `bridgeBufferSize` bytes.
     private var buffer: Data
@@ -42,10 +46,14 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
     let readableStream: ReadableStream
 
     /// A Foundation stream that will carry the bytes read from the readableStream as they become available.
-    let inputStream: InputStream
+    ///
+    /// May be replaced if needed by calling the `replaceStreams(_:)` method.
+    var inputStream: InputStream
 
-    /// A Foundation `OutputStream` that will read from the `ReadableStream`
-    private let outputStream: OutputStream
+    /// A Foundation `OutputStream` that will read from the `ReadableStream`.
+    ///
+    /// Will be replaced when `replaceStreams(_:)` is called to replace the input stream.
+    private var outputStream: OutputStream
 
     /// A Logger for logging events.
     private let logger: LogAgent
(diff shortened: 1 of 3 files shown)


Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -42,10 +46,14 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
+    ///
+    /// May be replaced if needed by calling the `replaceStreams(_:)` method.
+    var inputStream: InputStream
 
-    /// A Foundation `OutputStream` that will read from the `ReadableStream`
-    private let outputStream: OutputStream
+    /// A Foundation `OutputStream` that will read from the `ReadableStream`.
+    ///
+    /// Will be replaced when `replaceStreams(_:)` is called to replace the input stream.
+    private var outputStream: OutputStream
```
jbelkins, line 56: Input & output Foundation streams are now `var`s because they may need to be replaced during a HTTP request.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -110,32 +118,54 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
     ) {
+        self.bridgeBufferSize = bridgeBufferSize
+        self.boundStreamBufferSize = boundStreamBufferSize ?? bridgeBufferSize
+        self.buffer = Data(capacity: bridgeBufferSize)
+        self.readableStream = readableStream
+        self.logger = logger
+        (inputStream, outputStream) = Self.makeStreams(boundStreamBufferSize: self.boundStreamBufferSize, queue: queue)
+    }
+
+    func replaceStreams(completion: @escaping (InputStream?) -> Void) async {
```
jbelkins, line 129: This method will be called by the HTTP client when a new body stream is needed.

The output stream is closed, both streams are replaced with new ones & configured, and the output stream is reopened.

Note that I never do anything to the input stream, only the URLSessionDataTask ever manipulates that stream.

Sources/ClientRuntime/Networking/Http/URLSession/FoundationStreamBridge.swift
```diff
@@ -110,32 +118,54 @@ class FoundationStreamBridge: NSObject, StreamDelegate {
...
+
+        // Call the completion block.  When this method is called from `urlSession(_:task:needNewBodyStream:)`,
+        // the completion block will be that method's completion handler.
+        completion(inputStream)
+
+        // Re-open the `OutputStream` for writing.
+        await open()
+    }
+
+    private static func makeStreams(boundStreamBufferSize: Int, queue: DispatchQueue) -> (InputStream, OutputStream) {
```
jbelkins, line 145: This method just extracts the stream creation logic from the initializer so it can be used to replace the streams as well as at this class's initialization.

Sources/ClientRuntime/Networking/Http/URLSession/URLSessionConfiguration+HTTPClientConfiguration.swift
```diff
@@ -14,6 +14,9 @@ extension URLSessionConfiguration {
     public static func from(httpClientConfiguration: HttpClientConfiguration) -> URLSessionConfiguration {
         let config = URLSessionConfiguration.default
         config.timeoutIntervalForRequest = httpClientConfiguration.socketTimeout
+        config.httpShouldSetCookies = false
+        config.httpCookieAcceptPolicy = .never
+        config.httpCookieStorage = nil
```
jbelkins, line 19: Not strictly related to the rest of this PR but this should be done... the URLSession is configured to ignore cookies completely.

AWS HTTP APIs should never use cookies, but this guards against any weirdness from a proxy or gateway a customer may use.

(lol been bit by this before on other apps, hence I'm adding it here)

dayaffe (approved):

sichanyoo (approved):

PR #716, opened 2024-05-07
feat: Add deprecated shape remover

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1026

## Description of changes

- Adds model preprocessing that removes shapes with deprecated trait based on `since` property if earlier than cutoff date.

## Scope


jbelkins (commented):
The code looks good.

My larger question is whether this logic can lead to breakage where something that depends on a deprecated & removed shape will break (i.e. what if there is a member shape that targets a shape deprecated before our GA.)  Not sure if Smithy validates these deprecation traits to make sure that things that depend on deprecated shapes are themselves deprecated.

If it's indeed the case that Smithy validates against that or otherwise protects against this scenario, I'll approve.

jbelkins (changes requested):
Something that just occurred to me on this:

This belongs in SDK, not smithy-swift.

Move this over to aws-sdk-swift and I will re-review.

sichanyoo:
@jbelkins 

Smithy does protect against that. Using `ModelTransformer::removeShapesIf` ensures the model is left in consistent state after removal (any reference to removed shape is removed as well).

Re: this change being appropriate for SDK side, that's a great catch. I'll move changes there and submit PR soon.

sichanyoo:
Remover change moved to aws-sdk-swift: https://github.com/awslabs/aws-sdk-swift/pull/1511

PR #719, opened 2024-05-08
feat: update http requests, responses and endpoint with URI

AndrewFossAWS:
## Description of changes
Update http requests, responses and endpoint with uri to align with Smithy SRA design


Sources/ClientRuntime/Networking/Endpoint.swift
```diff
@@ -64,14 +71,16 @@ extension Endpoint {
...
+        components.host = uri.host.isEmpty ? nil : uri.host // If host is empty, URL is invalid
+        components.percentEncodedPath = uri.path
+        components.percentEncodedQuery = query
         return (components.host == nil || components.scheme == nil) ? nil : components.url
     }
 
-    var queryItemString: String? {
-        guard let queryItems = queryItems else { return nil }
+    var query: String? {
+        if (queryItems.isEmpty) {
```
sichanyoo, line 81: nit: swiftlint

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -17,18 +17,14 @@ import struct Foundation.URLRequest
...
     public var body: ByteStream
-    public let endpoint: Endpoint
+    public var endpoint: Endpoint
     public let method: HttpMethodType
-    private var additionalHeaders: Headers = Headers()
-    public var headers: Headers {
-        var allHeaders = endpoint.headers ?? Headers()
-        allHeaders.addAll(headers: additionalHeaders)
-        return allHeaders
-    }
```
sichanyoo, line 27: Note: this was done as a temporary measure to accommodate a customer request (https://github.com/smithy-lang/smithy-swift/issues/583) for allowing injection of new headers. But now that interceptors are implemented, this can be deleted.

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @discardableResult
+    public func withPath(_ value: String) -> URIBuilder {
+        self.path = value
+        if self.path.contains("%") {
+            self.urlComponents.percentEncodedPath = self.path
+        } else {
+            self.urlComponents.path = self.path
+        }
```
milesziemer, line 86: Instead of this check, you could just allow setting percent-encoded path and normal path, like URLComponents does

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+    public func withPort(_ value: Int16) -> URIBuilder {
+        self.port = value
+        return self
+    }
+
+    @discardableResult
+    public func withQueryItems(_ value: [SDKURLQueryItem]) -> URIBuilder {
+        self.query.append(contentsOf: value)
+        if !self.query.isEmpty {
+            self.urlComponents.percentEncodedQuery = self.query.map { queryItem in
```
milesziemer, line 107: Why not set this when `query` is empty? Also, do we know that `query` will be properly encoded (since we're assigning to `percentEncodedQuery`)? Rather than build the query string, you could map `SDKURLQueryItem` to `Foundation.URLQueryItem` - they're the same.

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    required public init() {
+        self.urlComponents = URLComponents()
+        self.urlComponents.scheme = self.scheme.rawValue
+        self.urlComponents.path = self.path
+        self.urlComponents.host = self.host
+    }
+
+    @discardableResult
+    public func withScheme(_ value: Scheme?) -> URIBuilder {
```
milesziemer, line 73: Why make `value` optional?

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public func build() -> URI {
+        return URI(scheme: self.scheme,
+                   path: self.path,
+                   host: self.host,
+                   port: self.port,
+                   query: self.query,
+                   username: self.username,
+                   password: self.password)
```
milesziemer, line 140: If my understanding is correct, `URIBuilder` is using `URLComponents` to throw errors when an invalid value is passed. But `URLComponents` also percent-encodes stuff when necessary - like if you did `urlComponents.query = "#"`, it'll encode the query to be `"%23"`. I think you could just use `URLComponents` instead of storing the other members.

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    @discardableResult
+    public func withPort(_ value: Int16) -> URIBuilder {
+        self.port = value
+        return self
+    }
+
+    @discardableResult
+    public func withQueryItems(_ value: [SDKURLQueryItem]) -> URIBuilder {
+        self.query.append(contentsOf: value)
```
milesziemer, line 105: Maybe more of a naming thing about `withQueryItems` vs `withQueryItem`, but appending here means you can't overwrite the query items

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+        self.toBuilder().getUrl()
+    }
+    public var queryString: String? {
+        if self.query.isEmpty {
+            return nil
+        }
+
+        return self.query.map { queryItem in
+            return [queryItem.name, queryItem.value].compactMap { $0 }.joined(separator: "=")
+        }.joined(separator: "&")
```
milesziemer, line 26: I think I saw this logic somewhere else too, you might be able to extract it into an extension method/property on `Array<SDKURLQueryItem>`

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import Foundation
+
+public struct URI: Hashable {
```
milesziemer, line 8: `URI` and `URIBuilder` need docs

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,148 @@
...
+    var urlComponents: URLComponents
+    var scheme: Scheme = Scheme.https
+    var path: String = "/"
+    var host: String = ""
+    var port: Int16 = Int16(Scheme.https.port)
+    var query: [SDKURLQueryItem] = []
+    var username: String?
+    var password: String?
+
+    required public init() {
```
milesziemer, line 65: Can avoid the `required` I think by making `URIBuilder` `final`

milesziemer (changes requested):

Sources/ClientRuntime/Networking/Http/URI.swift
AndrewFossAWS: The url will be `http://example.amazonaws.com?` if an empty query is set.

Sources/ClientRuntime/Networking/Http/URI.swift
milesziemer: A consequence of this though is that callers of `URI.path` now can't know whether the `URI` is percent encoded or not. I guess they use the String extension you added though. But if they wanted it to be encoded and it's not, they have to do it themselves?

Sources/ClientRuntime/Networking/Http/URI.swift
milesziemer: Oh yea I see `URL` and `URLComponents` don't strip the `?` (or omit the port, as I see from your other [comment](https://github.com/smithy-lang/smithy-swift/pull/719#discussion_r1600803964)). How important is the string representation though? I mean if it is that important, we couldn't omit the port anyways, because then it wouldn't be an accurate URI. And for `queryItems` here, they get defaulted to an empty array when constructing `URI` anyways.

milesziemer (changes requested):

Sources/ClientRuntime/Networking/Http/URI.swift
```diff
@@ -0,0 +1,173 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import Foundation
+
+/// Universal Resource Identifier component used to construct target location for a Request.
+public struct URI: Hashable {
```
milesziemer, line 9: I suggest moving URI to the `Message` directory - it isn't necessarily part of networking (you can have a `file://` uri), and def not just http.

AndrewFossAWS:
`port` in `SdkHttpRequest` and `Endpoint` have been updated to be nullable. This allows `URL`/url string to be constructed without port being auto-populated. In aws-sdk-swift, the presigned request endpoint url is also invalid if a port auto-populated.

A `defaultPort` member has also been added to URI to provide a default port if needed.

Sources/ClientRuntime/Message/URI.swift
```diff
@@ -0,0 +1,180 @@
...
+
+extension Array where Element == SDKURLQueryItem {
+    public var queryString: String? {
+        if self.isEmpty {
+            return nil
+        }
+        return self.map { [$0.name, $0.value].compactMap { $0 }.joined(separator: "=") }.joined(separator: "&")
+    }
+
+    public var toURLQueryItems: [URLQueryItem] {
```
milesziemer, line 171: Nit: naming of `to` sounds like a method, maybe this makes more sense as a method anyway, not sure. Same thing for `toSDKURLQueryItems` below

Tests/ClientRuntimeTests/NetworkingTests/Http/CRTClientEngineIntegrationTests.swift
```diff
@@ -32,7 +32,7 @@ class CRTClientEngineIntegrationTests: NetworkingTestUtils {
         var headers = Headers()
         headers.add(name: "Content-type", value: "application/json")
         headers.add(name: "Host", value: "httpbin.org")
-        let request = SdkHttpRequest(method: .get, endpoint: Endpoint(host: "httpbin.org", path: "/get", headers: headers))
+        let request = SdkHttpRequest(method: .get, endpoint: try Endpoint(host: "httpbin.org", path: "/get", headers: headers))
```
milesziemer, line 35: Might be missing something, but I don't think this ctor throws

milesziemer: There's a few places `try` is being used, by the way

Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
```diff
@@ -493,7 +493,7 @@ class OrchestratorTests: XCTestCase {
             let b = self.traceOrchestrator(trace: partitionIdTrace)
             // We can force a getPartitionId to fail by explicitly setting the host to ''
             b.interceptors.addModifyBeforeRetryLoop({ context in
-                context.updateRequest(updated: context.getRequest().toBuilder().withHost("").build())
+                context.updateRequest(updated: try context.getRequest().toBuilder().withHost("").build())
```
milesziemer, line 496: Same thing here - might be missing something but I don't think this method throws. It's in a few places

Tests/ClientRuntimeTests/NetworkingTests/Http/CRTClientEngineIntegrationTests.swift
AndrewFossAWS: Removed all the `try`s

Sources/ClientRuntime/Message/URI.swift
AndrewFossAWS: Updated to functions

Sources/ClientRuntime/Message/URI.swift
```diff
@@ -0,0 +1,197 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import Foundation
+
+/// A representation of the RFC 3986 Universal Resource Identifier
```
milesziemer, line 8: `Universal` -> `Uniform`

Tests/ClientRuntimeTests/MessageTests/URITests.swift
```diff
@@ -69,4 +69,26 @@ class URITests: XCTestCase {
...
+            .withPath("/x%2Dy%2Dz")
+            .withHost("xctest2.amazonaws.com")
+            .appendQueryItem(SDKURLQueryItem(name: "test", value: "1%2B2"))
+            .withFragment("fragment%21")
+            .withUsername("dan")
+            .withPassword("008")
+            .build()
+
+        XCTAssertEqual(uri.url?.absoluteString,
+           "https://dan:008@xctest2.amazonaws.com/x%2Dy%2Dz?abc=def&ghi=jkl&mno=pqr&test=1%2B2#fragment%21")
```
milesziemer, line 92: Add some some tests where the values being passed into `withPath`, `withHost`, etc _aren't_ already encoded, but should be. I tried it with `host` and found that it isn't being encoded.

milesziemer (changes requested):

milesziemer (dismissed):
This PR needs more test cases to lock down the expected percent-encoding of `URI` components. Especially considering the inconsistency we found with `URLComponents.percentEncodedHost` between Linux and Mac (discussed offline, but see [comment here](https://github.com/smithy-lang/smithy-swift/pull/719/files#diff-034346e4c429867fffbc386c855d0a4b4b6f95f1c92aab3875f097737ed98d7aR92-R95)). We should add cases for when different components are set to values that contain unencoded reserved characters, and a mix of encoded/unencoded reserved characters.

Sources/ClientRuntime/Message/URI.swift
```diff
@@ -0,0 +1,273 @@
...
+
+extension String {
+    var isPercentEncoded: Bool {
+        let decoded = self.removingPercentEncoding
+        return decoded != nil && decoded != self
+    }
+
+    public func percentEncodePathIncludingSemicolon() -> String {
+        let allowed =
+            // swiftlint:disable:next force_cast
```
sichanyoo, line 237: The disabled lint category needs to be enabled after the relevant changes. Otherwise the lint category stays disabled for the entire file. I see the force cast only happens once in this file, but still should have the enable statement afterwards to future-proof it.

Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift
```diff
@@ -155,9 +161,11 @@ extension SdkHttpRequest: CustomDebugStringConvertible, CustomStringConvertible
 
     public var description: String {
         let method = method.rawValue.uppercased()
-        let protocolType = endpoint.protocolType ?? ProtocolType.https
-        let query = String(describing: queryItems)
-        return "\(method) \(protocolType):\(endpoint.port) \n Path: \(endpoint.path) \n \(headers) \n \(query)"
+        let protocolType = self.destination.scheme
+        let query = self.destination.queryString ?? ""
+        let port = self.destination.port
+        return "\(method) \(protocolType):\(port) \n " +
```
sichanyoo, line 167: Handle swift warning that gets raised here. One way to do so is by providing default value after nil-coalescing operator like this: `:\(port ?? "no-port-found")`

sichanyoo (changes requested):
Comments on addressing swiftlint warning & a compiler warning

sichanyoo (approved):

PR #727, opened 2024-05-15
feat: add log for failed refresh of retry token

dayaffe:
## Scope

### DIFF
--- a/Sources/ClientRuntime/Middleware/RetryMiddleware.swift
+++ b/Sources/ClientRuntime/Middleware/RetryMiddleware.swift
@@ -61,7 +61,7 @@ public struct RetryMiddleware<Strategy: RetryStrategy,
             do {
                 try await strategy.refreshRetryTokenForRetry(tokenToRenew: token, errorInfo: errorInfo)
             } catch {
-                // TODO: log token error here
+                context.getLogger()!.error("Failed to refresh retry token!")
                 throw operationError
             }
             var estimatedSkew = context.attributes.get(key: AttributeKeys.estimatedSkew)


sichanyoo (approved):

Sources/ClientRuntime/Middleware/RetryMiddleware.swift
```diff
@@ -61,18 +61,18 @@ public struct RetryMiddleware<Strategy: RetryStrategy,
             do {
                 try await strategy.refreshRetryTokenForRetry(tokenToRenew: token, errorInfo: errorInfo)
             } catch {
-                // TODO: log token error here
+                context.getLogger()?.error("Failed to refresh retry token: \(errorInfo)")
```
jbelkins, line 64: What does printing `errorInfo` actually log?

dayaffe: I changed this to `errorInfo.errorType` which will print out the type of error based on RetryErrorType enum:

```
/// This is a connection level error such as a socket timeout, socket connect error, tls negotiation timeout etc...
/// Typically these should never be applied for non-idempotent, request types since in this scenario, it’s impossible
/// to know whether the operation had a side effect on the server.
case transient

/// This is an error where the server explicitly told the client to back off, such as a 429 or 503 HTTP error.
case throttling

/// This is a server error that isn’t explicitly throttling but is considered by the client to be something that should be retried.
case serverError

/// Doesn’t count against any budgets.  This could be something like a 401 challenge in HTTP.
case clientError
```

jbelkins (approved):

PR #733, opened 2024-05-21
chore: Update to Smithy 1.49.0

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1513

## Description of changes
Updates Smithy from 1.47.0 to 1.49.0.

There are corresponding changes in aws-sdk-swift PR https://github.com/awslabs/aws-sdk-swift/pull/1515 .

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/AuthSchemeResolverGenerator.kt
```diff
@@ -285,7 +285,7 @@ fun Parameter.toSymbol(): Symbol {
     val swiftType = when (type) {
         ParameterType.STRING -> SwiftTypes.String
         ParameterType.BOOLEAN -> SwiftTypes.Bool
-        else -> throw CodegenException("Unsupported parameter type: $type")
+        ParameterType.STRING_ARRAY -> SwiftTypes.StringArray
```
jbelkins, line 288: Add `STRING_ARRAY` case and eliminate the `else` so this fails to compile when a new `ParameterType` is added in the future.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/endpoints/EndpointParamsGenerator.kt
```diff
@@ -86,7 +86,7 @@ fun Parameter.toSymbol(): Symbol {
     val swiftType = when (type) {
         ParameterType.STRING -> SwiftTypes.String
         ParameterType.BOOLEAN -> SwiftTypes.Bool
-        else -> throw CodegenException("Unsupported parameter type: $type")
+        ParameterType.STRING_ARRAY -> SwiftTypes.StringArray
```
jbelkins, line 89: Same change as in `AuthSchemeResolverGenerator` above.

sichanyoo (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftTypes.kt
```diff
@@ -16,6 +16,8 @@ object SwiftTypes {
     val Double = builtInSymbol("Double")
     val Bool = builtInSymbol("Bool")
 
+    val StringArray = builtInSymbol("Array<Swift.String>")
```
sichanyoo, line 19: Might be good idea to make this `[String]` to make it idiomatic

sichanyoo (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftTypes.kt
jbelkins: This form is needed because when the SwiftWriter prints this symbol, it will preface it with the namespace:
```
Swift.Array<Swift.String>
```
Awkward but it compiles.  On the other hand:
```
Swift.[Swift.String]
```
does not compile: Swift doesn't allow tacking a namespace on when you declare an Array type with `[...]` syntax.

sichanyoo (approved):

PR #738, opened 2024-05-23
feat: Add OS version info to user agent for Linux

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1206

## Description of changes

- Use the `uname` POSIX function available in all Linux distros to get the release version of the OS & include that in User-Agent header of requests.

## Scope


dayaffe (changes requested):
Please see my comment on using ProcessInfo from swift-foundations instead of using Glibc

PR #742, opened 2024-05-29
feat!: Remove unknown case from allCases in int & string enums

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1463

## Description of changes
Removes the `.sdkUnknown()` case from the `allCases` property on both Int- & String-backed enums, because the case is not meaningful in the context of iterating or discovering the enum's cases.

## Scope

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -63,6 +63,7 @@ let package = Package(
         .target(
             name: "SmithyRetries",
             dependencies: [
+                "SmithyRetriesAPI",
                 .product(name: "AwsCommonRuntimeKit", package: "aws-crt-swift"),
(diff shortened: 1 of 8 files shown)


Package.swift
```diff
@@ -63,6 +63,7 @@ let package = Package(
         .target(
             name: "SmithyRetries",
             dependencies: [
+                "SmithyRetriesAPI",
```
jbelkins, line 66: This was an omission from the retries project that can lead to compile failures under certain conditions.

sichanyoo (approved):

PR #765, opened 2024-06-19
feat: Add copyright notice extension point

sichanyoo:
Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1581
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1547

## Description of changes

- Adds customization point for copyright notice via `smithy-build.json`'s `copyrightNotice` property, which gets read by `SwiftSettings`.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
```diff
@@ -90,6 +90,7 @@ class SwiftWriter(private val fullPackageName: String, swiftImportContainer: Swi
                         sectionWriter.write(codeWriter, previousValue)
                     }
                 }
+                GENERATED_FILE_HEADER = integration.copyrightNotice() ?: GENERATED_FILE_HEADER
```
jbelkins, line 93: I think we want to keep the `Code generated by smithy-swift-codegen` notice, but also add in the custom header right?

sichanyoo: Yeah, but that's done not by appending but just including it in AWS side as well.

jbelkins: I guess it allows customers to use their own "do not edit" warning as well, if they want to.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
```diff
@@ -90,6 +90,7 @@ class SwiftWriter(private val fullPackageName: String, swiftImportContainer: Swi
                         sectionWriter.write(codeWriter, previousValue)
```
jbelkins, line 90: Since this is in an `integrations.forEach { ... }` loop, I presume that if multiple integrations have notices, then the last one to be iterated will be the one that is used and the others are ignored?

jbelkins (commented):
see Q

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
sichanyoo: Yes, that is correct; only one should be setting copyright notice.

sichanyoo: Re: offline comment on whether `SwiftIntegration` is the right mechanism for this, I think it should be fine. Just need to ensure only one integration overrides the `copyrightNotice` method.

sichanyoo: I'll include that in the doc comment for the method.

sichanyoo: Changed to use `SwiftSettings` instead.

jbelkins (approved):

PR #767, opened 2024-06-25
feat: overhaul protocol test generation to use actual client

dayaffe:
## Description of changes
- Modify protocol test client generation to use the generated client rather than construct a set of middlewares in the test.
- Update CRT to 0.31
- Adds ProtocolTestClient() for intercepting request before it is sent. This allows protocol tests to assert against the request.
- Add a StaticEndpointResolver() per SRA which allows customers to specify an endpoint to use rather than rely on endpoint resolution. This is also used for testing to bypass not having a RuleSet in regular protocol tests.
- Adds `checkStreamLengthEqualityIfUnseekableActualStreamHasBeenRead` which takes care of an edge case where an unseekable stream has already been read by the signer so we cannot read its data during test assertions. Instead we can check for that condition and compare stream length to expected length (better than no check)

## Scope


Sources/SmithyTestUtil/RequestTestUtil/HttpRequestTestBase.swift
```diff
@@ -378,7 +423,14 @@ open class HttpRequestTestBase: XCTestCase {
...
             )
-            XCTAssertTrue(values.contains(expectedQueryItem.value),
+            let expectedValue = expectedQueryItem.value
+            let valueExists: Bool
+            if let expectedValue = expectedValue {
+                valueExists = values.contains(expectedValue)
+            } else {
+                valueExists = values.isEmpty
+            }
+            XCTAssertTrue(valueExists,
```
dayaffe, line 433: Test was failing as `nil != []` but these should be equivalent

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/endpoints/EndpointResolverGenerator.kt
```diff
@@ -62,4 +63,17 @@ class EndpointResolverGenerator(
         writer.write("")
         writer.write("extension DefaultEndpointResolver: EndpointResolver {}")
     }
+
+    private fun renderStaticResolver(writer: SwiftWriter) {
```
dayaffe, line 67: Static Resolver will be used by tests but are also part of SRA to allow users to bypass endpoint resolution and provide their own endpoint

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -29,20 +32,33 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
...
+
+        // Normalize the URI
+        val normalizedUri = when {
+            test.uri == "/" -> "/"
+            test.uri.isEmpty() -> ""
+            else -> {
+                val trimmedUri = test.uri.removeSuffix("/")
+                if (!trimmedUri.startsWith('/')) "/$trimmedUri" else trimmedUri
+            }
+        }
```
dayaffe, line 51: Tests were failing due to a missing preceding or ending `/`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -29,20 +32,33 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
...
         renderExpectedBlock(test)
         writer.write("")
+        renderClientBlock(test)
         renderOperationBlock(test)
     }
 
     private fun renderExpectedBlock(test: HttpRequestTestCase) {
-        var resolvedHostValue = test.resolvedHost?.let { it } ?: run { "nil" }
+        var resolvedHostValue = if (test.resolvedHost.isPresent && test.resolvedHost.get() != "") test.resolvedHost.get() else "example.com"
+        var hostValue = if (test.host.isPresent && test.host.get() != "") test.host.get() else "example.com"
```
dayaffe, line 41: Provide a host (example.com) to be used during tests that do not specify one

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -68,8 +84,26 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
         }
     }
 
+    private fun renderClientBlock(test: HttpRequestTestCase) {
```
dayaffe, line 87: This block renders the client call and its config options

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -68,8 +84,26 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
...
+
+        if (!serviceShape.getTrait(EndpointRuleSetTrait::class.java).isPresent) {
+            val host: String? = test.host.orElse(null)
+            val url: String = "http://${host ?: "example.com"}"
+            writer.write("\nlet config = try await ${clientName}.${clientName}Configuration(endpointResolver: StaticEndpointResolver(endpoint: try \$N(urlString: \$S)))", SmithyHTTPAPITypes.Endpoint, url)
+        } else {
+            writer.write("\nlet config = try await ${clientName}.${clientName}Configuration()")
+        }
+        writer.write("config.region = \"us-west-2\"")
+        writer.write("config.httpClientEngine = ProtocolTestClient()")
```
dayaffe, line 99: Use the ProtocolTestClient() as the httpClientEngine to ensure we get the request back and never send it

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -68,8 +84,26 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(builder: B
...
+        writer.write("config.idempotencyTokenGenerator = ProtocolTestIdempotencyTokenGenerator()")
```
dayaffe, line 100: Use our own token generator to ensure the expected test token will get generated during middleware execution

Sources/SmithyTestUtil/RequestTestUtil/HttpRequestTestBase.swift
```diff
@@ -237,6 +264,17 @@ open class HttpRequestTestBase: XCTestCase {
...
         let actualData = try await actual.readData()
+
+        // Unseekable streams may have already been read by Signer middleware and cannot be read again
+        // Compare stream lengths if ByteStream is a .stream and actualData is nil
+        if checkStreamLengthEqualityIfUnseekableActualStreamHasBeenRead(
+            expected: expected, actual: actual, file: file, line: line
+        ) {
+            // Stream lengths were checked and comparing data will result in failure due to above conditions
+            return
+        }
```
dayaffe, line 275: There is an edge case where the SignerMiddleware has already read an unseekable stream. In which case by the time all middlewares are executed we cannot re-read the data. In that case let's check if the actual and expected stream lengths are the same.

Note: This comparison should only fail if the actualStream.position == actualStream.length and !actualStream.isSeekable

Sources/SmithyTestUtil/ProtocolTestClient.swift
```diff
@@ -0,0 +1,31 @@
...
+    public init() {}
+}
+
+public enum TestCheckError: Error {
+    case actual(SdkHttpRequest)
+}
+
+extension ProtocolTestClient: HTTPClient {
+    public func send(request: SdkHttpRequest) async throws -> HttpResponse {
+        throw TestCheckError.actual(request)
```
dayaffe, line 21: throwing an error with the request will allow tests to do checks against the result

Sources/ClientRuntime/Networking/Http/Middlewares/SignerMiddleware.swift
```diff
@@ -69,10 +70,17 @@ extension SignerMiddleware: ApplySigner {
             )
         }
 
+        // Check if CRT should be provided a pre-computed Sha256 SignedBodyValue
+        var updatedSigningProperties = signingProperties
+        let sha256: String? = attributes.get(key: AttributeKey(name: "X-Amz-Content-Sha256"))
+        if let bodyValue = sha256 {
+            updatedSigningProperties.set(key: AttributeKey(name: "SignedBodyValue"), value: sha256)
+        }
```
dayaffe, line 78: SHA256Middleware if executed will set this attribute to be picked up by the SignerMiddleware and put into the signing properties

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/endpoints/EndpointResolverGenerator.kt
```diff
@@ -62,4 +63,17 @@ class EndpointResolverGenerator(
...
+        writer.write("")
+        writer.openBlock("public struct StaticEndpointResolver: EndpointResolver {", "}") {
```
jbelkins, line 69: Can we provide a reusable static endpoint resolver?

The only difference between this endpoint resolver for each service is that the `resolve` method takes that client's `EndpointParams` so we can make that type generic in a reusable static endpoint resolver, then simply use a typedef in each client to provide a specialized endpoint resolver.

jbelkins (changes requested):
See my question about a reusable static endpoint resolver.

Other than that, all is good.

jbelkins (approved):

PR #781, opened 2024-07-13
feat: finish interceptors migration

milesziemer:
## Description of changes

aws-sdk-swift pr: https://github.com/awslabs/aws-sdk-swift/pull/1630

Removes all Middleware types, references, and implementations, completing migration to using Orchestrator + Interceptors. Tests using middleware were also migrated to use Orchestrator + Interceptors. Codegen was updated, removing the `useInterceptors` setting, switching all code that was either generating middleware or interceptors to just generate interceptors. This also allowed removing some other existing generator code around middleware steps.

As with the corresponding aws-sdk-swift commit, this is already a large commit so I wanted to avoid doing too much refactoring or moving code around. For example, I didn't change file names or type names. There's probably a good deal of refactoring that can be done, but it is out of scope for this PR.

Also, this commit removes the amz-sdk-invocation-id/amz-sdk-request header impl as it was moved to aws-sdk-swift in the corresponding commit.

## Scope

- [x] Breaking change - Removes a bunch of types that are no longer used


milesziemer:
Closing for https://github.com/smithy-lang/smithy-swift/pull/782 so CI will run

PR #794, opened 2024-08-07
fix: Enable EC2 query empty list protocol test

sichanyoo:
Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1666

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1514

## Description of changes

- Empty lists in requests get serialized as nothing in EC2 query protocol. Codegen conditional that checks for empty list to determine whether to serialize list or not.

## Scope


jbelkins (approved):
I'm not a fan of using AWS protocols here in smithy-swift, but (1) we already do it here and (2) these references to AWS will likely get moved back to the SDK when we do schema-based serde.

PR #796, opened 2024-08-08
fix: Delete early exit check for unsigned chunk

sichanyoo:
## Scope

### DIFF
--- a/Sources/SmithyChecksums/ChunkedReader.swift
+++ b/Sources/SmithyChecksums/ChunkedReader.swift
@@ -114,11 +114,6 @@ public class ChunkedReader {
 
         self.chunkBody = chunk
 
-        // Early exit for empty chunk when already signed
-        if chunk.isEmpty {
-            return nil
-        }
-
         return constructChunk(chunk: chunk, signature: nil)
     }
 


Sources/SmithyChecksums/ChunkedReader.swift
```diff
@@ -114,11 +114,6 @@ public class ChunkedReader {
```
sichanyoo, line 114: This early exit for unsigned chunk was quitting early before entirety of body is streamed, making S3 return `IncompleteBody` error.

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -282,11 +282,13 @@ public class CRTClientEngine: HTTPClient {
                                         }
 
                                         var hasMoreChunks = true
```
sichanyoo, line 284: Changes here is to match how `ChunkedReader` is used by `URLSessionHTTPClient` via `ChunkedStream`. Before the change, the logic was slightly different.

sichanyoo (commented):
Comments for reviewers

dayaffe:
please fix the github warnings!

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -282,11 +282,13 @@ public class CRTClientEngine: HTTPClient {
...
+                                        var currentChunkBodyIsEmpty = false
                                         while hasMoreChunks {
                                             // Process the first chunk and determine if there are more to send
                                             hasMoreChunks = try await chunkedStream.chunkedReader.processNextChunk()
+                                            currentChunkBodyIsEmpty = chunkedStream.chunkedReader.getCurrentChunkBody().isEmpty
```
dayaffe, line 289: swiftlint warning! -- we should change these to failures rather than warns since we don't ever approve PRs that have this, will take this up

dayaffe (approved):
Please just take care of any relevant warnings!

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
jbelkins: The bigger issue here is that this is not a long line of code.  It's so far nested that much of the line length is just leading space.  It suggests this method should be broken into submethods

jbelkins (approved):

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
sichanyoo: Addressed warning.

Issue #792, opened 2024-08-02
Unable to set up SDKLoggingSystem


shepazon:
### Describe the bug

Since the recent concurrency warning fixes, trying to use `SDKLoggingSystem` in a project fails. Specifically, if I put this code into my project:

```
    await SDKLoggingSystem.initialize(logLevel: .trace)
````

I get a compiler error: `Instance member 'initialize' cannot be used on type 'SDKLoggingSystem'; did you mean to use a value of this type instead?`.

If I try to create a value of type `SDKLoggingSystem` like this:

```
    let loggingSystem = SDKLoggingSystem()
```

I get this compiler error instead: `'SDKLoggingSystem' initializer is inaccessible due to 'internal' protection level`

### Expected Behavior

The logging system should be able to be initialized and used, but cannot be.

### Current Behavior

Errors described in description.

### Reproduction Steps

See description.

### Additional Information/Context

If I back my code up to an earlier version, such as 0.47.0

### AWS SWIFT SDK version used

0.49.0

### Compiler and Version used

Xcode 15.4, Swift 5.10.0

### Operating System and version

macOS 14.6

sichanyoo:
Went out with the release of v0.58.0: https://github.com/smithy-lang/smithy-swift/releases/tag/0.58.0. Closing.

PR #797, opened 2024-08-09
chore: Misc. logging changes

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1654
- https://github.com/smithy-lang/smithy-swift/issues/792

## Description of changes

- Add `.none` case to `ClientLogMode` and make that default to reduce request / response logging unless opted in
- Adjust log levels for logging statements in `URLSessionHTTPClient` and `FoundationStreamBridge`
- Add a public initializer for `SDKLoggingSystem` and make default log level `.error` for `SDKLoggingSystem::initialize(defaultLogLevel:)`
- Fix `SwiftLogger::log` to pass in the log level of the logger instance to `Logger::log`, instead of using the argument value passed to it, which produced wrong behavior (e.g., before this fix, a `SwiftLogger` with `logLevel` of `.critical` would log out all levels of log without filter because instead of passing in `.critical`, it would pass in whatever log level that was the argument)

## Scope


jbelkins (changes requested):
Defaulting to `.none` log level means that errors will not be logged by default.  I don't think we want that... the default log level should be `.error` IMO

jbelkins (approved):

PR #803, opened 2024-08-28
feat: Add default trait support

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1330

## Description of changes

- Add default trait handling for codegen for both request and response flows
- Refer to comments on GH diff for details

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -104,21 +110,21 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
-    override fun byteShape(shape: ByteShape): Symbol = numberShape(shape, "Int8", "0")
+    override fun byteShape(shape: ByteShape): Symbol = numberShape(shape, "Int8")
 
-    override fun shortShape(shape: ShortShape): Symbol = numberShape(shape, "Int16", "0")
+    override fun shortShape(shape: ShortShape): Symbol = numberShape(shape, "Int16")
 
-    override fun bigIntegerShape(shape: BigIntegerShape): Symbol = numberShape(shape, "Int", defaultValue = "0")
+    override fun bigIntegerShape(shape: BigIntegerShape): Symbol = numberShape(shape, "Int")
 
-    override fun bigDecimalShape(shape: BigDecimalShape): Symbol = numberShape(shape, "Double", "0.0")
```
sichanyoo, line 121: Remove hardcoded zero default values given to number shapes.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -149,7 +155,7 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
     }
 
     override fun booleanShape(shape: BooleanShape): Symbol {
-        return createSymbolBuilder(shape, "Bool", namespace = "Swift", SwiftDeclaration.STRUCT).putProperty(SymbolProperty.DEFAULT_VALUE_KEY, "false").build()
```
sichanyoo, line 152: Remove hardcoded zero-equivalent default value given to boolean shapes

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -205,7 +211,7 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
                 .putProperty(SymbolProperty.NESTED_SYMBOL, symbol)
                 .build()
         }
-        return symbol
+        return handleDefaultValue(shape, symbol.toBuilder()).build()
```
sichanyoo, line 214: Before returning the built symbol for member shape, handle any default value it should have if applicable.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -243,25 +249,30 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+    ): Symbol.Builder {
         val builder = Symbol.builder()
             .putProperty("shape", shape)
             .putProperty("decl", declaration.keyword)
             .name(typeName)
         if (boxed) {
             builder.boxed()
         }
-        return builder
+        return handleDefaultValue(shape, builder)
```
sichanyoo, line 275: Before returning the builder, handle any default value it should have in its properties.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -270,7 +281,7 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
      * the namespace (and ultimately the package name) to `foo.bar` for the symbol.
      */
     private fun createSymbolBuilder(
-        shape: Shape?,
```
sichanyoo, line 273: Remove unnecessary optionality

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +295,109 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
         return ModelFileUtils.filename(swiftSettings, name)
     }
```
sichanyoo, line 297: The `handleDefaultValue` below extracts default value for the given shape, saves it to properties of the given symbol builder, then returns the modified builder back to the caller. If the shape does not or should not have a default value, it returns unmodified builder.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestResponseGenerator.kt
```diff
@@ -55,23 +60,63 @@ open class HttpProtocolUnitTestResponseGenerator protected constructor(builder:
...
+                    } else {
+                        writer.write("content: .data(\$L)", data)
+                    }
+                } else if (test.body.get().isBlank() && bodyHasDefaultValue()) {
+                    // Expected body is blank but not because it's nil, but because it's a default empty blob value
+                    writer.write("content: .data(Data(\"\".utf8))")
                 } else {
-                    writer.write("content: .data(\$L)", data)
+                    // Expected body is blank and underlying member shape does not have default values.
+                    writer.write("content: nil")
```
sichanyoo, line 87: Before we used to equate empty `body` string in the Smithy protocol test as `nil`. But an empty `body` can mean an empty blob of size 0 if one of the body members has a default value of an empty blob. Change here adds handling that edge case.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -147,6 +156,40 @@ open class MemberShapeDecodeGenerator(
...
+                    if (targetShape.hasTrait<StreamingTrait>()) {
+                        " ?? ByteStream.data(\"\".data(using: .utf8))"
+                    } else {
+                        " ?? \"\".data(using: .utf8)"
+                    }
+                }
+                // No default provided for other types
+                else -> { "" }
+            }
+        }
```
sichanyoo, line 192: Change here adds server-side serialization error where it fails to send values for required members of the output. In that case, the SDK fills it with zeros or zero-equivalents.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -88,12 +94,13 @@ open class MemberShapeDecodeGenerator(
...
         return writer.format(
-            "try \$L.\$L(memberReadingClosure: \$L, memberNodeInfo: \$L, isFlattened: \$L)",
+            "try \$L.\$L(memberReadingClosure: \$L, memberNodeInfo: \$L, isFlattened: \$L)\$L",
             reader(memberShape, false),
             readMethodName("readList"),
             memberReadingClosure,
             memberNodeInfo,
-            isFlattened
+            isFlattened,
+            default(memberShape)
```
sichanyoo, line 103: Add default value handling to deserializer for list shape

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -104,24 +111,26 @@ open class MemberShapeDecodeGenerator(
...
-            "try \$L.\$L(valueReadingClosure: \$L, keyNodeInfo: \$L, valueNodeInfo: \$L, isFlattened: \$L)",
+            "try \$L.\$L(valueReadingClosure: \$L, keyNodeInfo: \$L, valueNodeInfo: \$L, isFlattened: \$L)\$L",
             reader(memberShape, false),
             readMethodName("readMap"),
             valueReadingClosure,
             keyNodeInfo,
             valueNodeInfo,
-            isFlattened
+            isFlattened,
+            default(memberShape)
```
sichanyoo, line 121: Add default value handling to deserializer for map shape

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -104,24 +111,26 @@ open class MemberShapeDecodeGenerator(
...
         val memberTimestampFormatTrait = memberShape.getTrait<TimestampFormatTrait>()
         val swiftTimestampFormatCase = TimestampUtils.timestampFormat(ctx, memberTimestampFormatTrait, timestampShape)
         return writer.format(
-            "try \$L.\$L(format: \$L)",
+            "try \$L.\$L(format: \$L)\$L",
             reader(memberShape, false),
             readMethodName("readTimestamp"),
-            swiftTimestampFormatCase
+            swiftTimestampFormatCase,
+            default(memberShape)
```
sichanyoo, line 133: Add default value handling to deserializer for timestamp shape

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -168,6 +211,26 @@ open class MemberShapeDecodeGenerator(
...
+                        }
+                    }
+                }
+                is BlobShape -> {
+                    if (targetShape.hasTrait<StreamingTrait>()) {
+                        " ?? ByteStream.data(\"$it\".data(using: .utf8))"
+                    } else {
+                        " ?? \"$it\".data(using: .utf8)"
+                    }
+                }
```
sichanyoo, line 233: Add default value support to deserializer for timestamp, document, and blob shapes

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/SymbolExt.kt
```diff
@@ -42,12 +42,14 @@ fun Symbol.isBoxed(): Boolean {
...
-    if (isBoxed()) {
+    val default = getProperty(SymbolProperty.DEFAULT_VALUE_KEY, String::class.java)
+
+    // If shape is boxed (nullable) AND there is no default value set, return nil as default value
+    if (isBoxed() && !default.isPresent) {
         return "nil"
     }
 
-    val default = getProperty(SymbolProperty.DEFAULT_VALUE_KEY, String::class.java)
+    // If default value is present, return default value. Otherwise, return null
```
sichanyoo, line 52: Change logic to handle optional member that has a default value

sichanyoo (commented):
Comments for reviewers

Sources/SmithyTestUtil/JSONComparator.swift
```diff
@@ -47,6 +47,8 @@ fileprivate func anyValuesAreEqual(_ lhs: Any?, _ rhs: Any?) -> Bool {
         return anyDictsAreEqual(lhsDict, rhsDict)
     } else if let lhsArray = lhs as? [Any], let rhsArray = rhs as? [Any] {
         return anyArraysAreEqual(lhsArray, rhsArray)
+    } else if let lhn = lhs as? NSNumber, let rhn = rhs as? NSNumber {
+        return lhn == rhn
```
sichanyoo, line 51: Handles case where JSON values that are logically equivalent (e.g., `1` and `1.0`) but are different concrete specific types

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/Utils.kt
```diff
@@ -19,7 +19,7 @@ fun String.removeSurroundingBackticks() = removeSurrounding("`", "`")
  */
 fun swiftEnumCaseName(name: String?, value: String, shouldBeEscaped: Boolean = true): String {
     val resolvedName = name ?: value
-    var enumCaseName = CaseUtils.toCamelCase(resolvedName.replace(Regex("[^a-zA-Z0-9_ ]"), ""))
+    var enumCaseName = CaseUtils.toCamelCase(resolvedName.replace(Regex("[^a-zA-Z0-9_ -]"), ""))
```
sichanyoo, line 22: The `CaseUtils::toCamelCase` also uses the dash (`-`) as one of the delimiters for camel case condition.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/StructureGenerator.kt
```diff
@@ -148,6 +149,9 @@ class StructureGenerator(
                     val (memberName, memberSymbol) = memberShapeDataContainer.getOrElse(member) { Pair(null, null) }
                     if (memberName == null || memberSymbol == null) continue
                     val terminator = if (index == membersSortedByName.size - 1) "" else ","
+                    if (memberSymbol.properties.containsKey("NeedsDataImport")) {
+                        writer.addImport(FoundationTypes.Data)
+                    }
```
sichanyoo, line 154: When `SwiftSymbolProvider` constructs a symbol for a blob shape and it has a default trait applied to it, we take the default value from the trait (string value) and convert it to `Foundation.Data` type using `Data(stringValue.utf8)` initializer. But `SwiftSymbolProvider` doesn't have access to a writer as it is just a tool for getting corresponding symbol for a Smithy shape. So a flag is set to the symbol for adding this `Foundation.Data` import upstream.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -147,6 +157,34 @@ open class MemberShapeDecodeGenerator(
...
+                is TimestampShape -> " ?? Date(timeIntervalSince1970: 0)"
+                is DocumentShape -> {
+                    val node = requiredTrait.toNode()
+                    resolveDocumentDefault(true, node)
+                }
+                is BlobShape -> { resolveBlobDefault(targetShape) }
+                // No default provided for other types
+                else -> { "" }
+            }
+        }
```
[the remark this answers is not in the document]
dayaffe: Even if an output member has required trait it should still be optional as of now so it shouldn't be necessary to provide defaults right?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -147,6 +157,34 @@ open class MemberShapeDecodeGenerator(
     private fun default(memberShape: MemberShape): String {
         val targetShape = ctx.model.expectShape(memberShape.target)
         val defaultTrait = memberShape.getTrait<DefaultTrait>() ?: targetShape.getTrait<DefaultTrait>()
+        val requiredTrait = memberShape.getTrait<RequiredTrait>()
```
dayaffe, line 160: do we need required trait for default trait? We shouldn't be resolving required trait ourselves I think and leave that to the knowledge index to resolve then we treat accordingly. but that's once we're ready to say if something should be required we're gonna make it non-nullable

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +296,110 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+                 For: boolean, byte, short, integer, long, bigInteger, bigDecimal
+                    just take the literal string value from the trait.
+                 */
+                resolvedDefaultValue = defaultValueLiteral
+            }
+        }
+
+        // Return Symbol.Builder with resolved default value saved to Symbol.Builder's property bag
+        return builder.defaultValue(resolvedDefaultValue)
+    }
```
dayaffe, line 401: Should refactor this to make use of `when` as an expression in kotlin

example:
```
val resolvedDefaultValue = when (targetShape) {
    is ListShape -> "[]"
    is EnumShape -> ".${swiftEnumCaseName(null, defaultValueLiteral)}"
    ... (etc)
}
```

Same applies to the when inside of DocumentationShape. Since this is also a nested when consider moving the logic to handle document shape to a helper function

dayaffe (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
sichanyoo: This here is just for putting zeros or zero-equivalent values to deserialized output members if:
- It's required
- It does not have a default trait
- AND the response from server doesn't have any value for it (wrong behavior from server)

I think we can add this implementation regardless of what we do re: optionality down the line.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
sichanyoo: Refactored as suggested 👍

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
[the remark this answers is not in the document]
sichanyoo: Yes, it shouldn't be necessary, but it makes protocol tests pass and doesn't hurt to do it now.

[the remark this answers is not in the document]
dayaffe: What would happen if theoretically a service gets approved to make a previously required value optional with no default?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -193,4 +233,33 @@ open class MemberShapeDecodeGenerator(
...
+            }
+            node.isBooleanNode -> {
+                val resolvedValue = "false".takeIf { useZeroValue } ?: node.expectBooleanNode().value
+                " ?? Document.boolean($resolvedValue)"
+            }
+            node.isNumberNode -> {
+                val resolvedValue = "0".takeIf { useZeroValue } ?: node.expectNumberNode().value
+                " ?? Document.number($resolvedValue)"
+            }
+            else -> { return "" } // null node type means no default value but explicit
```
dayaffe, line 262: nit: `return` is unnecessary since the return is on 247 outside of when. scope is also unnecessary for single line statements

`else -> ""`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -193,4 +233,33 @@ open class MemberShapeDecodeGenerator(
...
+            " ?? ByteStream.data(Data(\"$value\".utf8))"
+        } else {
+            " ?? Data(\"$value\".utf8)"
+        }
+    }
+
+    private fun resolveDocumentDefault(useZeroValue: Boolean, node: Node): String {
+        return when {
+            node.isObjectNode -> { " ?? Document.object([:])" }
+            node.isArrayNode -> { " ?? Document.array([])" }
```
dayaffe, line 249: nit: can remove scope `{` `}`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -168,6 +205,9 @@ open class MemberShapeDecodeGenerator(
                 is ListShape, is SetShape -> " ?? []"
                 // Maps can only have empty map as default value
                 is MapShape -> " ?? [:]"
+                is TimestampShape -> " ?? Date(timeIntervalSince1970: ${it.expectNumberNode().value})"
+                is DocumentShape -> { resolveDocumentDefault(false, it) }
+                is BlobShape -> { resolveBlobDefault(targetShape, it.toString()) }
```
dayaffe, line 210: nit: only want to put scope if its multi-line, can remove

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -147,6 +156,34 @@ open class MemberShapeDecodeGenerator(
...
+                is FloatShape, is DoubleShape -> " ?? 0.0"
+                is BooleanShape -> " ?? false"
+                is ListShape -> " ?? []"
+                is MapShape -> " ?? [:]"
+                is TimestampShape -> " ?? Date(timeIntervalSince1970: 0)"
+                is DocumentShape -> {
+                    val node = requiredTrait.toNode()
+                    resolveDocumentDefault(true, node)
+                }
+                is BlobShape -> { resolveBlobDefault(targetShape) }
```
dayaffe, line 182: nit: single line should have no scope

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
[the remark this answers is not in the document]
dayaffe: NVM all good, this isnt making anything non-nullable

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
sichanyoo: Change adopted

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
sichanyoo: Change adopted

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
sichanyoo: Change adopted

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
sichanyoo: Change adopted

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+     */
+    private fun handleDefaultValue(shape: Shape, builder: Symbol.Builder): Symbol.Builder {
+        // Skip if the current shape does not have a default trait or if it's a member shape with @clientOptional trait
+        if (!shape.hasTrait<DefaultTrait>() || shape.hasTrait<ClientOptionalTrait>()) {
+            return builder
+        }
+        // Retrieve literal default value as a string from default trait
+        val defaultValueLiteral = shape.getTrait<DefaultTrait>()!!.toNode().toString()
+        // If default value is "null", it is explicit notation for no default value. Return unmodified builder.
+        if (defaultValueLiteral.equals("null")) { return builder }
```
dayaffe, line 321: consider refactoring this to remove `!!`, make use of `?.` and condense 


let me know but I think this is equivalent and removes need for `!!`
```
if (shape.isClientOptionalOrNoDefault()) return builder
val defaultValueLiteral = shape.getTrait<DefaultTrait>()?.toNode()?.toString() ?: return builder
if (defaultValueLiteral == "null") return builder
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+
+        // The current shape may be a member shape or a root level shape.
+        val targetShape = when (shape) {
+            is MemberShape -> {
+                // If containing shape is an input shape, return unmodified builder.
+                if (model.expectShape(shape.container).hasTrait<InputTrait>()) { return builder }
+                model.expectShape(shape.target)
+            }
+            else -> { shape }
+        }
```
dayaffe, line 331: this looks like a good candidate to move to a helper function to improve readability

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+
+        return when (targetShape) {
+            is ListShape -> builder.defaultValue("[]")
+            is EnumShape -> builder.defaultValue(".${swiftEnumCaseName(null, defaultValueLiteral)}")
+            is IntEnumShape -> {
+                // Get the corresponding enum member name (enum case name) for the int value from default trait
+                val enumMemberName = targetShape.enumValues.entries.firstOrNull {
+                    it.value == defaultValueLiteral.toInt()
+                }!!.key
```
dayaffe, line 340: consider changing this to handle nulls and remove `!!`

```
val enumMemberName = targetShape.enumValues.entries.firstOrNull {
    it.value == defaultValueLiteral.toIntOrNull()
}?.key ?: return builder
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+            is BlobShape -> {
+                builder.putProperty("NeedsDataImport", FoundationTypes.Data)
+                builder.defaultValue(
+                    if (targetShape.hasTrait<StreamingTrait>()) {
+                        "ByteStream.data(Data(\"$defaultValueLiteral\".utf8))"
+                    } else {
+                        "Data(\"$defaultValueLiteral\".utf8)"
+                    }
+                )
+            }
```
dayaffe, line 354: up to you but considering moving this to a helper function `handleBlobShape`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
sichanyoo: Adopted suggestion of switching out `!!` with `?.`

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
sichanyoo: Adopted suggestion

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
sichanyoo: It's guaranteed to exist (I assume Smithy model validation checks for it already)

dayaffe (approved):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
sichanyoo: After offline discussion, changed as suggested.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+            node.isNumberNode -> builder.defaultValue("Document.number($literal)")
+            else -> builder
+        }
+    }
+
+    private fun handleBlobDefaultValue(literal: String, shape: Shape, builder: Symbol.Builder): Symbol.Builder {
+        builder.putProperty("NeedsDataImport", FoundationTypes.Data)
+        return builder.defaultValue(
+            if (shape.hasTrait<StreamingTrait>()) {
+                "ByteStream.data(Data(\"$literal\".utf8))"
```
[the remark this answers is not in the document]
jbelkins: Same below, for `Foundation.Data`

jbelkins (changes requested):

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
```diff
@@ -284,6 +297,88 @@ class SwiftSymbolProvider(private val model: Model, val swiftSettings: SwiftSett
...
+    }
+
+    // Document: default value can be set to null, true, false, string, numbers, an empty list, or an empty map.
+    private fun handleDocumentDefaultValue(literal: String, node: Node, builder: Symbol.Builder): Symbol.Builder {
+        return when {
+            node.isObjectNode -> builder.defaultValue("Document.object([:])")
+            node.isArrayNode -> builder.defaultValue("Document.array([])")
+            node.isBooleanNode -> builder.defaultValue("Document.boolean($literal)")
+            node.isStringNode -> builder.defaultValue("Document.string(\"$literal\")")
+            node.isNumberNode -> builder.defaultValue("Document.number($literal)")
```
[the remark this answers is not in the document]
sichanyoo: [Smithy spec on document shape](https://smithy.io/2.0/spec/simple-types.html#document) says `arbitrary precision numbers` and given Swift's double range is e+308 and e-308 in magnitude in both positive and negative direction, I think it's safe to assume it will be a valid Swift number.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -168,6 +205,9 @@ open class MemberShapeDecodeGenerator(
                 is ListShape, is SetShape -> " ?? []"
                 // Maps can only have empty map as default value
                 is MapShape -> " ?? [:]"
+                is TimestampShape -> " ?? Date(timeIntervalSince1970: ${it.expectNumberNode().value})"
```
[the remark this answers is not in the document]
sichanyoo: I don't think so, I need to fix this; it just happened to be that all protocol tests that has default trait applied to timestamp used epoch-seconds format.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSymbolProvider.kt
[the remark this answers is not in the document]
sichanyoo: Fixed using closure that imports dependencies at the time of symbol usage

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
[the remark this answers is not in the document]
sichanyoo: Fixed handling as pointed out. Default value for a timestamp shape can either be an integer for epoch-seconds, or a string for date-time. http-date is not an option for default value of a timestamp shape.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/SymbolExt.kt
```diff
@@ -42,15 +46,36 @@ fun Symbol.isBoxed(): Boolean {
...
     return if (default.isPresent) default.get() else null
 }
```
sichanyoo, line 59: The `defaultValueFromClosure` and `defaultValueClosure` below allow saving a closure that takes a `SwiftWriter` and returns `String?` into Symbol properties. This allows handling dependency imports required for a certain default value. For example, a symbol X may have default value `Data("".utf8)`, which needs `Foundation.Data` import for it to compile. `SwiftSymbolProvider` doesn't have access to a `SwiftWriter` instance, so using this closure allows us to delay adding import until the symbol actually gets used by the writer.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/model/SymbolExt.kt
```diff
@@ -63,6 +86,19 @@ fun Symbol.Builder.boxed(): Symbol.Builder = apply { putProperty(SymbolProperty.
...
 
+/**
+ * Set the closure that gets called with a SwiftWriter to import symbols needed for default value
+ * Used in SwiftSymbolProvider (which doesn't have access to SwiftWriter).
+ * Allows default value of a symbol X returned by SwiftSymbolProvider to have needed imports
+ *      for symbol Y at the time symbol X is printed by SwiftWriter.
+ *
+ * Example: Default value for a symbol called X could be "Data()", which means when the symbol X is printed by SwiftWriter,
+ *      we need to import Foundation.Data.
+ */
```
jbelkins, line 97: I didn't think of making the closure a property on Symbol, but this is great.

jbelkins (approved):

PR #813, opened 2024-09-10
fix: Streamline HttpInterceptor type inference

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1681

## Description of changes
Change construction of HTTP interceptors to match non-HTTP, which is never flagged in compiler crashes.

This is an attempt to address compiler crashes on the automated build system.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Interceptor/HttpInterceptorProvider.swift
+++ b/Sources/ClientRuntime/Interceptor/HttpInterceptorProvider.swift
@@ -5,6 +5,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import class SmithyHTTPAPI.HTTPRequest
+import class SmithyHTTPAPI.HTTPResponse
+
 /// Provides implementations of `HttpInterceptor`.
 ///
 /// For the generic counterpart, see `InterceptorProvider`.
@@ -13,5 +16,5 @@ public protocol HttpInterceptorProvider {
     /// Creates an instance of an `HttpInterceptor` implementation.
     ///
     /// - Returns: The `HttpInterceptor` implementation.
-    func create<InputType, OutputType>() -> any HttpInterceptor<InputType, OutputType>
+    func create<InputType, OutputType>() -> any Interceptor<InputType, OutputType, HTTPRequest, HTTPResponse>
 }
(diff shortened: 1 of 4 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/middleware/MiddlewareExecutionGenerator.kt
```diff
@@ -51,25 +51,11 @@ class MiddlewareExecutionGenerator(
...
-            "}",
-            ClientRuntimeTypes.Core.HttpInterceptorProvider,
-        ) {
-            writer.write(
-                "let i: any \$N<\$N, \$N> = provider.create()",
-                ClientRuntimeTypes.Core.HttpInterceptor,
-                inputShape,
-                outputShape,
-            )
-            writer.write("builder.interceptors.add(i)")
```
jbelkins, line 72: Since the return value of `provider.create()` now matches the type of `interceptors.add()` exactly, there is no need to assign to a temp variable with an explicit type anymore.

Also:
- The writer statements for non-HTTP interceptor providers has been rewritten from a multiline statement.
- The block explicit type added previously in an attempt to address compiler crashes is removed since it seems to have had no effect.

Sources/ClientRuntime/Interceptor/HttpInterceptorProvider.swift
```diff
@@ -13,5 +16,5 @@ public protocol HttpInterceptorProvider {
     /// Creates an instance of an `HttpInterceptor` implementation.
     ///
     /// - Returns: The `HttpInterceptor` implementation.
-    func create<InputType, OutputType>() -> any HttpInterceptor<InputType, OutputType>
+    func create<InputType, OutputType>() -> any Interceptor<InputType, OutputType, HTTPRequest, HTTPResponse>
```
jbelkins, line 19: Have this method return a plain Interceptor, with the return & request types pinned to HTTP.
This exactly matches the type that an Interceptors instance stores.

sichanyoo (approved):

PR #823, opened 2024-09-13
fix: Correct default zero-value for int enum shapes

sichanyoo:
Companion CI PR: https://github.com/awslabs/aws-sdk-swift/pull/1743

## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1741

## Description of changes

- Zero-equivalent value given to int enum shapes for error correction is fixed

Misc.
- Also, updated an outdated doc comment on retry strategy options

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -164,7 +164,8 @@ open class MemberShapeDecodeGenerator(
         // If member is required but there isn't a default value, use zero-equivalents for error correction
         if (requiredTrait != null && defaultTrait == null) {
             return when (targetShape) {
-                is EnumShape, is IntEnumShape -> " ?? .sdkUnknown(\"\")"
+                is EnumShape -> " ?? .sdkUnknown(\"\")"
+                is IntEnumShape -> " ?? .sdkUnknown(-1)"
```
jbelkins, line 168: is there a reason you selected -1 for this and not 0?

jbelkins, line 168: is there a reason you selected -1 for this and not 0?

jbelkins: Also, what if your default value matches an actual enum case?  Would it be better to use `.init(rawValue: -1)` so that if -1 is a case member, it gets created correctly?

sichanyoo: Will use 0, as Kotlin and Java both use 0. Will continue to use unknown case to prevent undefined behavior.

jbelkins (approved):

PR #827, opened 2024-09-18
fix: set CRT log level without fatalError

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1761

## Description of changes
Fixes a crash that happens immediately upon setting the CRT log level.
The cause of the crash is that since `setLogLevel()` is currently an instance function, the logger has to be crated (as part of the shared instance) before it can be 

`SDKDefaultIO.setLogLevel()` is changed from an instance method to a static method; this is technically a breaking change but it has no effect since the current API crashes immediately when used.

## Scope


Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift
```diff
@@ -25,13 +19,24 @@ public final class SDKDefaultIO: @unchecked Sendable {
...
     }
 
     private init() {
         CommonRuntimeKit.initialize()
-        Self.setupLogger(level: .none)
+        do {
+            try Logger.initialize(target: .standardOutput, level: .none)
+        } catch CommonRunTimeError.crtError(let error)
+                    where error.code == 6 && error.name == "AWS_ERROR_UNSUPPORTED_OPERATION" {
+            // logger was already initialized, no need to initialize it
```
jbelkins, line 36: Catch the "logger already initialized" error and ignore it if the logger was already initialized when the log level was set.

Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift
```diff
@@ -6,12 +6,6 @@
 //
 
 import AwsCommonRuntimeKit
-import class Foundation.ProcessInfo
-#if os(Linux)
-import Glibc
-#else
-import Darwin
-#endif
```
jbelkins, line 14: The deleted imports above were unused.

Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift
sichanyoo: Does this mean the log level can't be changed after it's been initialized?

sichanyoo: Discussed offline, this is due to design of CRT logger

dayaffe (approved):

PR #832, opened 2024-09-24
fix: Consider an empty string header value to be present

jbelkins:
## Description of changes
An empty string is an acceptable value for a HTTP header, and SDKs should send an empty string value for a header when one is passed.

Swift implements this behavior correctly, but incorrectly validates "forbidden headers" in a protocol test by considering a HTTP header to "not exist" when its value is an empty string.

This PR fixes the `exists()` method on the `Headers` type so that it returns `true` rather than `false` when the header being checked has an empty string as its value.

Note that the current `NullAndEmptyHeaders` protocol tests (in RestJSON & RestXML) for this behavior are incorrect, and this PR will cause the current version of these tests to fail.  These tests are being corrected in a future version of Smithy and this PR should be merged only when that Smithy version is adopted.

Also:
- Create a new `SmithyHTTPAPITests` test target and move the `Headers` tests into it from `ClientRuntimeTests`.

## Scope


Sources/SmithyHTTPAPI/Headers.swift
```diff
@@ -138,15 +138,7 @@ public struct Headers: Sendable {
...
-        guard headers.index(of: name) != nil else {
-            return false
-        }
-
-        guard let value = value(for: name) else {
-            return false
-        }
-
-        return !value.isEmpty
+        headers.index(of: name) != nil
```
jbelkins, line 141: The previous definition of `exists` was wrong because it considered a header with an empty string for its value to "not exist".  Turns out a header with empty string value is valid, so it should definitely "exist".

I suspect that the reason for this logic is that the previous developers changed the definition of "exists" to suit the behavior demanded by empty-header protocol tests, which is wrong and is being fixed by Smithy team.

Package.swift
```diff
@@ -280,5 +280,9 @@ let package = Package(
             name: "SmithyRetriesTests",
             dependencies: ["ClientRuntime", "SmithyRetriesAPI", "SmithyRetries", "SmithyTestUtil"]
         ),
+        .testTarget(
+            name: "SmithyHTTPAPITests",
+            dependencies: ["SmithyHTTPAPI"]
+        ),
```
jbelkins, line 286: Added a new test target & moved the `Headers` tests into it, since `Headers` is in the `SmithyHTTPAPI` module.

Tests/SmithyHTTPAPITests/HeadersTests.swift
```diff
@@ -123,4 +124,21 @@ class HttpHeadersTests: XCTestCase {
...
+
+    func test_exists_falseIfHeaderDoesntExist() {
+        let subject = Headers(["a": "abc"])
+        XCTAssertFalse(subject.exists(name: "b"))
+    }
+
+    func test_exists_trueIfHeaderValueIsEmptyString() {
+        let subject = Headers(["a": ""])
+        XCTAssertTrue(subject.exists(name: "a"))
+    }
```
jbelkins, line 143: These new tests check the expected behavior for the `exists()` method on `Headers`.

jbelkins:
This PR is failing the downstream CI builds because the `NullAndEmptyHeaders` tests for RestJSON & RestXML are failing - this is expected because those tests have incorrect expectations for empty-string headers.

This PR should merge when this PR ships in a new version of Smithy:
https://github.com/smithy-lang/smithy/pull/2403

jbelkins:
Closing this PR because this change is being made in https://github.com/smithy-lang/smithy-swift/pull/843 (update to Smithy 1.52) instead.

PR #840, opened 2024-10-16
fix: Don't supply a default value when reading a non-optional value

jbelkins:
## Issue
https://github.com/awslabs/aws-sdk-swift/issues/1769

## Description of changes
Currently, expressions deserialized with `read()` (i.e. unlike `readIfPresent()` the value is non-optional, these are primarily union members) are deserialized with an unnecessary default value, when the model provides a default.

This PR omits a default value when deserializing with `read()`, avoiding a compile warning.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -290,4 +293,6 @@ open class MemberShapeDecodeGenerator(
             )
         }
     }
+
+    private var readingRequiredValue: Boolean = shapeContainingMembers.isUnionShape
```
jbelkins, line 297: Extracted `readingRequiredValue` to a helper.  In the future, if nonoptional values are supported, this logic may become more complex.

sichanyoo (approved):

PR #847, opened 2024-10-28
chore: Sort dependencies before rendering into package manifest

jbelkins:
## Description of changes
Sort the package dependencies & target dependencies by package name prior to rendering a `Package.swift`.
This will cause some "churn" in the manifests now, but will result in less churn later since manifest contents will be independent of the order dependencies are added.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PackageManifestGenerator.kt
```diff
@@ -34,16 +34,20 @@ class PackageManifestGenerator(val ctx: ProtocolGenerator.GenerationContext) {
...
-                val dependenciesByURL = externalDependencies.distinctBy {
-                    it.getProperty("url", String::class.java).getOrNull()
-                        ?: "${it.getProperty("scope", String::class.java).get()}.${it.packageName}"
-                }
+                val dependenciesByURL = externalDependencies
+                    .distinctBy {
+                        it.getProperty("url", String::class.java).getOrNull()
+                            ?: "${it.getProperty("scope", String::class.java).get()}.${it.packageName}"
+                    }
+                    .sorted()
```
jbelkins, line 42: Both `dependenciesByURL` and `dependenciesByTarget` are `List<SymbolDependency>`.

The `SymbolDependency` `compareTo()` implementation sorts by `packageName` when the `dependencyType` is the same between dependencies.  (Swift doesn't ever set `dependencyType`, and it sets `packageName` to match the Swift published library name.)
https://github.com/smithy-lang/smithy/blob/main/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/SymbolDependency.java#L243

So, these lists get sorted by package name, which in our case is the Swift library name.

sichanyoo (approved):

PR #852, opened 2024-11-08
fix: Retry handling for streamed request bodies

jbelkins:
## Description of changes
This change ensures that retry is only used when a request body is reusable, and that the request body is prepared for reuse if needed.
- When a request body is a seekable stream, rewind the stream to the beginning before retrying.  Do not retry if rewind is not successful.
- When a request body is a non-seekable stream, do not attempt to retry; fail the request immediately instead.
- Non-streamed request bodies (i.e. in-memory Data and empty bodies) are always retryable.

Additionally, unit tests are added to `Orchestrator` to verify correct retry behavior.

## Scope

### DIFF
--- a/Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
+++ b/Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
@@ -15,6 +15,7 @@ import SmithyRetriesAPI
 import SmithyRetries
 @_spi(SmithyReadWrite) import SmithyJSON
 @_spi(SmithyReadWrite) import SmithyReadWrite
+import SmithyStreams
 
 class OrchestratorTests: XCTestCase {
     struct TestInput {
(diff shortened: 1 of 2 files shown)


Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
```diff
@@ -167,20 +168,22 @@ class OrchestratorTests: XCTestCase {
     }
 
     class TraceExecuteRequest: ExecuteRequest {
```
jbelkins, line 170: The behavior of this type is unchanged; I added a variable to allow for counting of retry attempts that is separate from the original allowed count.

Tests/ClientRuntimeTests/OrchestratorTests/OrchestratorTests.swift
```diff
@@ -1315,4 +1318,43 @@ class OrchestratorTests: XCTestCase {
             }
         }
     }
+
```
jbelkins, line 1321: The `RetryBackoffStrategy` below always sets a retry delay of 0 so the retry tests can proceed without waiting for any retry delay.  This prevents the first test below from taking 0.5-2 seconds to finish.

dayaffe (approved):
LGTM

PR #853, opened 2024-11-13
fix: Make Context thread-safe

jbelkins:
## Description of changes
In testing with the Thread Sanitizer enabled, simultaneous mutating accesses to `Context` have been observed, related to the maintenance of clock skew calculations.  Crashes have been observed during integration tests which are suspected to be caused by this or other data race issues.

To address:
- Conceal the `attributes` property on `Context` since it exposes a non-thread safe value type & cannot be simultaneously get/set.  This is technically a breaking API change, but it is highly unlikely that any customer interfaces with it.
- All access to individual attributes is now performed via get/set methods on `Attributes` which now use a lock to enforce exclusive access.

## Scope

### DIFF
--- a/Sources/Smithy/Context.swift
+++ b/Sources/Smithy/Context.swift
@@ -5,46 +5,61 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import class Foundation.NSRecursiveLock
+
 public class Context {
-    public var attributes: Attributes
+    private var _attributes: Attributes
+    private var lock = NSRecursiveLock()
 
     public init(attributes: Attributes) {
-        self.attributes = attributes
+        self._attributes = attributes
     }
 
     public func toBuilder() -> ContextBuilder {
-        let builder = ContextBuilder()
-        builder.attributes = self.attributes
-        return builder
+        ContextBuilder(attributes: accessAttributes())
     }
 
     public func getLogger() -> LogAgent? {
-        return attributes.get(key: AttributeKeys.logger)
+        return accessAttributes().get(key: AttributeKeys.logger)
+    }
+
+    @discardableResult private func accessAttributes(accessor: ((inout Attributes) -> Void)? = nil) -> Attributes {
+        lock.lock()
+        defer { lock.unlock() }
+        accessor?(&_attributes)
+        return _attributes
     }
 }
 
 extension Context {
     public func get<T>(key: AttributeKey<T>) -> T? {
-        self.attributes.get(key: key)
+        accessAttributes().get(key: key)
     }
 
     public func contains<T>(key: AttributeKey<T>) -> Bool {
-        self.attributes.contains(key: key)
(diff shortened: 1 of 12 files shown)


Sources/Smithy/Context.swift
```diff
@@ -5,46 +5,63 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import class Foundation.NSRecursiveLock
+
 public class Context {
-    public var attributes: Attributes
+    private var _attributes: Attributes
+    private let lock = NSRecursiveLock()
```
jbelkins, line 12: `attributes` has been made private, and the lock enforces exclusive access to it.

Sources/Smithy/Context.swift
```diff
@@ -5,46 +5,63 @@
...
+    }
+
+    @discardableResult
+    private func accessAttributes(accessor: ((inout Attributes) -> Void)? = nil) -> Attributes {
+        lock.lock()
+        defer { lock.unlock() }
+        accessor?(&_attributes)
+        return _attributes
     }
 }
```
jbelkins, line 33: The `accessAttributes` function above enforces exclusive access to the now-private `_attributes` property.

Sources/Smithy/Context.swift
```diff
@@ -5,46 +5,63 @@
...
-        return attributes.get(key: AttributeKeys.logger)
+        return accessAttributes().get(key: AttributeKeys.logger)
+    }
+
+    @discardableResult
+    private func accessAttributes(accessor: ((inout Attributes) -> Void)? = nil) -> Attributes {
+        lock.lock()
+        defer { lock.unlock() }
+        accessor?(&_attributes)
+        return _attributes
```
dayaffe, line 31: Question: I am not sure if this is fully thread safe. If I am reading this correctly it looks like the thing that happens within scope of the lock is making a copy of _attributes. Then when get is called via accessAttributes().get(key: key) accessAttributes() creates a copy of _attributes under a lock, the lock is released, and then .get(key: key) is called without any lock. I think to make this thread safe we need the operations themselves to be performed under a lock -- lmk what you think

jbelkins: That operation (i.e. what happens in `getLogger()`) is safe because Attributes is a value type, i.e. it is copied when a new reference to it is created.  The copy of the attributes that is returned by `accessAttributes()` is not referenced anywhere else, so it is safe to access the `logger` (or any other attribute).

You would not want to modify that copy of attributes to modify the context, because you're modifying a copy, the change will not be reflected in the Context the attributes were copied from.

jbelkins: Honestly that getLogger() method can be simplified down to just `get(AttributeKeys.logger)`.  Same effect, same thread safety, simpler code.

dayaffe (approved):
Discussed offline about thread safety

PR #858, opened 2024-11-15
Avoid re-parsing #file values during logging

harlanhaskins:
This fixes a CPU usage issue with frequent debug logging.

## Description of changes

I noticed when tracing a large file upload through S3Client that a lot of time was spent inside `LogAgent.currentModule(filePath:)`. I checked and noticed that it's only called from within this file, and only ever used with the static value of `#file` (no callers ever change the parameter).

As such, we can hoist it into a file-private variable and avoid recomputing it frequently. I would like to keep it as a `static let` inside `LogAgent`, but that's not supported on protocols.
Screenshot 2024-11-15 at 1 18 52 PM

## Scope

### DIFF
--- a/Sources/Smithy/Logging/LogAgent.swift
+++ b/Sources/Smithy/Logging/LogAgent.swift
@@ -38,24 +38,26 @@ public enum LogAgentLevel: String, Codable, CaseIterable {
     case fatal
 }
 
+private let currentModule: String = {
+    let filePath = #file
+    let utf8All = filePath.utf8
+    return filePath.utf8.lastIndex(of: UInt8(ascii: "/")).flatMap { lastSlash -> Substring? in
+        utf8All[..<lastSlash].lastIndex(of: UInt8(ascii: "/")).map { secondLastSlash -> Substring in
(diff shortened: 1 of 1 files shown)


jbelkins:
@harlanhaskins Please have a look at #862 which I plan to merge in place of this PR.

In reviewing your proposed change, I noticed that the logging code you are changing above is not behaving correctly in the first place.  The module name passed into the swift-log `LogHandler` is supposed to be the module name at the logging call site, not the module of the logger itself.

So, the bad news is that I can't store the module name in a variable as you have suggested, since it will be different for every invocation of the `log` function.

The good news is that Swift now provides the `#fileID` macro for identifying the current Swift source file.  `#fileID` always has the module name right up front followed by a `/`, so it should be less CPU-intensive to parse than from `#file`.

I'll let you know here when #862 ships in a new SDK version, and you can re-evaluate your CPU usage situation.  If it's still high, we can look for additional ways to optimize.

harlanhaskins:
I had wondered if that was the intended behavior, and yeah that's unfortunate that we can't pre-cache these file requests. But it's understandable.

PR #862, opened 2024-11-19
fix: Correct module determination when logging

jbelkins:
## Issue \#
https://github.com/smithy-lang/smithy-swift/pull/858 (customer-proposed PR)

## Description of changes
- Fix the module name used when logging so that it is the module of the caller, not the logger itself
- Derive the module name from `#fileID` instead of `#file` which is:
  - more secure (does not include paths from the host system's filesystem)
  - easier to derive module name from it (should reduce CPU usage)
  - matches [modern Swift API guidelines](https://www.swift.org/documentation/api-design-guidelines/#parameter-names)
- Add a test for `SwiftLogger` to ensure that when it logs, it passes the right data into the swift-log `LogHandler`

also: A little code cleanup

## Scope


Package.swift
```diff
@@ -227,7 +227,12 @@ let package = Package(
         ),
         .testTarget(
             name: "ClientRuntimeTests",
-            dependencies: ["ClientRuntime", "SmithyTestUtil", "SmithyStreams"],
+            dependencies: [
+                "ClientRuntime",
+                "SmithyTestUtil",
+                "SmithyStreams",
+                .product(name: "Logging", package: "swift-log"),
+            ],
```
jbelkins, line 235: New logging test requires direct use of `swift-log` components, hence the test target dependency on `swift-log`

Sources/Smithy/Logging/LogAgent.swift
```diff
@@ -39,80 +41,85 @@ public enum LogAgentLevel: String, Codable, CaseIterable {
...
-    }
 
     /// Log a message passing with the `.info` log level.
-    func info(_ message: String, file: String = #file, function: String = #function, line: UInt = #line) {
+    func info(_ message: String, file: String = #fileID, function: String = #function, line: UInt = #line) {
         self.log(level: .info,
                  message: message,
                  metadata: nil,
-                 source: Self.currentModule(),
+                 source: currentModule(fileID: file),
```
jbelkins, line 50: All these log-level-specific functions are changed similarly.  We weren't passing the caller's source file into the function that parses the module, hence every time we logged, it showed that it originated from the logger's module, not the caller's.

jbelkins: Also, `#fileID` is the proper path to use in modern Swift production code, not `#file`.

`#fileID` omits path components from the host machine, and makes the module easier to parse.

See https://www.swift.org/documentation/api-design-guidelines/#parameter-names .

Sources/Smithy/Logging/LogAgent.swift
```diff
@@ -39,80 +41,85 @@ public enum LogAgentLevel: String, Codable, CaseIterable {
...
+/// - Parameter fileID: The value of the `#fileID` macro at the point of logging.
+/// - Returns: The name of the module, as parsed from the passed `#fileID`.
+private func currentModule(fileID: String) -> String {
+    let utf8All = fileID.utf8
+    if let slashIndex = utf8All.firstIndex(of: UInt8(ascii: "/")) {
+        return String(fileID[..<slashIndex])
+    } else {
+        return "n/a"
+    }
+}
```
jbelkins, line 125: Compare this `currentModule()` function to the one deleted above; this one is significantly simpler because `#fileID` always has the module name right out front, followed by a `/`.

Tests/ClientRuntimeTests/LoggingTests/SwiftLoggerTests.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import Smithy
+import ClientRuntime
+import Logging
+
+final class SwiftLoggerTests: XCTestCase {
```
jbelkins, line 13: This test creates a `SwiftLogger` with a mock `LogHandler` injected (`LogHandler` is the entry point to `swift-log`; the mock is defined below the tests.)

A log statement is executed at every log level, and it is verified that the correct log params are received by the `LogHandler`.

Sources/Smithy/Logging/LogAgent.swift
jbelkins: Also I should point out this is exactly how `swift-log` makes this determination for itself:
https://github.com/apple/swift-log/blob/main/Sources/Logging/Logging.swift#L1523

harlanhaskins:
This looks like a significant simplification, thank you!

PR #865, opened 2024-11-20
fix: Change logger params to @autoclosures

jbelkins:
## Description of changes
The `message`, `metadata`, and `source` params to the `LogAgent` protocol have been changed to Swift autoclosures.
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/closures/#Autoclosures

This defers evaluation of these params until they are actually used, if they are used at all.  If the params are never evaluated (as happens when a log statement is at a log level below the logger's level), needless consumption of CPU is prevented.

The `@autoclosure`s added in this PR match up with those used on the `swift-log` module's `LogHandler` protocol, so that we get the desired performance benefit when logging to any `swift-log`-compatible logger.

Note: although this replaces type `T` with `@autoclosure () -> T` in public function params, this is not a breaking change because the Swift compiler makes any function param `T` into a closure automatically (hence the name "autoclosure".)

## Scope


Sources/Smithy/Logging/LogAgent.swift
```diff
@@ -23,9 +23,9 @@ public protocol LogAgent {
     ///     - function: The function the log line was emitted from.
     ///     - line: The line the log message was emitted from.
     func log(level: LogAgentLevel,
-             message: String,
-             metadata: [String: String]?,
-             source: String,
+             message: @autoclosure () -> String,
+             metadata: @autoclosure () -> [String: String]?,
+             source: @autoclosure () -> String,
```
jbelkins, line 28: `message`, `metadata`, and `source` are now marked `@autoclosure`.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/closures/#Autoclosures

This prevents the parameter from being evaluated unless it is actually used by the function being called.  In the context of logging, this means that these params are not constructed when a log message is not going to be generated (i.e. when the log level of a statement is below the logger's threshold.)

This corresponds with `swift-log`'s `LogHandler` protocol, where these three params are also autoclosures.

(Note that `#file`, `#function`, and `#line` evaluate to static strings, so there is no real benefit to making them autoclosures.)

Tests/ClientRuntimeTests/NetworkingTests/URLSession/FoundationStreamBridgeTests.swift
```diff
@@ -117,8 +117,8 @@ private class TestLogger: LogAgent {
         self.level = level
     }
 
-    func log(level: LogAgentLevel = .info, message: String, metadata: [String : String]? = nil, source: String = "ChecksumUnitTests", file: String = #file, function: String = #function, line: UInt = #line) {
-        messages.append((level: level, message: message))
+    func log(level: LogAgentLevel = .info, message: @autoclosure () -> String, metadata: @autoclosure () -> [String : String]? = nil, source: @autoclosure () -> String = "ChecksumUnitTests", file: String = #file, function: String = #function, line: UInt = #line) {
+        messages.append((level: level, message: message()))
```
jbelkins, line 121: Here & below, test-only `LogAgent` implementations have been updated to the new protocol requirements.

Sources/Smithy/Logging/SwiftLog+LogAgent.swift
```diff
@@ -1,68 +1,64 @@
...
+    ) {
+        self.logger.log(
+            level: level.toLoggerLevel(),
+            Logger.Message(stringLiteral: message()),
+            metadata: metadata()?.mapValues { Logger.MetadataValue.string($0) },
+            source: source(),
+            file: file as String,
+            function: function as String,
+            line: line
+        )
```
jbelkins, line 56: The `log` function is updated with `@autoclosure`s.

sichanyoo (approved):

PR #884, opened 2024-12-12
chore: Run repo-sync workflow only in source repo

sichanyoo:
## Issue \#

1934

## Description of changes

- Adds `if` condition checks on repo-sync workflow and others so it only runs in source repo when triggered by commit to main / manually, or when it's triggered by pull request (regardless of public vs private repo)

## Scope


.github/workflows/codegen.yml
```diff
@@ -1,8 +1,6 @@
 name: Codegen
 
 on:
-  push:
-    branches: [ main ]
```
jbelkins, line 5: I don't want to remove CI from the main branch.  Instead let's figure out how to have it trigger only on the public repo.

jbelkins (changes requested):

.github/workflows/codegen.yml
sichanyoo: Reverted change after discussion, now uses `if` checks instead to filter running on commit to main in private repo.

jbelkins (approved):

PR #887, opened 2024-12-24
feat: Support RPCv2 CBOR wire protocol

dayaffe:
## Scope


Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,237 @@
...
+    public var hasContent: Bool { cborValue != nil && cborValue != .null }
+
+    public init(nodeInfo: NodeInfo, cborValue: CBORType?, parent: Reader? = nil) {
+        self.nodeInfo = nodeInfo
+        self.cborValue = cborValue
+        self.parent = parent
+        self.children = Self.children(from: cborValue, parent: self)
+    }
+
+    public static func from(data: Data) throws -> Reader {
```
dayaffe, line 36: this function calls CBORDecoder which is defined in CRT for actually decoding individual types

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,237 @@
...
+        let rootValue: CBORType
+        if decoder.hasNext() {
+            rootValue = try decoder.popNext()
+        } else {
+            rootValue = .null
+        }
+        return Reader(nodeInfo: "", cborValue: rootValue, parent: nil)
+    }
+
+    private static func children(from cborValue: CBORType?, parent: Reader) -> [Reader] {
```
dayaffe, line 47: This function helps form a graph for nested and complex types

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    public func write<T>(_ value: T?) throws where T: RawRepresentable, T.RawValue == String {
+        if let rawValue = value?.rawValue {
+            try write(rawValue)
+        }
+    }
+
+    public func writeMap<T>(
```
dayaffe, line 152: map and array encoding is particularly complex due to the possibility of nested types. A map can be within an array and vice versa

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+    ) throws {
+        guard let value else { return }
+        var map: [String: CBORType] = [:]
+
+        for (key, val) in value {
+            let writer = self[key]
+            try valueWritingClosure(val, writer)
+
+            // If the writer itself doesn't have a cborValue, build it from its children
+            if writer.cborValue == nil, !writer.children.isEmpty {
```
dayaffe, line 167: its possible cborValue wasnt yet assigned, in that case let's move on and parse the children first then come back

Sources/SmithyTestUtil/CBORComparator.swift
```diff
@@ -0,0 +1,129 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+import AwsCommonRuntimeKit
+import Foundation
+@_spi(SmithyReadWrite) import SmithyCBOR
+
+public struct CBORComparator {
```
dayaffe, line 11: this is solely used for testing. we need to compare the semantic values rather than the exact binary values. CRT should have sufficiently tested their de-serializer against SEP prescribed tests to ensure that using the Decoder in protocol test comparison is ok

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HTTPBindingProtocolGenerator.kt
```diff
@@ -415,16 +415,12 @@ abstract class HTTPBindingProtocolGenerator(
...
-             * This is because endpoint middleware must be generated into client operation stack first with position: .before,
-             * AFTER which auth scheme must be generated into client operation stack with position: .before, which ensures
-             * auth scheme middleware is executed FIRST before endpoint middleware is executed, as SRA flow defines.
-             */
             operationMiddleware.appendMiddleware(operation, AuthSchemeMiddleware(ctx.model, ctx.symbolProvider))
             for (integration in ctx.integrations) {
                 integration.customizeMiddleware(ctx, operation, operationMiddleware)
             }
+            // must be last to support adding business metrics
+            addUserAgentMiddleware(ctx, operation)
```
dayaffe, line 423: bug fix: addUserAgentMiddleware handles feature id (business metric) tracking. This was supposed to be last in the stack but it never was prior to this change due to protocol-specific middleware

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestErrorGenerator.kt
```diff
@@ -76,7 +83,25 @@ open class HttpProtocolUnitTestErrorGenerator protected constructor(builder: Bui
...
-            val data = writer.format("Data(\"\"\"\n\$L\n\"\"\".utf8)", test.body.get().replace("\\\"", "\\\\\""))
+            val isCbor = ctx.service.requestWireProtocol == WireProtocol.CBOR ||
+                ctx.service.awsProtocol == AWSProtocol.RPCV2_CBOR ||
+                ctx.service.responseWireProtocol == WireProtocol.CBOR ||
+                test.protocol == Rpcv2CborTrait.ID
+
+            val bodyContent = test.body.get().replace("\\\"", "\\\\\"")
+            val data: String = if (isCbor) {
+                // Attempt to decode Base64 data once for CBOR
+                try {
```
dayaffe, line 94: CBOR data is binary data but for protocol tests it need to be passed as a base64 encoded string

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,237 @@
...
+        case .map(let map):
+            for (key, value) in map {
+                let child = Reader(nodeInfo: key, cborValue: value, parent: parent)
+                children.append(child)
+            }
+        case .array(let array):
+            for (index, value) in array.enumerated() {
+                let child = Reader(nodeInfo: "\(index)", cborValue: value, parent: parent)
+                children.append(child)
+            }
```
sichanyoo, line 59: Q:How are indefinite map and indefinite array CBOR values handled?

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,237 @@
...
+        case .bytes(let data): return data
+        case .text(let string): return Data(base64Encoded: string)
+        default: return nil
+        }
+    }
+
+    public func readIfPresent() throws -> Document? {
+        guard let cborValue else { return nil }
+        return Document(cborValue as! SmithyDocument)
+    }
```
sichanyoo, line 154: Q: RPCv2 CBOR SEP says Smithy's document type isn't yet supported in this protocol; should this just return nil?

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+
+    public func write(_ value: Int16?) throws {
+        guard let value else { return }
+        self.cborValue = .int(Int64(value))
+    }
+
+    public func write(_ value: UInt8?) throws {
+        guard let value else { return }
+        self.cborValue = .uint(UInt64(value))
+    }
```
sichanyoo, line 118: Should the integer types be written using "smallest" type that can fit the value, instead of Int64 / UInt64 here? I.e., SEP says `For example, a field modeled as a long with a value of 1 should be sent in a single byte.` so checking int size might be necessary to use minimum number of bits here

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+
+    public func write(_ value: String?) throws {
+        guard let value else { return }
+        self.cborValue = .text(value)
+    }
+
+    public func write(_ value: Float?) throws {
+        guard let value else { return }
+        self.cborValue = .double(Double(value))
+    }
```
sichanyoo, line 93: Why is Float written using "bigger" type Double? Should this be changed?

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+
+            // If the writer itself doesn't have a cborValue, build it from its children
+            if writer.cborValue == nil, !writer.children.isEmpty {
+                var childMap: [String: CBORType] = [:]
+                for child in writer.children {
+                    if let childCborValue = child.cborValue {
+                        childMap[child.nodeInfo] = childCborValue
+                    }
+                }
+                writer.cborValue = .map(childMap) // Construct the map for the writer
```
sichanyoo, line 174: Q1: What happens if any of the children's cborValue wasn't yet assigned either? Or is that not possible?
Q2: What if a child map contains another nested child map? Doesn't this only handle one level of nesting?

Sources/SmithyCBOR/Writer/Writer.swift
```diff
@@ -0,0 +1,224 @@
...
+                array.append(cborValue)
+            } else if !childWriter.children.isEmpty {
+                // If no cborValue but has children, create a map from its children
+                var childMap: [String: CBORType] = [:]
+                for child in childWriter.children {
+                    if let childCborValue = child.cborValue {
+                        childMap[child.nodeInfo] = childCborValue
+                    }
+                }
+                array.append(.map(childMap)) // Append the constructed map
```
sichanyoo, line 213: Same questions as ones for map writing above, except for list instead of map.

Sources/SmithyReadWrite/SmithyReader.swift
```diff
@@ -208,4 +208,5 @@ public extension SmithyReader {
 
 public enum ReaderError: Error {
     case requiredValueNotPresent
+    case invalidBase64
```
sichanyoo, line 211: Q: Where is this used?

Sources/SmithyTestUtil/CBORComparator.swift
sichanyoo: Is the testing in CRT side done already? If not, what's the known timeline?

sichanyoo (commented):
Mostly Qs

Sources/SmithyCBOR/Reader/Reader.swift
dayaffe: oh yeah will change, in write its just empty

Sources/SmithyCBOR/Writer/Writer.swift
dayaffe: Float is a SmithyWriter type but its not a supported CBOR value type. So we convert it to bigger type Double in order to not lose precision once we're in CBOR territory

Sources/SmithyCBOR/Writer/Writer.swift
dayaffe: Int64/UInt64 is what CRT supports so here we're writing a SmithyWriter supported swift value to a value that's ready to be CBOR encoded by CRT and then sent. CRT handles using the "smallest" type to fit the value, we handle moving a Swift type to a CBOR supported type. These write functions are not what gets sent. 

https://github.com/awslabs/aws-crt-swift/blob/697efa7317bc7110fd9b249722d7fcb37092e839/Source/AwsCommonRuntimeKit/crt/CBOR.swift#L9

Sources/SmithyCBOR/Reader/Reader.swift
dayaffe: I added handling for them to CRT Decoder and CRT Encoder in this PR: https://github.com/awslabs/aws-crt-swift/pull/305

Sources/SmithyCBOR/Writer/Writer.swift
dayaffe: **Q1:** It is possible for a child’s cborValue to remain unassigned if you never write a value to it. However, that’s okay: if a child has children of its own, writeMap may build a .map(...) from those children, or the final encode(encoder:) pass will encode it as an indefinite-length map (see code block starting line 48). If the child has no children and no cborValue, it just gets skipped or encoded as an empty map.

**Q2:** Multiple nesting levels are supported. Even though writeMap appears to build just a one-level map, it considers each child to be a writer itself which allows the real nesting to occur during encode(encoder:), which recurses down the entire tree of child writers. This means a child map can contain another map, which can contain another map, and so on—no matter how deeply nested.

Sources/SmithyCBOR/Writer/Writer.swift
dayaffe: let me know if I answered above -- in this case see: ```// Create a child writer for each list element
            let childWriter = Writer(nodeInfo: memberNodeInfo, parent: self)```

Sources/SmithyReadWrite/SmithyReader.swift
dayaffe: will delete

sichanyoo (approved):
LGTM; should be good to merge after Josh's approval as well

Sources/SmithyTestUtil/CBORComparator.swift
```diff
@@ -0,0 +1,129 @@
...
+    /// Returns true if the logical CBOR values represented by the `Reader` instances are equal.
+    /// - Parameters:
+    ///   - dataA: The first CBOR data object to compare.
+    ///   - dataB: The second CBOR data object to compare.
+    /// - Returns: Returns true if the CBOR documents are equal.
+    public static func cborData(_ dataA: Data, isEqualTo dataB: Data) throws -> Bool {
+        let readerA = try Reader.from(data: dataA)
+        let readerB = try Reader.from(data: dataB)
+        return compareReaders(readerA, readerB)
+    }
```
jbelkins, line 21: This type looks to me like it does the right things, but a few unit tests around it to verify basic cases around lists, maps, etc. would be useful.

Sources/SmithyTestUtil/XCTAssertions.swift
```diff
@@ -30,6 +31,24 @@ public func XCTAssertJSONDataEqual(
     }
 }
 
+public func XCTAssertCBORDataEqual(
+    _ expression1: @autoclosure () throws -> Data,
+    _ expression2: @autoclosure () throws -> Data,
+    _ message: @autoclosure () -> String = "",
+    file: StaticString = #filePath,
```
jbelkins, line 38: nit: Modern Swift recommendations are to not bother with `StaticString` and just use plain ol' `String` instead.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSettings.kt
```diff
@@ -29,6 +36,17 @@ private const val SWIFT_VERSION = "swiftVersion"
...
+private val DEFAULT_PROTOCOL_RESOLUTION_PRIORITY = setOf<ShapeId>(
+    Rpcv2CborTrait.ID,
+    AwsJson1_0Trait.ID,
+    AwsJson1_1Trait.ID,
+    RestJson1Trait.ID,
+    RestXmlTrait.ID,
+    AwsQueryTrait.ID,
+    Ec2QueryTrait.ID,
+)
+
```
jbelkins, line 49: Is this priority defined in some SEP?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HTTPBindingProtocolGenerator.kt
jbelkins: This explains why the user-agent interceptor moved in the service client diffs, I guess

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/protocols/core/StaticHttpBindingResolver.kt
```diff
@@ -28,6 +28,8 @@ open class StaticHttpBindingResolver(
     private val httpTrait: HttpTrait,
     private val defaultContentType: String
 ) : HttpBindingResolver {
+    protected fun getServiceName(): String = context.service.id.name
```
jbelkins, line 31: Any reason not to make this a computed `val` instead of a function?

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/readwrite/ServiceUtils.kt
```diff
@@ -7,16 +7,18 @@ import software.amazon.smithy.aws.traits.protocols.Ec2QueryTrait
 import software.amazon.smithy.aws.traits.protocols.RestJson1Trait
 import software.amazon.smithy.aws.traits.protocols.RestXmlTrait
 import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.protocol.traits.Rpcv2CborTrait
 import software.amazon.smithy.swift.codegen.model.hasTrait
 
-// An enum expressing the six defined AWS protocols used with Smithy.
+// An enum expressing the seven defined AWS protocols used with Smithy.
```
jbelkins, line 13: lol nice catch

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/swiftmodules/SmithyReadWriteTypes.kt
```diff
@@ -13,7 +13,8 @@ import software.amazon.smithy.swift.codegen.SwiftDependency
 object SmithyReadWriteTypes {
     val SmithyReader = runtimeSymbol("SmithyReader", SwiftDeclaration.PROTOCOL)
     val SmithyWriter = runtimeSymbol("SmithyWriter", SwiftDeclaration.PROTOCOL)
-    val ReaderError = runtimeSymbol("ReaderError", SwiftDeclaration.ENUM, emptyList())
+    val ReaderError = runtimeSymbol("ReaderError", SwiftDeclaration.ENUM)
```
jbelkins, line 16: The change above causes `ReaderError` to be imported with `@_spi(SmithyReadWrite)` even though it is a fully public type.  Please revert.

jbelkins (changes requested):
Some minor requested changes in the comments.  Overall this is excellent, CBOR falls pretty nicely into the frameworks we already have for JSON, XML, etc.

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,241 @@
...
+        switch cborValue {
+        case .bytes(let data): return data
+        case .text(let string): return Data(base64Encoded: string)
+        default: return nil
+        }
+    }
+
+    public func readIfPresent() throws -> Document? {
+        // No operation.  Smithy document not supported in CBOR
+        return nil
```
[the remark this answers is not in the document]
dayaffe: I believe in RPCv3 so not anytime in the near future. Not supporting Document type is part of the SEP https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/rpcv2-cbor/protocol-selection_smithy-rpc-v2.md

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSettings.kt
dayaffe: Yeah in section "This priority ordered list SHOULD be the following, with protocols that a specific SDK does not support removed" https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/rpcv2-cbor/protocol-selection_smithy-rpc-v2.md

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/swiftmodules/SmithyReadWriteTypes.kt
dayaffe: I presume I should do the same for WriterError -- will revert and change

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,241 @@
...
+        memberReadingClosure: (Reader) throws -> Member,
+        memberNodeInfo: NodeInfo,
+        isFlattened: Bool
+    ) throws -> [Member]? {
+        guard let cborValue else { return nil }
+        guard case .array = cborValue else { return nil }
+
+        return try children.map { child in
+            // Check if the child is an indefinite-length text
+            if case .indef_text_start = child.cborValue {
```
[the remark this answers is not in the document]
dayaffe: Good catch, I'll check & test but I think I can safely remove this now that indef_text handling is moved to CRT side.

Sources/SmithyTestUtil/XCTAssertions.swift
dayaffe: Oh that's in every DataEqual function for every protocol -- example:
```
public func XCTAssertXMLDataEqual(
    _ expression1: @autoclosure () throws -> Data,
    _ expression2: @autoclosure () throws -> Data,
    _ message: @autoclosure () -> String = "",
    file: StaticString = #filePath,
    line: UInt = #line
)
```

dayaffe: Tried to change to String, turns out XCT uses StaticString and `file` is passed to XCT. We have ~30 references to StaticString in our repo for this reason

Sources/SmithyCBOR/Reader/Reader.swift
```diff
@@ -0,0 +1,241 @@
...
+                    throw ReaderError.invalidType("Expected Member type, but got \(type(of: combinedText))")
+                }
+            } else {
+                // Handle regular values
+                return try memberReadingClosure(child)
+            }
+        }
+    }
+
+    public func readNullIfPresent() throws -> Bool? {
```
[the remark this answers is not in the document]
dayaffe: Addressed in latest revision

jbelkins (approved):

PR #901, opened 2025-02-05
fix!: Remove mutating 'addInterceptor' methods from client config

jbelkins:
## Description of changes
Removes the `addInterceptorProvider` methods from client config for both interceptors & HTTP interceptors.

Interceptors may be added by using the `interceptorProvider` and `httpInterceptorProvider` params on the config initializer.

This is technically a breaking change because it removes a means of adding interceptors to a client config.  Impact is expected to be minimal because (1) customers are likely not providing their own interceptors, and (2) there is an alternate, non-mutating method (described above) to add interceptors.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
+++ b/Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
@@ -31,10 +31,5 @@ public protocol DefaultClientConfiguration: ClientConfiguration {
     /// If none is provided, only a default logger provider will be used.
     var telemetryProvider: TelemetryProvider { get set }
 
-    /// Adds an `InterceptorProvider` that will be used to provide interceptors for all operations.
-    ///
-    /// - Parameter provider: The `InterceptorProvider` to add.
-    func addInterceptorProvider(_ provider: InterceptorProvider)
-
     /// TODO(plugins): Add Checksum, etc.
 }
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
@@ -106,6 +106,8 @@ class SwiftWriter(
 
     fun addImport(symbol: Symbol) {
         symbol.references.forEach { addImport(it.symbol) }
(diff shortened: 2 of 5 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftWriter.kt
```diff
@@ -106,6 +106,8 @@ class SwiftWriter(
 
     fun addImport(symbol: Symbol) {
         symbol.references.forEach { addImport(it.symbol) }
+        val additionalImports = symbol.getProperty("additionalImports").getOrElse { emptyList<Symbol>() } as List<Symbol>
+        additionalImports.forEach { addImport(it) }
```
jbelkins, line 110: Additional imports were moved from the bottom to the top of this method because the early-exit on line 111 below causes additional imports to not be added.

Sources/ClientRuntime/Config/DefaultClientConfiguration.swift
```diff
@@ -31,10 +31,5 @@ public protocol DefaultClientConfiguration: ClientConfiguration {
     /// If none is provided, only a default logger provider will be used.
     var telemetryProvider: TelemetryProvider { get set }
 
-    /// Adds an `InterceptorProvider` that will be used to provide interceptors for all operations.
-    ///
-    /// - Parameter provider: The `InterceptorProvider` to add.
-    func addInterceptorProvider(_ provider: InterceptorProvider)
-
```
jbelkins, line 38: The add methods are removed from this & the following default config protocols

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/config/ClientConfiguration.kt
```diff
@@ -22,11 +21,6 @@ interface ClientConfiguration {
 
     fun getProperties(ctx: ProtocolGenerator.GenerationContext): Set<ConfigProperty>
 
-    /**
-     * The methods to render in the generated client configuration
-     */
-    fun getMethods(ctx: ProtocolGenerator.GenerationContext): Set<Function> = setOf()
```
jbelkins, line 28: This was only used for the add interceptor methods, so it is now removed.

jbelkins:
Closing in favor of an alternate approach

PR #917, opened 2025-03-19
chore: Make LogAgent conform to Sendable

sichanyoo:
Companion PR: https://github.com/awslabs/aws-sdk-swift/pull/1908

## Issue \#

N/A

## Description of changes

- Make `LogAgent` protocol conform to `Sendable`. This makes `SwiftLogger` struct `Sendable`, removing the need for `@preconcurrency` prefix for import when customers try to use the logger in concurrent context.
- Makes the test logger implementations with @unchecked Sendable; they contain mutable string array that stores logged messages for validating test results. Since tests don't need concurrency safety, just marked as unchecked.

## Scope

### DIFF
--- a/Sources/Smithy/Logging/LogAgent.swift
+++ b/Sources/Smithy/Logging/LogAgent.swift
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-public protocol LogAgent {
+public protocol LogAgent: Sendable {
     /// name of the struct or class where the logger was instantiated from
     var name: String { get }
 
--- a/Tests/ClientRuntimeTests/NetworkingTests/URLSession/FoundationStreamBridgeTests.swift
+++ b/Tests/ClientRuntimeTests/NetworkingTests/URLSession/FoundationStreamBridgeTests.swift
@@ -104,7 +104,7 @@ class FoundationStreamBridgeTests: XCTestCase {
     }
 }
 
-private class TestLogger: LogAgent {
+private class TestLogger: LogAgent, @unchecked Sendable {
     var name: String
 
     var messages: [(level: LogAgentLevel, message: String)] = []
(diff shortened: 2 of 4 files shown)


dayaffe (approved):

Sources/Smithy/Logging/LogAgent.swift
```diff
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-public protocol LogAgent {
+public protocol LogAgent: Sendable {
```
jbelkins, line 8: See new warnings related to TestLogger.  They should be easily fixed.

jbelkins (changes requested):

Sources/Smithy/Logging/LogAgent.swift
sichanyoo: Good catch, I think there's an instance in aws-sdk-swift unit tests also where this needs to be addressed.

Tests/ClientRuntimeTests/NetworkingTests/URLSession/FoundationStreamBridgeTests.swift
```diff
@@ -104,7 +104,7 @@ class FoundationStreamBridgeTests: XCTestCase {
     }
 }
 
-private class TestLogger: LogAgent {
+private class TestLogger: LogAgent, @unchecked Sendable {
```
jbelkins, line 107: It's preferred to fix any sendability issues when possible rather than just sidestepping the issue with `@unchecked`.

`@unchecked` is typically for when we rely on some mechanism for concurrent safety that the compiler can't check, such as GCD or C code that uses primitive locking.

sichanyoo: Summary of offline discussion:

The test loggers have string array "storage" that it appends logs to, and that's how the relevant tests validate on expected logger output.making test loggers an actor or putting @MainActor tag to the string array var doesn't work bc it complains about how actor isolated property / function can't count for protocol conformance. And using lock doesn't get recognized by concurrency check.

Keeping these the way they are since storing logs in nested property is something unique for testing purposes.

PR #920, opened 2025-04-14
feat: Add Swift 6 language mode support

dayaffe:
## Description of changes
- Bumps CRT version from 0.48 to 0.52.1 and includes changes from #925 
- Refactor CRTClientEngine to break up long functions
- Endpoint `properties` now is of type `EndpointProperty` as defined by CRT. A wrapper `EndpointPropertyValue` is defined to obfuscate CRT dependency
- Add github actions workflow that modifies tools version to 6.0 and builds

## Scope


.github/workflows/continuous-integration.yml
dayaffe, line 1: Ignore this

Package.resolved
dayaffe, line 1: Ignore this it will be removed

Package.swift
```diff
@@ -56,7 +56,7 @@ let package = Package(
     ],
     dependencies: {
         var dependencies: [Package.Dependency] = [
-            .package(url: "https://github.com/awslabs/aws-crt-swift.git", exact: "0.48.0"),
+            .package(url: "https://github.com/awslabs/aws-crt-swift.git", branch: "swift6-support"),
```
dayaffe, line 59: ignore this

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -477,6 +363,180 @@ public class CRTClientEngine: HTTPClient {
...
+    private func handleHTTP1Request(
+        connection: HTTPClientConnection,
+        request: HTTPRequest,
+        telemetryContext: TelemetryContext
+    ) async throws -> HTTPResponse {
+        logger.debug("Using HTTP/1.1 connection")
+        let crtRequest = try request.toHttpRequest()
+        let serverAddress = Self.makeServerAddress(request: request)
+
+        return try await withCheckedThrowingContinuation { [weak self] continuation in
```
sichanyoo, line 386: Q: What's the reason for having `[weak self]` in HTTP1 only and not for HTTP2? Also, is `[weak self]` actually needed here? What's the scenario where this closure would live on even after CRTClientEngine is deallocated?

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -182,203 +199,69 @@ public class CRTClientEngine: HTTPClient {
...
-        self.serialExecutor = SerialExecutor(config: config)
+
+        self.serialExecutor = SerialExecutor(
+            windowSize: config.windowSize,
+            maxConnectionsPerEndpoint: config.maxConnectionsPerEndpoint,
+            telemetry: config.telemetry,
+            connectTimeoutMs: config.connectTimeoutMs,
+            crtTLSOptions: config.crtTlsOptions,
+            socketTimeout: config.socketTimeout
+        )
```
sichanyoo, line 210: Q: What was the reason for adding a new initializer that's essentially a duplicate to initializer that takes in config?

sichanyoo (commented):
6 Qs & 1 suggestion

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -516,7 +576,9 @@ public class CRTClientEngine: HTTPClient {
             // resume the continuation as soon as we have all the initial headers
             // this allows callers to start reading the response as it comes in
             // instead of waiting for the entire response to be received
-            continuation.safeResume(response: response)
+            Task {
+                await continuation.safeResume(response: response)
+            }
```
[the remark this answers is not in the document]
sichanyoo: Resolved with changes in [a6a8040](https://github.com/smithy-lang/smithy-swift/pull/920/commits/a6a80403f5999f06dbd1a4cecea8fc32988bd87a).

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
```diff
@@ -131,7 +131,7 @@ public class DefaultAuthSchemeResolverParameters: AuthSchemeResolverParameters {
     }
 }
 
-public class DefaultAuthSchemeResolver: AuthSchemeResolver {
+public class DefaultAuthSchemeResolver: AuthSchemeResolver, @unchecked Sendable {
```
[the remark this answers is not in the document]
dayaffe: unchecked = we guarantee its safe, w/o unchecked compiler guarantees its safe
- If a class isnt final a subclass could add mutual global state making it suddenly unsafe so it needs unchecked instead of regular Sendable.
- HTTPRequest has mutable state like var headers which we allow to be changed. But we're saying its ok we know we're not changing headers in any concurrent contexts.

Sources/SmithyHTTPAPI/EndpointPropertyValue.swift
```diff
@@ -0,0 +1,62 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Smithy
+import enum AwsCommonRuntimeKit.EndpointProperty
+
+public enum EndpointPropertyValue: Sendable, Equatable {
```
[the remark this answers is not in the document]
dayaffe: I am indifferent, I wouldn't name it `SDKEndpointProperty` since this is defined in Smithy, but its also not part of the Smithy architecture so having that prefix could also be confusing. I also suffixed it `Value` since this is used in a dictionary context only as the value.

Sources/SmithyHTTPAPI/Endpoint.swift
```diff
@@ -72,24 +73,44 @@ public struct Endpoint: Hashable {
...
         self.uri = uri
         self.headers = headers
         self.properties = properties
     }
 }
 
+extension Endpoint {
+    public init(urlString: String,
+                headers: Headers = Headers(),
+                endpointProperties: [String: EndpointProperty]) throws {
```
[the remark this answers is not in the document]
dayaffe: I made it a different name to signal to the user that this is expecting a different type than the init with the same arguments -- CRT's EndpointProperty vs wrapper EndpointPropertyValue

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
dayaffe: HTTP1 method captures self when we call `makeHttpRequestStreamOptions` whereas HTTP2 doesn't. That method leads to a bunch of CRT callbacks that execute later

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
dayaffe: Ah good point, I can remove this now. Originally needed the CRTClientConfig to be made Sendable which required changes in CRT that ended up being made.

Sources/SmithyHTTPAPI/Endpoint.swift
```diff
@@ -72,24 +73,44 @@ public struct Endpoint: Hashable {
...
+        guard let url = URLComponents(string: urlString)?.url else {
+            throw ClientError.unknownError("invalid url \(urlString)")
+        }
+
+        let properties = endpointProperties.mapValues(EndpointPropertyValue.init)
+        try self.init(url: url, headers: headers, properties: properties)
+    }
+}
```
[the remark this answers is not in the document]
dayaffe: Its used in DefaultEndpointResolver when we get endpoint properties directly from CRT. This does the work converting EndpointProperties (CRT) to EndpointPropertyValues (Wrapper)

Sources/ClientRuntime/Config/DefaultSDKRuntimeConfiguration.swift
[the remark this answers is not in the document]
dayaffe: I made DefaultAuthSchemeResolver Sendable, but Context, DefaultMessageDecoder, and HTTPRequest maintain some mutable state which should be ok but require unchecked.

sichanyoo (approved):

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -561,21 +589,31 @@ public class CRTClientEngine: HTTPClient {
...
 
-        public func safeResume(error: Error) {
-            continuation?.resume(throwing: error)
-            self.continuation = nil
+        /// Fetch the continuation exactly once. Returns the current value under
+        /// a lock, then sets it to `nil` so later callers get nothing.
+        private func take() -> CheckedContinuation<Response, Error>? {
+            lock.lock()
+            defer { lock.unlock() }
+            defer { continuation = nil }
```
jbelkins, line 615: Multiple `defer` statements execute in reverse order they are written, so this is correct, but for simplicity, why not just use one defer with two lines and return a temp var?

Sources/SmithyHTTPAPI/Endpoint.swift
```diff
@@ -7,8 +7,9 @@
 
 import Foundation
 import Smithy
+import enum AwsCommonRuntimeKit.EndpointProperty
 
-public struct Endpoint: Hashable {
+public struct Endpoint: Sendable, Equatable {
```
jbelkins, line 12: So we were using Endpoint equality but not hashability?

Sources/SmithyHTTPAPI/EndpointPropertyValue.swift
```diff
@@ -0,0 +1,62 @@
...
+    case bool(Bool)
+    case string(String)
+    indirect case array([EndpointPropertyValue])
+    indirect case dictionary([String: EndpointPropertyValue])
```
jbelkins, line 16: Where are the possible endpoint property types defined?  Do we need an unknown case for this?

Sources/SmithyHTTPAPI/EndpointPropertyValue.swift
```diff
@@ -0,0 +1,62 @@
...
+
+extension EndpointPropertyValue {
+    init(_ endpointProperty: EndpointProperty) {
+        switch endpointProperty {
+        case .bool(let value):
+            self = .bool(value)
+        case .string(let value):
+            self = .string(value)
+        case .array(let values):
+            self = .array(values.map(EndpointPropertyValue.init))
```
jbelkins, line 33: lol nice use of function pointers

Sources/SmithyHTTPAPI/HTTPRequest.swift
```diff
@@ -274,3 +274,5 @@ extension HTTPRequestBuilder {
         return String(signatureSequence)
     }
 }
+
+extension HTTPRequest: @unchecked Sendable {}
```
jbelkins, line 278: Rather than bury this extension at the bottom, make sure that it's clear at the point that the type is defined that it's unchecked and the developer must ensure the type is Sendable, not the compiler

jbelkins (changes requested):

Package.resolved
```diff
@@ -0,0 +1,33 @@
+{
```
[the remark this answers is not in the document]
dayaffe: Yes it will be

Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
dayaffe: I can change it to that if its preferred 👍 -- I wrote out both options and liked this one more.

Sources/SmithyHTTPAPI/Endpoint.swift
jbelkins: Do we actually use endpoint equality though?

dayaffe: Its used [here](https://github.com/smithy-lang/smithy-swift/blob/da2b2e906b6a8ef3674541b372c79c8eafd811c6/Sources/ClientRuntime/Endpoints/SmithyEndpoint.swift#L14)

jbelkins (approved):

PR #921, opened 2025-04-15
fix: Fix silent failure with CRT HTTP client when response stream has an error

sichanyoo:
CI PR in SDK: https://github.com/awslabs/aws-sdk-swift/pull/1919

## Issue \#

N/A

## Description of changes

- Bug found as part of investigating why the integration test in the PR https://github.com/awslabs/aws-sdk-swift/pull/1917 throws an error in CRT HTTP client but doesn't fail the test case.
- Ran the same integ test with the change in this PR and error gets thrown up and fails the test case as expected.

## Scope


jbelkins (approved):
Merge only if this passes our SDK integration test suite.

sichanyoo:
Copy PR: https://github.com/smithy-lang/smithy-swift/pull/922

PR #924, opened 2025-04-22
fix!: Represent TCP port as UInt16

jbelkins:
## Description of changes
[BREAKING] Changes the type for port number from `Int16` to `UInt16`.  TCP port may be from 0-65535 which is beyond the range of `Int16`.

This affects public interfaces for URI, Endpoint, HTTPRequest, and their associated builder types.

The change is breaking because port number is exposed as a property of several public types and parameters to public functions.  The change is being made nonetheless because it is a significant impediment to not allow use of all possible TCP ports.

## Scope
- [x] Breaking bug fix (breaking change which fixes an issue)


dayaffe (approved):
Same comment as companion PR. I think port should be `Int` until the exact place it needs to be converted. At which point we could use something like https://developer.apple.com/documentation/kernel/in_port_t to ensure it is a valid port.

PR #925, opened 2025-04-28
chore: Bump CRT version to 0.51.0 from 0.48.0

sichanyoo:
Corresponding aws-sdk-swift PR: https://github.com/awslabs/aws-sdk-swift/pull/1927

## Issue \#

2819

## Description of changes

- CRT 0.51.0 removes FileHandle's conformance to IStreamable which used to contain 3 methods, one of which is `length()`.  The `length()` function was the only thing we were using; so conformance was removed by CRT and that method is moved to SDK side.
- `isEndOfStream()` is added as corresponding change to CRT-side bug fix. The fix addresses CRT HTTP client issue where it used to send empty frame after every nonempty data frame in HTTP/2 connection.

## Scope

### DIFF
--- a/Sources/SmithyStreams/StreamableHttpBody.swift
+++ b/Sources/SmithyStreams/StreamableHttpBody.swift
@@ -109,4 +109,8 @@ public class StreamableHttpBody: IStreamable {
             return nil
         }
     }
+
+    public func isEndOfStream() -> Bool {
+        return false
+    }
 }
--- a/Sources/SmithyStreams/StreamableHttpBody.swift
+++ b/Sources/SmithyStreams/StreamableHttpBody.swift
@@ -111,6 +111,10 @@ public class StreamableHttpBody: IStreamable {
     }
 
     public func isEndOfStream() -> Bool {
-        return false
+        do {
+            return try self.position >= self.length()
+        } catch {
+            return false
+        }
     }
 }
--- a/Package.swift
+++ b/Package.swift
@@ -56,7 +56,7 @@ let package = Package(
     ],
     dependencies: {
         var dependencies: [Package.Dependency] = [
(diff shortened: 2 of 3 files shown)


Sources/SmithyStreams/StreamableHttpBody.swift
```diff
@@ -109,4 +109,8 @@ public class StreamableHttpBody: IStreamable {
             return nil
         }
     }
+
+    public func isEndOfStream() -> Bool {
+        return false
```
jbelkins, line 114: What's the rationale for this always being false?

jbelkins (commented):
1 Q inline

jbelkins:
> 1 Q inline

(also see integration test failures in aws-sdk-swift)

Sources/SmithyStreams/StreamableHttpBody.swift
sichanyoo: Rationale was provided in CRT side [here](https://github.com/awslabs/aws-crt-swift/compare/v0.50.0...v0.51.0#diff-1fa8771c6bca4f8944174b59d9deb27d25e8536e47f7ecd7dad3d3bf9ab8a74dR33-R35).

sichanyoo: After discussion with Waqar realized it has to have actual impl in SDK side; updated here: https://github.com/smithy-lang/smithy-swift/pull/925/commits/55dd83c014dc630702a84c4d290c32b4cd4ea6e0

sichanyoo:
Version bump to go out with https://github.com/smithy-lang/smithy-swift/pull/920, as there exists a bug in 0.51.0 and fix for the bug was added by rolling forward to 0.52.1 due to branch mgmt complications.

PR #926, opened 2025-05-01
fix: Disambiguate .none default values

jbelkins:
## Description of changes
Default values for enum-typed model properties can be ambiguous when the enum includes a case named `.none` because the compiler cannot currently differentiate between the enum's `.none` value and the `Optional.none` value.  To make matters worse, the compiler assumes `Optional.none` is the desired outcome, which is not what we actually want.

To allow the compiler to correctly resolve which `.none` to use, render enum-typed default values with the fully-qualified type name, i.e. render `AWSService.CustomEnum.none` as the default value instead of merely `.none`.  This namespace is added for all enum default values, not exclusively for the ones named `.none`.

This PR affects rendered code for 14 current AWS services, although only one of those has an enum case named `.none`.

## Scope

### DIFF
--- a/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
+++ b/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
@@ -219,7 +219,7 @@ open class MemberShapeDecodeGenerator(
             }
             // Provide a default value dependent on the type.
             return when (targetShape) {
-                is EnumShape -> " ?? .${enumDefaultValue(targetShape, it.expectStringNode().value)}"
+                is EnumShape -> " ?? ${enumDefaultValue(writer, targetShape, it.expectStringNode().value)}"
                 is IntEnumShape -> intEnumDefaultValue(it)
                 is StringShape -> " ?? \"${it.expectStringNode().value}\""
                 is ByteShape -> " ?? ${it.expectNumberNode().value}"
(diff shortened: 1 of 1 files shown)


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -219,7 +219,7 @@ open class MemberShapeDecodeGenerator(
             }
             // Provide a default value dependent on the type.
             return when (targetShape) {
-                is EnumShape -> " ?? .${enumDefaultValue(targetShape, it.expectStringNode().value)}"
+                is EnumShape -> " ?? ${enumDefaultValue(writer, targetShape, it.expectStringNode().value)}"
```
jbelkins, line 222: `writer` is passed in so the symbol in the default value can be tracked for import

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/serde/member/MemberShapeDecodeGenerator.kt
```diff
@@ -250,14 +250,19 @@ open class MemberShapeDecodeGenerator(
...
         val matchingMember =
             enumShape.members().first { member ->
                 value == member.expectTrait<EnumValueTrait>().expectStringValue()
             }
-        return swiftEnumCaseName(matchingMember.memberName, value)
+        return writer.format(
+            "\$N.\$L",
+            ctx.symbolProvider.toSymbol(enumShape),
+            swiftEnumCaseName(matchingMember.memberName, value),
+        )
```
jbelkins, line 265: Writer is used to generate a string with the fully-qualified default value

dayaffe (approved):

PR #930, opened 2025-05-27
fix: Silenced network error in CRTClientEngine onStreamComplete callback

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1943

## Description of changes

Adds back `continuation.resume(throwing:)` to `onStreamComplete` callback in `CRTClientEngine`. The continuation resumption was removed in this previous PR: https://github.com/smithy-lang/smithy-swift/pull/922. At the time of removal, the logic was that by the time `onStreamComplete` callback is reached, the continuation must have been resumed in the `onResponse` callback since `onResponse` callback gets called as soon as we get headers from the service. But on transient network errors, only the `onStreamComplete` could get called & `onResponse` could be skipped. This causes continuation to never get resumed, silencing error & essentially "freezing" a thread.

## Scope

### DIFF
--- a/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
+++ b/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
@@ -571,6 +571,7 @@ public class CRTClientEngine: HTTPClient {
             case .failure(let error):
                 self.logger.error("Response encountered an error: \(error)")
                 stream.closeWithError(error)
+                continuation.resume(throwing: error)
             }
         }
 


Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift
```diff
@@ -571,6 +571,7 @@ public class CRTClientEngine: HTTPClient {
             case .failure(let error):
                 self.logger.error("Response encountered an error: \(error)")
                 stream.closeWithError(error)
+                continuation.resume(throwing: error)
```
jbelkins, line 574: We're sure we don't need this in the success branch of the switch as well?

sichanyoo: I think so, the success with status code means valid response was returned from the service, meaning onResponse would've been called already that resumes the continuation. Not to mention customer's been using v1.2.9 of aws-sdk-swift that depends on v.0.112.0 of smithy-swift without any issues, which only has this resume logic in failure case.

jbelkins (approved):

PR #937, opened 2025-06-17
chore: Update Smithy to 1.59.0

jbelkins:
## Description of changes
Updates the SDK to Smithy 1.59.0 (latest version.)
- Fixes a bug where protocol test expectations render incorrectly
- Changes behavior so that a prefix header will not overwrite a header with the same name

Companion aws-sdk-swift PR: https://github.com/awslabs/aws-sdk-swift/pull/1957

## Scope


Sources/SmithyHTTPAPI/Headers.swift
```diff
@@ -191,11 +191,9 @@ extension Headers: Equatable {
...
     /// - Returns: `true` if the two values are equal irrespective of order, otherwise `false`.
     public static func == (lhs: Headers, rhs: Headers) -> Bool {
-        lhs.access { lhsHeaders in
-            rhs.access { rhsHeaders in
-                lhsHeaders.sorted() == rhsHeaders.sorted()
-            }
-        }
+        let lhsHeaders = lhs.access { $0 }.sorted()
+        let rhsHeaders = rhs.access { $0 }.sorted()
+        return lhsHeaders == rhsHeaders
```
jbelkins, line 196: Refactored to avoid simultaneously holding two locks.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/ShapeValueGenerator.kt
```diff
@@ -54,6 +55,13 @@ class ShapeValueGenerator(
         shape: Shape,
         params: Node,
     ) {
+        // If the node for the value to be written is NULL, then the value for this shape is not present.
+        // Just write a Swift `nil` and stop rendering.
+        if (params.type == NodeType.NULL) {
+            writer.writeInline("nil")
+            return
+        }
+
```
jbelkins, line 64: We were incorrectly rendering the expected value when a collection or aggregate shape when it is supposed to be absent (i.e. "null" in the Smithy node tree for expected value.)  This if-else handles that case.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/providers/HttpHeaderProvider.kt
```diff
@@ -180,6 +180,10 @@ class HttpHeaderProvider(
                 val mapValueShapeTargetSymbol = ctx.symbolProvider.toSymbol(mapValueShapeTarget)
 
                 writer.openBlock("for (prefixHeaderMapKey, prefixHeaderMapValue) in $memberName { ", "}") {
+                    // Don't write a prefix header over a specific header that was also written to this request.
+                    // See the HttpEmptyPrefixHeadersRequestClient protocol tests on the REST protocols.
+                    writer.write("guard !items.exists(name: \"$paramName\\(prefixHeaderMapKey)\") else { continue }")
+
```
jbelkins, line 186: New protocol tests specify that specific headers bound to names can overwrite prefix headers.  This `guard` creates that behavior.

sichanyoo (approved):

PR #945, opened 2025-07-02
feat: Credential feature IDs

sichanyoo:
Corresponding PR: 
-

## Issue \#

2302

## Description of changes

- Adds `properties` field to identity types.

## Scope


Sources/SmithyIdentityAPI/Identity.swift
```diff
@@ -6,8 +6,10 @@
 //
 
 import struct Foundation.Date
+import struct Smithy.Attributes
```
sichanyoo, line 9: Adds grab-bag property called `properties` to `AWSCredentialIdentity`. For now it's only used for business metrics feature IDs for credentials. Feature IDs are AWS specific and so this property bag acts as a way to keep smithy-swift generic. Not to mention it allows maximum flexibility for miscellaneous values not fundamental to credential object itself, in case SDK needs to store even more misc. AWS specific values with credential objects down the line.

All the other changes in this PR is just to accommodate this addition to the protocol.

PR #946, opened 2025-07-09
feat: Checksum reuse between retries in chunked streaming requests

sichanyoo:
Companion PR: 
- https://github.com/awslabs/aws-sdk-swift/pull/1980

## Issue \#

3359

## Scope


Sources/ClientRuntime/Networking/Http/Middlewares/SignerMiddleware.swift
```diff
@@ -57,6 +57,11 @@ extension SignerMiddleware: ApplySigner {
             updatedSigningProperties.set(key: AttributeKey(name: "SignedBodyValue"), value: bodyValue)
         }
 
+        if case .stream(let stream) = request.body, stream.isEligibleForChunkedStreaming {
+            // Pass in context object via signing properties to reuse final checksum value in chunked streaming.
+            updatedSigningProperties.set(key: AttributeKey(name: "Context"), value: attributes)
```
jbelkins, line 62: Be careful here... if we put Context into signing properties then we put those signing properties into Context, a reference cycle is formed and a memory leak will occur.

Maybe better to copy the specific things we need from Context into signingProperties?

jbelkins (commented):
See comment on storing Context in signing properties.

PR #952, opened 2025-07-30
chore: Refactor internal clients

jbelkins:
## Description of changes
Supports the changes to internal clients in https://github.com/awslabs/aws-sdk-swift/pull/1995 .
- Adds a protocol that can be added to credential resolver types to signify that a credential resolver was not customer-supplied.
- Adds implementations of that protocol to the static identity resolver types.
- Adds an `internalClient` Boolean field to SwiftSettings, to help make codegen code more clear.
- Adds awareness of the `package` keyword introduced to Swift in Swift 5.9.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSettings.kt
```diff
@@ -64,6 +65,7 @@ class SwiftSettings(
     val mergeModels: Boolean,
     val copyrightNotice: String,
     val visibility: String,
+    val internalClient: Boolean,
```
jbelkins, line 68: Added this `internalClient` field to SwiftSettings to help make codegen more clear, i.e. can query this setting directly rather than rely on some string value of `visibility`.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/lang/ReservedWords.kt
```diff
@@ -83,6 +83,7 @@ val reservedWords =
         "operator",
         "optional",
         "override",
+        "package",
```
jbelkins, line 86: Adding `package` as a Swift reserved word so that it gets escaped with backticks in generated code.

There actually do seem to be services that already use `package` as an identifier without escaping it, but to date it seems not to cause any problems.

Sources/SmithyIdentity/BearerTokenIdentityResolvers/ClientConfigDefaultBearerTokenIdentityResolver.swift
```diff
@@ -0,0 +1,27 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import struct Smithy.Attributes
```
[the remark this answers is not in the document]
sichanyoo: _A suggestion_

The new wrapper `ClientConfigDefaultBearerTokenIdentityResolver`  and its usage feels a bit hacky to me. The name also isn't too clear on what it's really used for but it would be hard to come up with a name that really describes what it's used for in a concise manner.

What if we just passed in a closure (named something like `tokenResolutionClosure: () -> BearerTokenIdentity`) to `BedrockAPIKeyInterceptor` at initialization time, and call that closure if bearer token should be resolved? Then `BedrockAPIKeyInterceptor` doesn't need _any_ compile time access to a token resolver type. We could codegen in relevant service clients (e.g., bedrock and bedrock runtime) something like this in the operation stack:
```
builder.interceptors.add(AWSClientRuntime.BedrockAPIKeyInterceptor(
  isCustom: { config.bearerTokenIdentityResolver is DefaultBearerTokenIdentityResolverChain },
  tokenResolutionClosure: { try await config.bearerTokenIdentityResolver.getIdentity() } 
))
```
The `isCustom` can be used to just `guard !isCustom else { return }` at the beginning of the `BedrockAPIKeyInterceptor`.

sichanyoo (commented):
A suggestion on `ClientConfigDefaultBearerTokenIdentityResolver`

Sources/SmithyIdentity/AWSCredentialIdentityResolvers/StaticAWSCredentialIdentityResolver.swift
```diff
@@ -5,11 +5,17 @@
...
 //
 
+@_spi(ClientConfigDefaultIdentityResolver) import protocol SmithyIdentityAPI.ClientConfigDefaultIdentityResolver
 import struct Smithy.Attributes
 
 /// A credential identity resolver that provides a fixed set of credentials
 public struct StaticAWSCredentialIdentityResolver: AWSCredentialIdentityResolver {
-    private let credentials: AWSCredentialIdentity
+    fileprivate let credentials: AWSCredentialIdentity
+
```
jbelkins, line 14: The `@_spi` initializer below creates a AWS credential identity resolver with empty credentials.  This is used in internal clients, where the default AWS credential identity resolver is never used.

Sources/SmithyIdentity/AWSCredentialIdentityResolvers/StaticAWSCredentialIdentityResolver.swift
```diff
@@ -24,3 +30,11 @@ public struct StaticAWSCredentialIdentityResolver: AWSCredentialIdentityResolver
         return credentials
     }
 }
+
```
jbelkins, line 33: The `ClientConfigDefaultIdentityResolver` extension below marks this resolver as a client config default if the credentials are empty, something that the new initializer above does, but a real customer should never do.

Sources/SmithyIdentity/BearerTokenIdentityResolvers/StaticBearerTokenIdentityResolver.swift
```diff
@@ -5,11 +5,17 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+@_spi(ClientConfigDefaultIdentityResolver) import protocol SmithyIdentityAPI.ClientConfigDefaultIdentityResolver
```
jbelkins, line 8: Pretty much the same changes in this `StaticBearerTokenIdentityResolver` as in the `StaticAWSCredentialIdentityResolver` just above.

Sources/SmithyIdentityAPI/ClientConfigDefaultIdentityResolver.swift
```diff
@@ -0,0 +1,17 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// A protocol on identity resolver used to signify that this resolver is a default resolver created because the client config was not passed a custom resolver at creation.
+///
+/// Resolvers that do not implement this protocol should be presumed to not be a client config default.
+
```
jbelkins, line 11: The protocol below is used to determine if a credential resolver came from client config defaults.  It's protected with `@_spi` so it isn't visible to customers.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/swiftmodules/ClientRuntimeTypes.kt
```diff
@@ -108,12 +108,12 @@ private fun runtimeSymbol(
     name: String,
     declaration: SwiftDeclaration?,
     additionalImports: List<Symbol> = emptyList(),
```
jbelkins, line 110: Just fixed a typo in this file

sichanyoo (approved):

PR #968, opened 2025-09-25
feat: Automatic clock skew adjustment

jbelkins:
## Description of changes
Automatically adjusts for differences between client and server wall clock time, to prevent signed requests from being rejected due to expired signature.

The basic smithy-swift implementation has a customization point for clock skew determination, but never detects a clock skew by default.  The actual clock-skew determination to be used on AWS services is implemented in the companion aws-sdk-swift PR.
- When clock skew is detected:
  - Clock skew is calculated.
  - The time difference is stored in a global store for future use when talking to that server.
  - A retry of the request is initiated.  The retry will use the calculated clock skew to adjust its signature expiration.
- Subsequent requests to the same host will continue to use the previously calculated clock skew.
- `SignerMiddleware` and `SigV4Signer` are modified to use clock skew when one is set.

Also:
- Removed unneeded `SmithyRetries` dependency on CRT
- Code cleanup of imports, unused code, etc.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/middlewares/ClockSkewMiddleware.kt
```diff
@@ -0,0 +1,31 @@
...
+package software.amazon.smithy.swift.codegen.integration.middlewares
+
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.codegen.core.SymbolProvider
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.swift.codegen.SwiftWriter
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.middleware.MiddlewareRenderable
+
```
jbelkins, line 15: This type renders the clock skew provider into the orchestrator.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HTTPBindingProtocolGenerator.kt
```diff
@@ -437,6 +438,7 @@ abstract class HTTPBindingProtocolGenerator(
             )
             operationMiddleware.appendMiddleware(operation, DeserializeMiddleware(ctx.model, ctx.symbolProvider))
             operationMiddleware.appendMiddleware(operation, LoggingMiddleware(ctx.model, ctx.symbolProvider))
+            operationMiddleware.appendMiddleware(operation, ClockSkewMiddleware(ctx.model, ctx.symbolProvider, clockSkewProviderSymbol))
```
jbelkins, line 441: Creates the type directly above to render the proper middleware.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/ProtocolGenerator.kt
```diff
@@ -85,6 +87,9 @@ interface ProtocolGenerator {
      */
     var serviceErrorProtocolSymbol: Symbol
 
+    val clockSkewProviderSymbol: Symbol
+        get() = DefaultClockSkewProviderSymbol
+
```
jbelkins, line 92: Here & above is the symbol for the clock skew provider.

In aws-sdk-swift, this property is overridden to provide an AWS-specific implementation.

Sources/ClientRuntime/ClockSkew/ClockSkewProvider.swift
```diff
@@ -0,0 +1,11 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import struct Foundation.Date
+import struct Foundation.TimeInterval
+
+public typealias ClockSkewProvider<Request, Response> = @Sendable (Request, Response, Error, Date) -> TimeInterval?
```
jbelkins, line 11: Just a typealias for the closure type, since it's pretty long.

Sources/ClientRuntime/ClockSkew/ClockSkewStore.swift
```diff
@@ -0,0 +1,30 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import struct Foundation.TimeInterval
```
jbelkins, line 8: This type is just a concurrency-safe store for clock skew values, keyed by the hostname of the server where that clock skew was measured.

Sources/ClientRuntime/ClockSkew/DefaultClockSkewProvider.swift
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import struct Foundation.Date
+import struct Foundation.TimeInterval
+import class SmithyHTTPAPI.HTTPRequest
+import class SmithyHTTPAPI.HTTPResponse
+
+public enum DefaultClockSkewProvider {
+
```
jbelkins, line 14: This is the "generic Smithy" clock skew provider.  It does not do anything.

aws-sdk-swift provides its own implementation of this which actually does measure clock skew.

Sources/ClientRuntime/Networking/Http/Middlewares/SignerMiddleware.swift
```diff
@@ -50,8 +50,14 @@ extension SignerMiddleware: ApplySigner {
             )
         }
 
-        // Check if CRT should be provided a pre-computed Sha256 SignedBodyValue
         var updatedSigningProperties = signingProperties
+
```
jbelkins, line 54: Below, the clock skew store is checked for a clock skew value. If one exists, it is entered into signing properties. Then it will be applied when calculating the signature.

Sources/ClientRuntime/Orchestrator/Orchestrator.swift
```diff
@@ -262,9 +270,20 @@ public struct Orchestrator<
         do {
             _ = try context.getOutput()
             await strategy.recordSuccess(token: token)
-        } catch let error {
-            // If we can't get errorInfo, we definitely can't retry
-            guard let errorInfo = retryErrorInfoProvider(error) else { return }
+        } catch {
```
jbelkins, line 273: Below, the clock skew provider is called to obtain a clock skew value.  If one is provided, it is recorded in the clock skew store and a retry will be performed (subject to if there are any retries remaining.)

Sources/SmithyHTTPAuth/SigV4Signer.swift
```diff
@@ -127,7 +128,7 @@ public class SigV4Signer: SmithyHTTPAuthAPI.Signer, @unchecked Sendable {
             signedBodyHeader: signedBodyHeader,
             signedBodyValue: signedBodyValue,
             flags: flags,
-            date: Date(),
+            date: Date().addingTimeInterval(clockSkew),
```
jbelkins, line 131: Clock skew is applied to current date in the signing config to adjust signature expiration for the difference in client vs. server clock time.

Sources/SmithyRetries/ExponentialBackOffJitterType.swift
```diff
@@ -4,23 +4,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 //
-import AwsCommonRuntimeKit
```
jbelkins, line 7: The extension below is unused so it & the CRT dependency are removed.

jbelkins: RetryErrorType+CRT below is entirely unneeded and is removed.

Tests/SmithyRetriesTests/RetryIntegrationTests.swift
```diff
@@ -54,7 +54,13 @@ final class RetryIntegrationTests: XCTestCase {
...
             .retryErrorInfoProvider(DefaultRetryErrorInfoProvider.errorInfo(for:))
             .retryStrategy(subject)
-            .deserialize({ _, _ in TestOutputResponse() })
+            .deserialize({ response, _ in
+                if response.statusCode == .ok {
+                    return TestOutputResponse()
+                } else {
+                    throw TestHTTPError(statusCode: response.statusCode)
+                }
+            })
```
jbelkins, line 63: The deserializer above and the switch statement below were corrected to ensure that a HTTP response is available during retries.

Tests/SmithyRetriesTests/RetryIntegrationTests.swift
```diff
@@ -221,9 +227,10 @@ private class TestOutputHandler: ExecuteRequest {
         // Return either a successful response or a HTTP error, depending on the directions in the test step.
         switch testStep.response {
         case .success:
-            return HTTPResponse()
+            return HTTPResponse(statusCode: .ok)
         case .httpError(let statusCode):
-            throw TestHTTPError(statusCode: statusCode)
+            let httpStatusCode = HTTPStatusCode(rawValue: statusCode)!
+            return HTTPResponse(statusCode: httpStatusCode)
```
jbelkins, line 233: Rather than throw here, successfully return an HTTP response that indicates a non-success error code.
Instead, the service error will get thrown in the deserializer above (this is how "real" clients work.)

Package.swift
```diff
@@ -128,10 +128,7 @@ let package = Package(
         ),
         .target(
             name: "SmithyRetries",
-            dependencies: [
-                "SmithyRetriesAPI",
-                .product(name: "AwsCommonRuntimeKit", package: "aws-crt-swift"),
-            ]
+            dependencies: ["SmithyRetriesAPI"]
```
jbelkins, line 131: Removed the CRT dependency from SmithyRetries since the only CRT references in that target were unused and are now removed.

Sources/ClientRuntime/Orchestrator/OrchestratorBuilder.swift
```diff
@@ -33,6 +35,7 @@ public class OrchestratorBuilder<
     internal var deserialize: ((ResponseType, Context) async throws -> OutputType)?
     internal var retryStrategy: (any RetryStrategy)?
     internal var retryErrorInfoProvider: ((Error) -> RetryErrorInfo?)?
+    internal var clockSkewProvider: (ClockSkewProvider<RequestType, ResponseType>)?
```
jbelkins, line 38: The `OrchestratorBuilder` is modified to accept a clock skew provider.  This is set in codegen when the orchestrator is constructed for an operation.

Sources/Smithy/RequestMessage.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 /// Message that is sent from client to service.
-public protocol RequestMessage {
+public protocol RequestMessage: Sendable {
```
jbelkins, line 9: `Sendable` is added to request & response types so they can be safely passed through async contexts.

(`HTTPRequest` and `HTTPResponse` are both already `Sendable` so no implementation change needed for this.)

Sources/SmithyHTTPAPI/HTTPRequest.swift
```diff
@@ -35,24 +36,27 @@ public final class HTTPRequest: RequestMessage, @unchecked Sendable {
     public var path: String { destination.path }
     public var queryItems: [URIQueryItem]? { destination.queryItems }
     public var trailingHeaders: Headers = Headers()
-    public var endpoint: Endpoint {
-        return Endpoint(uri: self.destination, headers: self.headers)
-    }
+    public var endpoint: Endpoint { .init(uri: destination, headers: headers) }
+    public internal(set) var signedAt: Date?
```
jbelkins, line 40: `signedAt` property is added to `HTTPRequest`.  It is `nil` until the request is signed, then it is set with the `Date` that was used as "now" for signing.

Sources/SmithyHTTPAPI/HTTPRequest.swift
```diff
@@ -156,6 +161,7 @@ public final class HTTPRequestBuilder: RequestMessageBuilder {
     public private(set) var port: UInt16?
     public private(set) var protocolType: URIScheme = .https
     public private(set) var trailingHeaders: Headers = Headers()
+    public private(set) var signedAt: Date?
```
jbelkins, line 164: Here, `signedAt` is added to the builder for `HTTPRequest`.

Sources/SmithyHTTPAuth/SigV4Signer.swift
```diff
@@ -52,7 +52,13 @@ public class SigV4Signer: SmithyHTTPAuthAPI.Signer, @unchecked Sendable {
             )
         }
 
-        let signingConfig = try constructSigningConfig(identity: identity, signingProperties: signingProperties)
+        let signedAt = Date()
```
jbelkins, line 55: The signer gets current `Date` and keeps it in a var, so the same date can be used for signing and for updating the `HTTPRequest.signedAt` property.

Sources/SmithyHTTPAuth/SigV4Signer.swift
```diff
@@ -127,7 +139,7 @@ public class SigV4Signer: SmithyHTTPAuthAPI.Signer, @unchecked Sendable {
             signedBodyHeader: signedBodyHeader,
             signedBodyValue: signedBodyValue,
             flags: flags,
-            date: Date(),
+            date: signedAt.addingTimeInterval(clockSkew),
```
jbelkins, line 142: Correct the `signedAt` value with clock skew, then pass it into the signer.

Tests/ClientRuntimeTests/NetworkingTests/Http/SdkRequestBuilderTests.swift
```diff
@@ -18,7 +18,7 @@ class SdkRequestBuilderTests: XCTestCase {
         let crtRequest = try HTTPRequest()
         crtRequest.path = pathToMatch
 
-        let updatedRequest = HTTPRequestBuilder().update(from: crtRequest, originalRequest: originalRequest).build()
+        let updatedRequest = HTTPRequestBuilder().update(from: crtRequest, originalRequest: originalRequest, signedAt: Date()).build()
```
jbelkins, line 21: A `signedAt` value had to be passed to allow compile.  It is not used in this test.

Tests/SmithyHTTPClientTests/HttpRequestTests.swift
```diff
@@ -150,4 +150,10 @@ class HttpRequestTests: NetworkingTestUtils {
         let endpoint = Endpoint(host: "", path: "")
         XCTAssertNil(endpoint.url, "An invalid endpoint should result in a nil URL.")
     }
+
+    func test_requestBuilder_SignedAt() {
```
jbelkins, line 154: Simple test to verify builder sets `signedAt` field in built request

dayaffe (approved):

PR #976, opened 2025-10-06
fix: Eliminate compile warning in generated auth scheme resolvers

jbelkins:
## Description of changes
Eliminates a compile warning by using `let` instead of `var` for variables that hold auth scheme options that are not modified.

## Scope


smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/requestandresponse/requestflow/AuthSchemeResolverGeneratorTests.kt
```diff
@@ -42,28 +42,28 @@ public struct DefaultExampleAuthSchemeResolver: ExampleAuthSchemeResolver {
         }
         switch serviceParams.operation {
```
jbelkins, line 43: Below is just test expectations that are changed due to changing `var` to `let`.

sichanyoo (approved):

PR #993, opened 2025-11-09
feat: Add Swift-native codegen plugin

jbelkins:
## Description of changes
Adds the framework for a native-Swift code generator, in the form of a Swift build tool plugin plus a CLI tool.

The plugin is added to the service client target by adding it to the service client target's `plugins` param.  (For `aws-sdk-swift`, this is performed in companion PR https://github.com/awslabs/aws-sdk-swift/pull/2053).

Using existing code generation, the `smithy-model-info.json` is added to the `Sources` in each service client.  That file contains the path in the project to the Smithy JSON model for that service.  Example contents of `smithy-model-info.json`:
```json
{"path":"codegen/sdk-codegen/aws-models/sts.json"}
```

The code generator:
- Loads the `smithy-model-info.json` file for that client, then uses the `path` from that file to load the model.
- Creates a generated Swift file `ModelCount.swift` with a global variable `modelCount` holding the size of the model file in bytes (this proves that the code generator can access the model.)
- The `modelCount` var has a documentation comment, to show that doc comments in generated Swift files get picked up.

Example of generated code (for `AWSSTS`):
```swift
//
// Copyright Amazon.com Inc. or its affiliates.
// All Rights Reserved.
//
//

// Code generated by SmithyCodegenCLI. DO NOT EDIT!

/// The count of bytes in the model.
public let modelCount = 287051
```

This example code generation is trivial, but it does prove that code generation from a model is possible using Swift build plugins.

Dependency changes:
- Adds https://github.com/apple/swift-argument-parser as a dependency.
  - This allows the code generator executable (target `SmithyCodegenCLI`, defined below) to benefit from the utility `swift-argument-parser` provides.
  - This dependency is already present because it is required by `aws-crt-swift`, so no change in development footprint.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HTTPBindingProtocolGenerator.kt
```diff
@@ -209,8 +209,9 @@ abstract class HTTPBindingProtocolGenerator(
 
     private fun usesSchemaBasedSerialization(ctx: ProtocolGenerator.GenerationContext): Boolean =
         // This fun is temporary; it will be eliminated when all services/protocols are moved to schema-based
-        ctx.service.allTraits.keys
-            .any { it.name == "rpcv2Cbor" }
+        false
```
jbelkins: Smithy/JVM codegen of schemas is intentionally disabled here

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SmithyModelFileInfoGenerator.kt
```diff
@@ -0,0 +1,25 @@
+package software.amazon.smithy.swift.codegen
+
+import software.amazon.smithy.aws.traits.ServiceTrait
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.model.getTrait
+
```
jbelkins: This just generates a very simple JSON file named `smithy-model-info.json` into the directory where the Swift source files go.  Its purpose is to tell the Swift code generator where it can find the JSON model file for this service client, relative to project root. Example contents:
```json
{"path":"codegen/sdk-codegen/aws-models/s3.json"}
```

Sources/SmithyCodegenCLI/SmithyCodegenCLI.swift
```diff
@@ -0,0 +1,44 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import ArgumentParser
+import Foundation
+import SmithyCodegenCore
+
```
jbelkins: This is a very simple code generator that:
- Reads the Smithy JSON model file into memory.
- Generates a Swift source file with a public variable `modelCount: Int` that contains the number of bytes in the model file.
- Adds a simple doc comment to the variable.

While this example codegen is trivial and not useful in itself, it does prove that:
- A Swift build plugin code generator can read from our model files elsewhere in the project.
- It can generate a Swift file that is added to the target.
- Doc comments in generated code get picked up with the rest of the target's doc comments.

Follow-on development would code-generate more substantial content from the model.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/DirectedSwiftCodegen.kt
```diff
@@ -107,6 +107,9 @@ class DirectedSwiftCodegen(
                 DependencyJSONGenerator(ctx).writePackageJSON(writers.dependencies)
             }
 
+            LOGGER.info("Generating Smithy model file info")
+            SmithyModelFileInfoGenerator(ctx).writeSmithyModelFileInfo()
+
```
jbelkins: Handles generation of the `smithy-model-info.json` file.

PR #999, opened 2025-11-15
feat: Create code generator build tool plugin

jbelkins:
## Description of changes
This PR provides a build plugin that will be used to code generate using a Swift-native code generator.
(An empty Swift file is produced as the code generator output.  Actually generating code will be a follow-on project.)

Changes:
- Write a `smithy-model-info.json` file to the project which is an input to the code generator.
- Define a build plugin (`SmithyCodeGenerator`) that will insert code generation into the target's Swift build process.
- Define a CLI tool (`SmithyCodegenCLI`) which will be invoked to perform code generation. (Actual codegen is unimplemented, this PR just establishes the structure.)

## Scope

### DIFF
--- a/Sources/SmithyCodegenCLI/SmithyCodegenCLI.swift
+++ b/Sources/SmithyCodegenCLI/SmithyCodegenCLI.swift
@@ -30,10 +30,13 @@ struct SmithyCodegenCLI: AsyncParsableCommand {
 
         // If --schemas-path was supplied, create the schema file URL
         let schemasFileURL = resolve(paramName: "--schemas-path", path: schemasPath)
-        print("Schemas file path: \(schemasFileURL?.path ?? "nil")")
 
         // All file URLs needed for code generation have now been resolved.
         // Implement code generation here.
+        if let schemasFileURL {
+            print("Schemas file path: \(schemasFileURL)")
+            FileManager.default.createFile(atPath: schemasFileURL.path, contents: Data())
+        }
     }
 
     private func currentWorkingDirectoryFileURL() -> URL {


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SmithyModelFileInfoGenerator.kt
```diff
@@ -0,0 +1,25 @@
+package software.amazon.smithy.swift.codegen
+
+import software.amazon.smithy.aws.traits.ServiceTrait
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.model.getTrait
+
```
jbelkins: This file generates a file named `smithy-model-info.json` into a service client target.

Its contents are
```json
{"path":"relative/path/to/model.json"}
```
where `relative/path/to/model.json` is the path to the JSON AST file for this service, relative to project root.

The Swift code generator will use this file to locate the model at build time.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/DirectedSwiftCodegen.kt
```diff
@@ -107,6 +107,9 @@ class DirectedSwiftCodegen(
                 DependencyJSONGenerator(ctx).writePackageJSON(writers.dependencies)
             }
 
+            LOGGER.info("Generating Smithy model file info")
+            SmithyModelFileInfoGenerator(ctx).writeSmithyModelFileInfo()
```
jbelkins: Invokes the model file info generator (see generator implementation immediately below.)

Plugins/SmithyCodeGenerator/SmithyCodeGeneratorPlugin.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+
+import struct Foundation.Data
+import class Foundation.FileManager
+import class Foundation.JSONDecoder
+import struct Foundation.URL
+import PackagePlugin
+
+@main
+struct SmithyCodeGeneratorPlugin: BuildToolPlugin {
```
jbelkins: This is the build tool plugin implementation.
- Reads the `smithy-model-info.json` file to get the model location in the form of a file URL.  (All other source files in the target are ignored.)
- Constructs a build command that uses the `SmithyCodegenCLI` tool to read the model & generate a schemas file from it.

PR #1000, opened 2025-11-17
chore: Opt-in service client tests

jbelkins:
## Description of changes
These changes support moving service clients from being committed to the project to being an opt-in feature to be utilized only during development & distribution.
- Move generation of the `Dependencies.json` to aws-sdk-swift since this file is only used by that project
- Package manifest is now generated for all clients.  (Rather than disable its generation, aws-sdk-swift can just not copy it into place)
- Package manifest generation is moved out of the protocol generator codegen because it is not specific to a protocol.
- Remove the `SwiftSettings` fields associated with `Package.swift` and `Dependencies.json` since they are no longer needed.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/SwiftSettings.kt
```diff
@@ -67,8 +65,6 @@ class SwiftSettings(
     val copyrightNotice: String,
     val visibility: String,
     val internalClient: Boolean,
-    val generatePackageManifest: Boolean,
-    val generateDependencyJSON: Boolean,
```
jbelkins: These two `SwiftSettings` fields are no longer used since the package manifest is now always generated, and the `Dependencies.json` file generation has been moved to SDK.

smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/manifestanddocs/PackageManifestGeneratorTests.kt
```diff
@@ -98,15 +98,15 @@ class PackageManifestGeneratorTests {
...
-        context.generationCtx.delegator.useFileWriter("xyz.swift") { writer ->
+        testContext.generationCtx.delegator.useFileWriter("xyz.swift") { writer ->
             writer.addDependency(SwiftDependency.CLIENT_RUNTIME)
         }
-        PackageManifestGenerator(context.generationCtx).writePackageManifest(context.generationCtx.delegator.dependencies)
-        context.generationCtx.delegator.flushWriters()
-        return context
+        PackageManifestGenerator(testContext.context).writePackageManifest(testContext.generationCtx.delegator.dependencies)
+        testContext.generationCtx.delegator.flushWriters()
+        return testContext
```
jbelkins: Test setup modified to create a `GenerationContext` instead of a `ProtocolGenerator.GenerationContext` to match the changes in `PackageManifestGenerator`.

dayaffe (approved):

PR #1002, opened 2025-11-20
Refactor NIOHTTPClient to SwiftNIOHTTPClient in SmithySwiftNIO

dayaffe:
## Scope

### DIFF
--- a/Sources/ClientRuntime/Telemetry/DefaultTelemetry.swift
+++ b/Sources/ClientRuntime/Telemetry/DefaultTelemetry.swift
@@ -1,0 +1,12 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import SmithyTelemetryAPI
+
+/// Typealias for backward compatibility.
+/// The actual implementation is now in SmithyTelemetryAPI.
+public typealias DefaultTelemetry = SmithyTelemetryAPI.DefaultTelemetry


Sources/SmithyHTTPClientAPI/HttpClientConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+import struct Foundation.TimeInterval
+import enum Smithy.URIScheme
+import struct SmithyHTTPAPI.Headers
+
+public class HttpClientConfiguration {
```
jbelkins: nit: `HTTPClientConfiguration` (note the caps)

As long as we're moving this to a new module, might as well make the naming consistent.

Sources/SmithyTelemetry/Context/TelemetryContext.swift
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
```
jbelkins: Presuming these `Sources/SmithyTelemetry` files are all largely identical to the ones that were in `ClientRuntime`?

Package.swift
```diff
@@ -125,6 +144,19 @@ let package = Package(
                 .copy("PrivacyInfo.xcprivacy")
             ]
         ),
+        .target(
+            name: "SmithySwiftNIO",
+            dependencies: [
+                "Smithy",
+                "SmithyHTTPAPI",
+                "SmithyHTTPClient",
```
jbelkins: `SmithyHTTPClient` depends on CRT, which means that `SmithySwiftNIO` depends on CRT.

We need to find a way to break that dependency.

Sources/ClientRuntime/Telemetry/DefaultTelemetry.swift
```diff
@@ -9,6 +9,7 @@ import struct Smithy.AttributeKey
 import struct Smithy.Attributes
 import protocol Smithy.LogAgent
 import struct Smithy.SwiftLogger
+import SmithyTelemetry
```
jbelkins: Move `DefaultTelemetry` from `ClientRuntime` to `SmithyTelemetry`, then `typealias` it here to prevent the breaking change.

Even though it is technically an "implementation", it has no dependencies other than `Smithy` and `SmithyTelemetry` so `SmithyTelemetry` is the right place for it.

Also, consider renaming `SmithyTelemetry` to `SmithyTelemetryAPI`, which would follow the SRA naming pattern for packages that contain interfaces but not implementations (at least not implementations with dependencies.)

jbelkins (changes requested):
See recommended adjustments to module membership.  Also let's try to see if we can make the SwiftNIO client not depend on CRT.

jbelkins (approved):

PR #1003, opened 2025-11-24
chore!: Move CRT dependency from SmithyHTTPAPI to ClientRuntime

jbelkins:
## Description of changes
`SmithyHTTPAPI` has a dependency on CRT `AwsCommonRuntimeKit` because it includes code for converting endpoint properties to/from the CRT types.

That code is moved to `ClientRuntime` with the`DefaultEndpointResolver` that uses it, and the dependency of `SmithyHTTPAPI` on `AwsCommonRuntimeKit` is removed.

**This is a breaking change** due to moving public components from one module to the other, but should be zero impact on customers since only `DefaultEndpointResolver` (located in `ClientRuntime`) calls the moved methods, and it is unlikely customers ever interact with them.

## Scope


Sources/ClientRuntime/Networking/Http/CRT/Endpoint+CRT.swift
```diff
@@ -0,0 +1,43 @@
...
+import enum Smithy.ClientError
+import struct SmithyHTTPAPI.Endpoint
+import enum SmithyHTTPAPI.EndpointPropertyValue
+import struct SmithyHTTPAPI.Headers
+
+extension Endpoint {
+
+    public init(urlString: String,
+                headers: Headers = Headers(),
+                endpointProperties: [String: EndpointProperty]) throws {
```
jbelkins: The initializer above is only called by `DefaultEndpointResolver` so it is moved here from `SmithyHTTPAPI`.

Package.swift
```diff
@@ -177,10 +177,7 @@ let package = Package(
         ),
         .target(
             name: "SmithyHTTPAPI",
-            dependencies: [
-                "Smithy",
-                .product(name: "AwsCommonRuntimeKit", package: "aws-crt-swift")
-            ]
+            dependencies: ["Smithy"]
```
jbelkins: CRT dependency is removed from `SmithyHTTPAPI` since no remaining types in `SmithyHTTPAPI` reference it.

dayaffe (approved):

PR #1011, opened 2025-12-17
feat: Add generation context

jbelkins:
## Description of changes
- Adds `Model` and the code to create it from the AST
- Adds `SymbolProvider`, to convert Smithy shapes to their Swift equivalents
- Adds `GenerationContext` to hold references to codegen components
- Deletes `ASTNode`.  `Smithy.Node` is used in AST instead (with a `Decodable` extension) since it is entirely the same type.

## Scope


Sources/SmithyCodegenCore/AST/Node+AST.swift
```diff
@@ -5,24 +5,11 @@
...
-/// See the definition of node value in the Smithy spec: https://smithy.io/2.0/spec/model.html#node-values
-enum ASTNode {
-    case object([String: ASTNode])
-    case list([ASTNode])
-    case string(String)
-    case number(Double)
-    case boolean(Bool)
-    case null
-}
+import enum Smithy.Node
```
jbelkins: `ASTNode` is deleted & `Smithy.Node` is used instead.

The `Decodable` extension below is applied to `Smithy.Node` instead of to `ASTNode`.

Sources/SmithyCodegenCore/Model/Model+AST.swift
```diff
@@ -0,0 +1,164 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import enum Smithy.Node
+import struct Smithy.ShapeID
+
+extension Model {
```
jbelkins: Here, an extension on `Model` is provided to create one from JSON AST.

See the ample inline comments below for details.

Sources/SmithyCodegenCore/Shape/EnumShape.swift
```diff
@@ -0,0 +1,28 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import enum Smithy.Node
+import struct Smithy.ShapeID
+import enum Smithy.ShapeType
```
jbelkins: Here & below are specialized Shape types for use in the Model.

All shapes descend from a `Shape` base class, similar to in Smithy codegen.

PR #1018, opened 2026-01-29
chore: Use client to perform HTTP response tests

jbelkins:
## Description of changes
Makes HTTP response & error protocol tests independent of internal implementation by using a specially configured client to perform the tests, similar to what we already do for request tests.

This change will allow schema-based services and services using current ReadWrite serde to use the same protocol tests, since the test no longer utilizes client internals.

Specific changes:
- Client is created with a special HTTP client for response tests, that returns the test's `httpResponse` when called.
- Similar to request tests, client uses dummy credential and endpoint resolution.
- Input is populated to the extent needed to allow request creation to succeed.
- Response & error tests are refactored to share setup.

Also:
- Clients initialize CRT at the start of client initialization, to ensure that CRT is initialized before any CRT functions are called.  (This has been an ongoing problem when clients are configured with non-default components; performing initialization in the client initializer should remedy this for good.)

## Scope

### DIFF
--- a/smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/HttpProtocolClientGeneratorTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/HttpProtocolClientGeneratorTests.kt
@@ -25,6 +25,7 @@ public class RestJsonProtocolClient: ClientRuntime.Client {
     let serviceName = "Rest Json Protocol"
 
     public required init(config: RestJsonProtocolClient.RestJsonProtocolClientConfiguration) {
+        ClientRuntime.initialize()
         client = ClientRuntime.SdkHttpClient(engine: config.httpClientEngine, config: config.httpClientConfiguration)
         self.config = config
     }
--- a/smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/HttpProtocolUnitTestErrorGeneratorTests.kt
+++ b/smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/HttpProtocolUnitTestErrorGeneratorTests.kt
@@ -15,46 +15,61 @@ class HttpProtocolUnitTestErrorGeneratorTests : HttpProtocolUnitTestResponseGene
         contents.shouldSyntacticSanityCheck()
         val expectedContents = """
     func testRestJsonComplexErrorWithNoMessage() async throws {
-        do {
-            guard let httpResponse = buildHttpResponse(
-                code: 403,
-                headers: [
-                    "Content-Type": "application/json",
-                    "X-Amzn-Errortype": "ComplexError",
-                    "X-Header": "Header"
-                ],
-                content: .data(Data(""${'"'}
-                {
-                    "TopLevel": "Top level",
-                    "Nested": {
-                        "Fooooo": "bar"
-                    }
+        guard let httpResponse = buildHttpResponse(
+            code: 403,
+            headers: [
+                "Content-Type": "application/json",
+                "X-Amzn-Errortype": "ComplexError",
+                "X-Header": "Header"
+            ],
+            content: .data(Data(""${'"'}
+            {
+                "TopLevel": "Top level",
+                "Nested": {
+                    "Fooooo": "bar"
                 }
-                ""${'"'}.utf8))
-            ) else {
-                XCTFail("Something is wrong with the created http response")
-                return
             }
+            ""${'"'}.utf8))
+        ) else {
+            XCTFail("Something is wrong with the created http response")
+            return
+        }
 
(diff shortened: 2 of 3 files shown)


Sources/ClientRuntime/Util/Initialize.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+
+import struct AwsCommonRuntimeKit.CommonRuntimeKit
+
+/// Sets up CRT-related shared resources such as the global allocator, event loops, etc.
+///
+/// Calls to CRT functions may crash the SDK if `CommonRuntimeKit.initialize()` is not called first.
+///
+/// This function may safely be called multiple times.
+public func initialize() { CommonRuntimeKit.initialize() }
```
jbelkins: Some protocol tests were failing because they were calling CRT before it had been initialized.  We have seen this problem crop up from time to time when customers swap out our CRT-based implementations of things for mocks or for their own.

The issue is that we typically initialize CRT at the point of use, but not all points of use.  So if we remove a config component that is initializing CRT, then the first call might get made from another component that doesn't initialize CRT but was depending on the removed one to do it.  That results in a crash.

To fix this problem once & for all, we'll be calling this during initialization of any client, so that we're assured CRT is already initialized before any operation is started.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestErrorGenerator.kt
```diff
@@ -10,59 +10,47 @@ import software.amazon.smithy.protocol.traits.Rpcv2CborTrait
...
+import software.amazon.smithy.swift.codegen.model.toLowerCamelCase
 import software.amazon.smithy.swift.codegen.swiftmodules.SmithyHTTPAPITypes
+import software.amazon.smithy.swift.codegen.swiftmodules.SwiftTypes
 import java.util.Base64
 
 open class HttpProtocolUnitTestErrorGenerator protected constructor(
     builder: Builder,
 ) : HttpProtocolUnitTestResponseGenerator(builder) {
     val error: Shape = builder.error ?: throw CodegenException("builder did not set an error shape")
-    override val outputShape: Shape? = error
```
jbelkins: This override is confusing because the error is not an output shape; these terms have specific meaning.  Better to just use the `error` property (just above) directly.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestErrorGenerator.kt
```diff
@@ -10,59 +10,47 @@ import software.amazon.smithy.protocol.traits.Rpcv2CborTrait
...
     private fun renderCompareActualAndExpectedErrors(
         test: HttpResponseTestCase,
         errorShape: Shape,
-        operationErrorType: String,
     ) {
-        val operationErrorVariableName = operationErrorType.replaceFirstChar { it.lowercase() }
         val errorType = symbolProvider.toSymbol(errorShape).name
 
-        writer.openBlock("if let actual = \$L as? \$L {", "} else {", operationErrorVariableName, errorType) {
+        writer.openBlock("if let actual = operationError as? \$L {", "} else {", errorType) {
```
jbelkins: Here & elsewhere in this file, `operationErrorVariableName` is deleted and the name `operationError` is used in all cases.

Trying to use a different name per-test for the variable holding the error adds complexity for no gain.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestResponseGenerator.kt
```diff
@@ -163,9 +175,69 @@ open class HttpProtocolUnitTestResponseGenerator protected constructor(
         }
     }
 
-    private fun renderActualOutput(outputStruct: Symbol) {
-        val responseClosure = ResponseClosureUtils(ctx, writer, operation).render()
-        writer.write("let actual: \$N = try await \$L(httpResponse)", outputStruct, responseClosure)
```
jbelkins: The code above accessed client internals to process the HTTP response into an output.

It is replaced with a specially configured client, see below.

smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/HttpProtocolClientGeneratorTests.kt
```diff
@@ -25,6 +25,7 @@ public class RestJsonProtocolClient: ClientRuntime.Client {
     let serviceName = "Rest Json Protocol"
 
     public required init(config: RestJsonProtocolClient.RestJsonProtocolClientConfiguration) {
+        ClientRuntime.initialize()
```
jbelkins: Here & below are codegen test changes.

PR #1025, opened 2026-02-10
chore: Zero retry delay for response/error protocol tests

jbelkins:
## Description of changes
https://github.com/smithy-lang/smithy-swift/pull/1018 recently changed response & error protocol tests to use a client (with the appropriate mocks installed) to perform those protocol tests.

It was discovered that the error tests now include delay time associated with retry, which is irrelevant for the purposes of these tests and just delays their completion.  This change adds special retry strategy options to the client that set the delay to zero so that tests can finish immediately.

## Scope


Sources/SmithyTestUtil/ResponseTestUtil/ProtocolTestRetryStrategyOptions.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import struct Foundation.TimeInterval
+import struct SmithyRetriesAPI.RetryStrategyOptions
+import protocol SmithyRetriesAPI.RetryBackoffStrategy
```
jbelkins: Below are the retry strategy options to be used for protocol tests.

They're default settings, except that the delay is 0 seconds in all cases.

dayaffe (approved):

PR #1042, opened 2026-03-25
feat: Schema-based description & debug description for structs & unions

jbelkins:
## Description of changes

Note: there is a companion SDK branch https://github.com/awslabs/aws-sdk-swift/tree/jbe/schema_based_description that is the same content as the epic branch, so that this branch will build on CI.  There is no associated SDK PR.

- Provides a `StringSerializer` for printing a `SerializableShape` to a string, and redacting fields marked with the `smithy.api#sensitive` trait.  Behavior is checked with tests.
- Extends `SerializableStruct` with `CustomStringConvertible` and `CustomDebugStringConvertible` conformance, using the `StringSerializer` to create the (debug) description.
- Disables code-generation of `CustomDebugStringConvertible` conformance for schema-based clients

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/DirectedSwiftCodegen.kt
```diff
@@ -121,9 +122,11 @@ class DirectedSwiftCodegen(
...
         }
 
-        if (shape.hasTrait<SensitiveTrait>() || shape.members().any { it.isSensitive(model) }) {
-            writers.useShapeExtensionWriter(shape, "CustomDebugStringConvertible") { writer: SwiftWriter ->
-                CustomDebugStringConvertibleGenerator(symbolProvider, writer, shape, model).render()
+        if (!SerdeUtils.useSchemaBased(settings, model)) {
+            if (shape.hasTrait<SensitiveTrait>() || shape.members().any { it.isSensitive(model) }) {
+                writers.useShapeExtensionWriter(shape, "CustomDebugStringConvertible") { writer: SwiftWriter ->
+                    CustomDebugStringConvertibleGenerator(symbolProvider, writer, shape, model).render()
+                }
```
jbelkins: Don't code-generate a `CustomDebugStringConvertible` extension when the service is schema-based, since the extension on `SerializableStruct` will be used instead.

Sources/SmithySerialization/StringSerializer.swift
```diff
@@ -0,0 +1,157 @@
...
+import struct Smithy.SensitiveTrait
+import struct Smithy.ShapeID
+import protocol Smithy.SmithyDocument
+
+public class StringSerializer: ShapeSerializer {
+    public private(set) var string = ""
+    private var isFirstElement = true
+
+    private let interstitial = ", "
+    private let redacted = "[REDACTED]"
```
dayaffe: I think we currently use `CONTENT_REDACTED` should we keep that for consistency or was this intenally changed

Sources/SmithySerialization/Serialization/SerializableStruct.swift
```diff
@@ -5,6 +5,29 @@
...
+public extension SerializableStruct {
+
+    /// A written description of this type and its contents.
+    ///
+    /// Fields marked with the `sensitive` trait will be written to the description as "redacted".
+    var description: String {
+        let serializer = StringSerializer()
+        // Safe to try! here because StringSerializer never throws
+        // swiftlint:disable:next force_try
+        try! serialize(serializer)
```
dayaffe: in the future an update to the string serializer could cause it to throw? consider wrapping in a do/catch anyway but wont block on this

Sources/SmithySerialization/StringSerializer.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public func writeBlob(_ schema: Schema, _ value: Data) throws {
+        guard !isSensitive(schema) else { return }
+        string += "<\(value.count) bytes>"
+    }
+
+    public func writeTimestamp(_ schema: Schema, _ value: Date) throws {
+        guard !isSensitive(schema) else { return }
+        string += ISO8601DateFormatter().string(from: value)
```
dayaffe, line 125: nit: this would create a new formatter on each call consider creating a private static version that gets reused

dayaffe (approved):
Left a few comments to consider but nothing I feel we need to block on

Sources/SmithySerialization/StringSerializer.swift
jbelkins: You are correct, I will change the text used here to `CONTENT_REDACTED` as well.

Sources/SmithySerialization/Serialization/SerializableStruct.swift
jbelkins: I can put a try-catch here that logs some text asking the reader to file a bug.

Sources/SmithySerialization/StringSerializer.swift
jbelkins: https://developer.apple.com/documentation/foundation/iso8601dateformatter

I wrote it without a static date formatter because the `ISO8601DateFormatter` type is not `Sendable`, and a static value would break Swift 6 Language Mode compatibility.

jbelkins: Apple suggests using this type as an alternative, which is Sendable:
https://developer.apple.com/documentation/foundation/date/iso8601formatstyle

but it's not supported on our minimum platforms.

PR #1047, opened 2026-04-02
feat: schema based XML

dayaffe:
## Scope


Sources/SmithyEventStreams/Deserializers/EventContentDeserializer.swift
```diff
@@ -32,11 +32,17 @@ struct EventContentDeserializer: ShapeDeserializer {
 
         // Deserialize the event payload, to the member marked with @eventPayload if it exists,
         // to the structure's members otherwise.
-        // Use a deserializer for the protocol in use, by making it from the codec.
-        let payloadDeserializer = try codec.makeDeserializer(data: message.payload)
         if let payloadMember = schema.members.first(where: { $0.hasTrait(EventPayloadTrait.self) }) {
-            try T.readConsumer(payloadMember, &value, payloadDeserializer)
```
jbelkins, line 38: Per the spec for event payload, it can be a blob, string, structure, or union.
https://smithy.io/2.0/spec/streaming.html#eventpayload-trait

We should probably just refactor this to do all 4.

Sources/SmithyRestXML/TimestampUtils.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+@_spi(SchemaBasedSerde)
+import struct Smithy.Schema
+@_spi(SchemaBasedSerde)
+import struct Smithy.TimestampFormatTrait
+@_spi(SmithyTimestamps) import enum SmithyTimestamps.TimestampFormat
+
+func resolveTimestampFormat(_ schema: Schema) -> TimestampFormat {
+    guard let traitFormat = try? schema.getTrait(TimestampFormatTrait.self)?.format else {
```
jbelkins, line 15: what does `try?` buy us here?  Seems we should throw an exception if it happens; it would be highly unusual for this to throw

Sources/SmithyRestXML/Deserializer.swift
```diff
@@ -0,0 +1,411 @@
...
+import struct SmithySerialization.SerializerError
+@_spi(SchemaBasedSerde)
+import protocol SmithySerialization.ShapeDeserializer
+@_spi(SmithyTimestamps) import enum SmithyTimestamps.TimestampFormat
+@_spi(SmithyTimestamps) import struct SmithyTimestamps.TimestampFormatter
+@_spi(SmithyReadWrite) import struct SmithyXML.NodeInfo
+@_spi(SmithyReadWrite) import class SmithyXML.Reader
+
+@_spi(SchemaBasedSerde)
+public struct Deserializer: ShapeDeserializer {
```
jbelkins, line 47: This type is created in the RestXML module.

We'll have AwsQuery/Ec2Query as well, which also use XML in responses.  Can this be generalized for other formats?

Sources/Smithy/TraitLibrary/XmlNamespaceTrait.swift
```diff
@@ -0,0 +1,33 @@
...
+        }
+        guard case .string(let uri) = dict["uri"] else {
+            throw TraitError("xmlNamespace trait requires a 'uri' string")
+        }
+        self.uri = uri
+        if case .string(let prefix) = dict["prefix"] {
+            self.prefix = prefix
+        } else {
+            self.prefix = nil
+        }
```
jbelkins, line 31: A simpler way:
```suggestion
        self.prefix = dict["prefix"]?.string
```

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PackageManifestGenerator.kt
```diff
@@ -48,8 +50,20 @@ class PackageManifestGenerator(
...
+                            writer.openBlock("plugins: [", "]") {
+                                writer.openBlock(".plugin(", ")") {
+                                    writer.write("name: \"SmithyCodeGeneratorPlugin\",")
+                                    writer.write("package: \"smithy-swift\"")
+                                }
+                            }
+                        } else {
+                            writer.openBlock("dependencies: [", "]") {
+                                dependenciesByTarget.forEach { writeTargetDependency(writer, it) }
+                            }
```
jbelkins, line 66: Probably okay to write the plugin to the manifest for all protocols, including not yet schema-based.  Non-schema-based protocols don't get a `swift-settings.json` so the plugin just exits early.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HTTPBindingProtocolGenerator.kt
```diff
@@ -149,7 +150,7 @@ abstract class HTTPBindingProtocolGenerator(
                     continue
                 }
                 val httpBindingResolver = getProtocolHttpBindingResolver(ctx, defaultContentType)
-                if (!usesSchemaBased) {
+                if (!usesSchemaBased || ctx.service.hasTrait<RestXmlTrait>()) {
```
jbelkins, line 153: Is this condition needed?  `usesSchemaBased` should be true for RestXML anyway

jbelkins (changes requested):

PR #1078, opened 2026-05-22
feat: Serde benchmark testing

jbelkins:
## Description of changes
Adds a specialization of the protocol test suite for testing serialization performance.  "Serde benchmark" tests are defined as either HTTP request or HTTP response protocol tests, and are identified by the `serde-benchmark` tag applied to the test.

When a serde benchmark test is to be generated, the usual request / response protocol test body is modified to perform testing.

Specific changes:
- Special models for testing various size payloads are used to generate tests (note that these are unchanged from the cross-platform spec.)
- A special serde benchmark telemetry provider is installed to measure time taken during serialization.
- Times to perform request serialization & response deserialization are measured repeatedly, and the results are transformed into measurements (i.e. mean, deviation, p50/90/99/etc.)
- Generates a JSON report from collected measurements

Changes to protocol tests generally:
- When `Data` is to be constructed as an input or expectation of a test, Swift "raw strings" are used, and the string is no longer rendered with any escaping.
- Format and spacing changes

**Note:** companion PR https://github.com/awslabs/aws-sdk-swift/pull/2172 has no code changes and is opened only to run SDK testing with this change applied.

## Scope


smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/config/DefaultClientConfiguration.kt
```diff
@@ -30,6 +30,20 @@ class DefaultClientConfiguration : ClientConfiguration {
...
+                false,
+                false,
+            ),
+            ConfigProperty(
+                "signingRegion",
+                SwiftTypes.String.toOptional(),
+                { "self.region" },
+                false,
+                false,
+            ),
```
jbelkins, line 46: `region` and `signingRegion` are added to non-AWS, smithy-swift-based clients because region & signing region are needed to configure `sigv4` which is a Smithy-native auth scheme.

The properties are optional, so they may be omitted when configuring clients to use other auth schemes.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -72,15 +73,15 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(
...
+                                test.body.get(),
                             )
                         }
                     } else {
                         // Default case for non-CBOR protocols
                         writer.format(
-                            "Data(\"\"\"\n\$L\n\"\"\".utf8)",
-                            test.body.get().replace("\\\"", "\\\\\""),
+                            "Data(#\"\"\"\n\$L\n\"\"\"#.utf8)",
+                            test.body.get(),
```
jbelkins, line 86: Here, we use Swift "raw strings" (i.e. `#""" ... """#`) to include string literal content exactly as written, and we remove the backslash escaping.

smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/integration/HttpProtocolUnitTestRequestGenerator.kt
```diff
@@ -118,13 +120,18 @@ open class HttpProtocolUnitTestRequestGenerator protected constructor(
                     url,
                 )
             }
+            writer.write("telemetryProvider: telemetryProvider,")
```
jbelkins: The serde benchmark telemetry provider, which takes the performance measurements, is installed in serde benchmark tests and in protocol tests alike, though it has no effect when running protocol tests.

dayaffe (approved):

PR #1081, opened 2026-05-25
perf: Optimize JSON serialization

jbelkins:
## Description of changes
Our current JSON serializer builds a DOM model of the JSON to be encoded as it traverses the input, then serializes that to JSON once the entire input data structure has been traversed.  While this approach results in simple serialization code and correct behavior, its performance is deficient because it involves many steps that are ultimately unnecessary.

This rewrite of the JSON serializer serializes directly into JSON, encoded in UTF-8, as the input structure is traversed.  Serde performance is 2-3x higher for serialization of JSON payloads, and substantially higher than that for certain data types such as blobs.

Document type modeling, both during codegen and at runtime, has been changed to better utilize existing serializer components for document content.

Specific optimizations:
- The tree structure that represents the JSON data which we used to build is no longer built; instead, UTF-8 bytes are written directly into data as the input is traversed by the serializer.  This eliminates two entire phases of operation from the previous serializer: building a tree of JSON values, and converting the tree into a Foundation object for input to the Foundation `NSJSONSerialization` serializer.
- Previously, a new serializer was constructed for sub-elements.  Now, the same serializer is used for the entire tree, reducing allocations.  State (for whether or not a comma is needed to separate from the previous element) is saved in local vars before, and restored after traversing structures & collections.

Also adds a wide variety of tests for proper JSON serialization.

## Scope


Sources/Smithy/Schema/Schema.swift
```diff
@@ -36,7 +36,7 @@ public final class Schema: Sendable {
     ///
     /// When this schema is itself a Member, it returns the target's members.
     public var members: [Schema] {
-        _containerType != nil ? _target()!._members : _members
+        containerType != nil ? _target()!._members : _members
```
dayaffe, line 39: Something to check but I think a test like this would fail where there is a nested List/Map document:

```
  func test_writeDocument_writesANestedListDocument() throws {
      let subject = SmithyJSON.Serializer(usesJSONNameTrait: false)
      let document = Document(ListDocument(value: [
          ListDocument(value: [StringDocument(value: "x")]),   // list-of-list
      ]))
      try subject.writeDocument(Smithy.Prelude.documentSchema, document)
      let json = String(data: try subject.data, encoding: .utf8)
      XCTAssertEqual(json, #"[["x"]]"#)   // currently traps instead of asserting
  }
```

If that's not a valid case then no worries

Sources/SmithyJSON/Serializer.swift
```diff
@@ -26,139 +29,289 @@ import protocol SmithySerialization.ShapeSerializer
...
+            case Self.tab:
+                _data.append(Self.backslash)
+                _data.append(Self.t)
+            default:
+                if let ascii {
+                    _data.append(ascii)
+                } else {
+                    _data.append(contentsOf: character.utf8)
+                }
+            }
```
dayaffe: I think the previously serializer caught additional control values for us behind the scenes. Consider changing to:

```
default:
      if let ascii, ascii < 0x20 {
          // Any C0 control without a short form must be \u00XX-escaped (RFC 8259)
          _data.append(contentsOf: "\\u00".utf8)
          _data.append(hexDigit(ascii >> 4))
          _data.append(hexDigit(ascii & 0x0F))
      } else if let ascii {
          _data.append(ascii)
      } else {
          _data.append(contentsOf: character.utf8)
      }
```

dayaffe (changes requested):
2 potential changes requested

Sources/Smithy/Schema/Schema.swift
jbelkins: I added codegen tests for nested list and nested map.  They pass.  I know this code looks weird for that case but it's good recursively in general.

Sources/SmithyJSON/Serializer.swift
jbelkins: Added code (plus a test) to escape the non-special escape characters as `\u00hh`.

dayaffe (approved):

PR #1086, opened 2026-05-28
fix: Limit CBOR decoding depth to prevent stack overflow

jbelkins:
## Description of changes
Limit the nested depth to follow when decoding CBOR to 128 levels, and throw an error when that depth is exceeded.  This prevents malformed data from overflowing the call stack and crashing the application.

Depth checking is applied to structs, lists, and maps, both when being decoded to a value and when being "skipped" because the target data structure does not implement that field.

Tests show the proper operation of the depth limiting feature.

## Scope


dayaffe (approved):
LGTM, pulled locally to test and ensure it works

PR #1104, opened 2026-06-23
chore: revert "Convert Schema to class (#1082)"

jbelkins:
## Description of changes
This reverts commit 06fe84e845b3cbff4e2649397219dea21e8baad0.

## Scope
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1105, opened 2026-06-23
chore: Convert Schema to class (2nd attempt)

jbelkins:
## Description of changes
In the original PR https://github.com/smithy-lang/smithy-swift/pull/1082, schemas were converted from value types that are created on demand using computed properties, to lazy globals that are created and retained on first use, then reused.

However, the change started causing compiler OOMs in the build system.  That change was reverted in https://github.com/smithy-lang/smithy-swift/pull/1104 so that we could determine the cause of the OOMs and try the feature over again.

The cause was found to be that: member schemas were previously declared inline in the `members` array property of the parent schema.  Sometimes (i.e. in the case of enums with hundreds of cases), this resulted in code-generated Swift literal expressions that were thousands of lines long.  When these expressions were used in computed properties, they were handled by the compiler without problems.

However, when we instead tried to use these same large literal expressions as the RHS of an assignment, they started to consume a lot of time and RAM at type checking time.  It has been noted before that very large arrays of complex literal expressions (i.e. structures like our schemas) cause high times & memory usage to type-check.  See: https://github.com/swiftlang/swift/issues/60549

Member schemas are now declared as their own separate schemas, assigned to a variable in the global namespace, and included by reference in the parent schema's `members` instead of being declared inline.  This breaks up large composite schemas into much smaller expressions that are type-checked much easier.

Another change from the first attempt is that the schema global vars have explicit typing of `Swift.Schema`.  Not providing explicit type info causes the compiler to incorrectly detect a circular reference when schema targets form a reference cycle.  This discovery allows us to avoid the extra code-generated getters that we were using before, reducing the volume of code-generated expressions.  (In reality, the circular reference chain is broken by the use of `@autoclosure` on the schema's target property, so the compiler is correct to not flag the reference.)

## Scope


Sources/Smithy/Schema/Schema.swift
```diff
@@ -67,7 +67,7 @@ public struct Schema: Sendable {
         [ShapeType.structure, .union].contains(_containerType) ? id.member! : nil
     }
 
-    private var _containerType: ShapeType?
+    private let _containerType: ShapeType?
```
jbelkins, line 70: it was an oversight for this property to be mutable, changed to `let`

Sources/SmithyCodegenCore/Deserialize/DeserializeCodegen.swift
```diff
@@ -58,9 +51,12 @@ package struct DeserializeCodegen {
                 }
                 writer.write("")
                 let consumerType = "SmithySerialization.ReadStructConsumer<Self>"
+                let paramSwiftType = try ctx.symbolProvider.swiftType(shape: shape, forParamUse: true)
                 try writer.openBlock(
                     "public static var readConsumer: \(consumerType) {", "}") { writer in
-                    try writer.openBlock("{ memberSchema, \(varName), deserializer in", "}") { writer in
+                    try writer.openBlock("{ (memberSchema: Smithy.Schema, \(varName): inout \(paramSwiftType), " +
+                                           "deserializer: any SmithySerialization.ShapeDeserializer) throws -> Void in",
+                                         "}") { writer in
```
jbelkins: Explicit type added to this closure.

Sources/SmithyCodegenCore/Deserialize/DeserializeCodegen.swift
```diff
@@ -118,10 +114,12 @@ package struct DeserializeCodegen {
             guard let listShape = target as? ListShape else {
                 throw SymbolProviderError("Shape has type .list but is not a ListShape")
             }
+            let listMemberSwiftType = try ctx.symbolProvider.swiftType(shape: listShape.member.target)
             let isSparse = listShape.hasTrait(SparseTrait.self)
             let methodName = isSparse ? "readSparseList" : "readList"
             try writer.openBlock(
-                "let value: \(propertySwiftType) = try deserializer.\(methodName)(\(schemaVarName)) { deserializer in",
+                "let value: \(propertySwiftType) = try deserializer.\(methodName)(\(schemaVarName)) { (deserializer: " +
+                "any SmithySerialization.ShapeDeserializer) throws -> \(listMemberSwiftType) in",
```
jbelkins: Here & below, explicit type info added to the consumer closures for list & map elements, respectively.

Sources/SmithyCodegenCore/Schemas/SchemasCodegen.swift
```diff
@@ -44,25 +44,11 @@ package struct SchemasCodegen {
...
-            // will target the duplicate instead.
-            //
-            // This happens ~20 times in AWS models so it is not so frequent that the extra var will bloat
-            // service clients.
-            if let hm = shape as? HasMembers, try hm.members.contains(where: { $0.targetID == shape.id }) {
-                try writer.openBlock("var dup_of_\(shape.schemaVarName): Smithy.Schema {", "}") { writer in
-                    try writer.write(shape.schemaVarName)
-                }
-                writer.write("")
-            }
```
jbelkins, line 65: With reference-type schemas, Swift is okay with direct references to other schemas.  (Those references are already auto-closures, so that avoids infinite circular references.)

Sources/SmithyCodegenCore/Serialize/SerializeCodegen.swift
```diff
@@ -55,7 +50,10 @@ package struct SerializeCodegen {
                 try writer.openBlock(
                     "public static var writeConsumer: SmithySerialization.WriteStructConsumer<Self> {", "}"
                 ) { writer in
-                    try writer.openBlock("{ memberSchema, \(varName), serializer in", "}") { writer in
+                    let paramSwiftType = try ctx.symbolProvider.swiftType(shape: shape, forParamUse: true)
+                    try writer.openBlock("{ (memberSchema: Smithy.Schema, \(varName): \(paramSwiftType), " +
+                                            "serializer: any SmithySerialization.ShapeSerializer) throws -> Void in",
+                                         "}") { writer in
```
jbelkins: Type information added to this closure as well.

Sources/SmithyCodegenCore/Serialize/SerializeCodegen.swift
```diff
@@ -111,13 +118,16 @@ package struct SerializeCodegen {
             guard let listShape = target as? ListShape else {
                 throw ModelError("Shape \(target.id) is type .\(target.type) but not a ListShape")
             }
+            let listMemberSwiftType = try ctx.symbolProvider.swiftType(shape: listShape.member.target)
             let isSparse = listShape.hasTrait(SparseTrait.self)
             let methodName = isSparse ? "writeSparseList" : "writeList"
             try writer.openBlock(
-                "try serializer.\(methodName)(\(schemaVarName), value) { value, serializer in",
+                "try serializer.\(methodName)(\(schemaVarName), value) { (value: \(listMemberSwiftType), serializer: " +
+                "any SmithySerialization.ShapeSerializer) throws -> Void in",
```
jbelkins: Type info added to list member and map value closures here and below, respectively.

Sources/SmithyCodegenCore/SymbolProvider/SymbolProvider.swift
```diff
@@ -81,7 +81,7 @@ public struct SymbolProvider {
         case .timestamp:
             return "Foundation.Date"
         case .document:
-            return "Smithy.Document"
+            return forParamUse ? "(any Smithy.SmithyDocument)" : "Smithy.Document"
```
jbelkins: When used as a parameter, the protocol type `SmithyDocument` is used.

In other contexts, the concrete document wrapper type `Document` is used. (This is the type of a document on our Swift models.)

dayaffe (approved):

PR #1110, opened 2026-06-30
chore: Use typed traits at runtime

jbelkins:
## Description of changes
Currently, during both codegen and at runtime, traits for a model or a schema (respectively) are stored as a dictionary of `[ShapeID: Node]`.  When a certain type of trait is needed, the node is retrieved by the trait ID and the trait is constructed from the node and returned to the caller.  While this approach is simple, it has some drawbacks:
- Many traits never get created until runtime, meaning that if the trait node contains invalid data, then the resulting error will first get detected when the trait is used at runtime.
- Creating a trait when it is queried is slow and costly.
- Storing pre-initialized traits opens the door to performance optimizations and type-specific behavior that we can't reproduce with ShapeID & Node.

So, the `[ShapeID: Node]` type storage is changed to `[ShapeID: any Trait]`, allowing for the storage of previously created traits, keyed by their trait ID.

Specific changes made here:
- `TraitCollection` now stores a map of ShapeID to Traits instead of ShapeID to Node.
- Unused `TraitCollection` accessors have been deleted.
- `TraitCollection` accessors no longer throw since the throwing part (trait creation) no longer happens during access.  `try` & `throws` is removed numerous places as a result.
- Model now accepts an array of `Trait` types at creation.  The modeled Traits are created from AST, based on their trait ID.  Model is pre-loaded with the predefined trait types (i.e. the ones that smithy-swift already understands & uses).
- The `RuntimeTrait` "marker protocol" identifies traits that are used at runtime so that they may be copied into schemas.
- Schemas are code-generated with a literal array of traits instead of the previous `[ShapeID: Node]` dictionary.

## Scope


Sources/Smithy/Trait/TraitCollection.swift
```diff
@@ -7,16 +7,35 @@
...
+
+        // Finally, convert the ShapeID - modeled Trait pairs into a dictionary.
+        self.traitDict = Dictionary(uniqueKeysWithValues: traitPairs)
+    }
+
+    /// Creates a TraitCollection from an array of modeled Traits.
+    /// - Parameter traits: A list of modeled traits to store in this collection.  All traits must be unique by ShapeID.
+    public init(traits: [any Trait]) {
+        // Convert the modeled Trait list to a dictionary, mapped by trait ShapeID.
+        self.traitDict = Dictionary(uniqueKeysWithValues: traits.map { ($0.id, $0) })
```
dayaffe, line 38: nit: I don't think that we need all these comments, AI tends to comment on everything in order to explain to you what's going on even if the code is self-explanatory (ex. "Finally, convert .... into a dictionary")

Sources/SmithyCodegenCore/Schemas/SchemasCodegen.swift
```diff
@@ -78,14 +76,20 @@ package struct SchemasCodegen {
...
                 writer.openBlock("traits: [", "],") { writer in
-                    for (traitID, node) in traitPairs {
-                        writer.write("\(traitID.rendered): \(node.rendered),")
+                    for (TraitType, node) in traitPairs {
+                        // Traits are initialized using try? to convert any thrown error to nil.
+                        // In practice, a trait should never throw at runtime, because every
+                        // trait was already successfully constructed by the code generator
+                        // while processing the model.
+                        // The TraitCollection initializes from this array of optional Traits;
+                        // nil elements are simply compacted out.
```
dayaffe, line 91: that's a lot of comments for 1 line of code!

dayaffe (approved):

PR #1127, opened 2026-07-20
fix: Deterministic rendering of Node during codegen

RufusMall:
## Description of changes

I joined a new company last week, and I am working on a couple of apps that uses the amplify/aws iOS sdks. These use this library under the hood. Our app has quite large compile/incremental compile times relative to its size.

I can see that a bunch of files in the AWS SDK are recompiling even when nothing has changed. This bubbles up into amplify and portions of our app needlessly recompiling.

I have proposed a fix for this - which ensures the output of the SDK is deterministic - by simply ordering the keys in dictionaries when outputting them in the schema. Before the output would change each time causing the schema to change, and then cascading up the file/module dependancy tree.

Example Diff in the scheme:
Screenshot 2026-07-20 at 03 37 14

  Something like::
  AWSCognitoIdentitySchemas.swift (changes 1 line)
   
  Causes recompilation of:
  AWSCognitoIdentityOperations.swift
  AWSCognitoIdentitySerialize.swift
  AWSCognitoIdentityDeserialize.swift
  AWSCognitoIdentity module 
  AWSCognitoAuthPlugin 
etc

I made a minimal sample app which depends on Amplify to test this.

Before including the SDK (sample app with 0 dependancies):
```
======================================
Results:
======================================
Total time for 3 builds: 2.896397000s
Average build time: .96s
======================================

Including the amplify SDK - without my diff:

======================================
Results:
======================================
Total time for 3 builds: 71.689864000s
Average build time: 23.89s
======================================

Including the amplify SDK - with my sorted keys fix:
======================================
Results:
======================================
Total time for 3 builds: 42.986865000s
Average build time: 14.32s
======================================
```

As you can see the difference is quite significant. It's even bigger in the full production app since these changes cause a chain reaction in the main app.

Even though the output is functionally the same, the changed hash causes Xcode to recompile the file.

There is another "fix" I am testing, which also only writes files if the contents have changed (this PR is a prerequisite). This allows Xcode to bypass even more checks as it can see the timestamp will not be changed leading to further build improvements. This will be done as part of a separate PR though!

Note: I couldn't find instructions to run the unit test suite locally (test-sdks), some instructions on this would be greatly appreciated. 

I did update the tests to assert that the keys are always ordered, with a basic message so someone doesn't add the || behaviour back.

I am pretty new to these sdks and tools so appreciate the feedback :D.

## Scope

- [ x ] Bug fix (non-breaking change which fixes an issue)


Sources/SmithyCodegenCore/SwiftRendering/Node+Rendered.swift
RufusMall, line 1: Todo:// fix the commit message to match the contribution guidelines

RufusMall:
Hi, @jbelkins is there a doc somewhere on how to run these tests?

I want to run them a few thousand times to make sure I didn't miss something and introduce a bunch of flakiness. 

thanks.

trickart:
We hit this exact issue and independently root-caused it to the same line before finding this PR — posting our findings as additional confirmation.

**Environment**: Xcode 26.5 (17F42), macOS 26.5.2, amplify-swift 2.58.4 / aws-sdk-swift 1.7.27 / smithy-swift 0.223.0. App depends on AWSCognitoAuthPlugin + AWSS3StoragePlugin.

**Impact**: with zero source changes, every Xcode/xcodebuild build re-runs the 3 Smithy codegen commands (AWSCognitoIdentity, InternalAWSCognitoIdentity, AWSCognitoIdentityProvider) and recompiles ~86 Swift compile tasks cascading through AWSCognitoAuthPlugin into our app — adding **~85–90 seconds to every incremental build** (Apple Silicon).

We confirmed the nondeterminism by running the built `SmithyCodegenCLI` binary twice with identical arguments and diffing the outputs — only the `*Schemas.swift` files differ, and only in dictionary key order:

```diff
-        Smithy.ShapeID("aws.api", "service"): ["arnNamespace": "cognito-identity","cloudFormationName": "Cognito","cloudTrailEventSource": "cognito-identity.amazonaws.com","sdkId": "Cognito Identity","endpointPrefix": "cognito-identity"],
+        Smithy.ShapeID("aws.api", "service"): ["endpointPrefix": "cognito-identity","arnNamespace": "cognito-identity","cloudTrailEventSource": "cognito-identity.amazonaws.com","sdkId": "Cognito Identity","cloudFormationName": "Cognito"],
```

The plugin build command itself is byte-identical between builds (we diffed the full command line and exported environment from `xcodebuild` logs), so making the output deterministic as this PR does should restore incremental build caching. This currently costs our team ~90 s on every build — would love to see it merged.

RufusMall:
> 90

Nice!

The timings I posted in the PR description were from a minimal sample app.

it also costs us around 90 seconds in the app I use at work.

jbelkins:
@RufusMall @trickart 
I just returned from vacation today.  Please give me a couple days to dig out of my unread-email hole, and I will look at this in detail... thanks!

jbelkins:
@RufusMall @trickart I have looked into this issue in detail.  Here's my findings:

While it is intended that the generated files should be "deterministic", i.e. should be the same content byte-for-byte every time they are generated for the same inputs, this is not the actual cause of the constant rebuilding that you are observing.

In fact, it should not matter for incremental builds at all if the codegen output is different, because the build system should not even be attempting to regenerate the affected files in the first place.  And it's not the changed content that triggers rebuild, it's the changed file timestamps, which would happen whether or not the generated code is different.

The real cause is apparently that the Xcode build system has a bug which causes the `SmithyCodegenCLI` tool to be rebuilt on every build, even when its inputs do not change.  See this ticket, which describes the problem and notes that other Swift packages have experienced the same:
https://github.com/swiftlang/swift-build/issues/305

I have confirmed that frequent rebuilds do not occur on the non-Xcode Swift toolchain, bolstering the belief that this particular swift-build issue is the cause.  (I can't propose a fix for it either, because the cause is apparently in closed-source Xcode source code.)

So these are the actions I plan to take this week:
- I will merge this proposed PR after a couple mods, because it does correct the issue of deterministic code generation.
- I will modify the build plugin to mitigate the effect of the swift-build bug cited above.  I will probably do something similar to https://github.com/liamnichols/xcstrings-tool/pull/129 which addresses this issue but in a different package that utilizes a similar plugin.

(If you're an amplify-swift customer, I'll work with that team to get the fix published in an upcoming amplify-swift release once it's fixed here.)

RufusMall:
> @RufusMall @trickart I have looked into this issue in detail. Here's my findings:
> 
> While it is intended that the generated files should be "deterministic", i.e. should be the same content byte-for-byte every time they are generated for the same inputs, this is not the actual cause of the constant rebuilding that you are observing.
> 
> In fact, it should not matter for incremental builds at all if the codegen output is different, because the build system should not even be attempting to regenerate the affected files in the first place. And it's not the changed content that triggers rebuild, it's the changed file timestamps, which would happen whether or not the generated code is different.
> 
> The real cause is apparently that the Xcode build system has a bug which causes the `SmithyCodegenCLI` tool to be rebuilt on every build, even when its inputs do not change. See this ticket, which describes the problem and notes that other Swift packages have experienced the same: [swiftlang/swift-build#305](https://github.com/swiftlang/swift-build/issues/305)
> 
> I have confirmed that frequent rebuilds do not occur on the non-Xcode Swift toolchain, bolstering the belief that this particular swift-build issue is the cause. (I can't propose a fix for it either, because the cause is apparently in closed-source Xcode source code.)
> 
> So these are the actions I plan to take this week:
> 
> * I will merge this proposed PR after a couple mods, because it does correct the issue of deterministic code generation.
> * I will modify the build plugin to mitigate the effect of the swift-build bug cited above.  I will probably do something similar to [Support caching generated outputs to workaround incremental build bugs liamnichols/xcstrings-tool#129](https://github.com/liamnichols/xcstrings-tool/pull/129) which addresses this issue but in a different package that utilizes a similar plugin.
> 
> (If you're an amplify-swift customer, I'll work with that team to get the fix published in an upcoming amplify-swift release once it's fixed here.)


Interesting, my next change was going to be this [PR](https://github.com/RufusMall/smithy-swift/pull/2/changes) which further speeds things up due to the timestamp of the generated files no longer being updated. Is this similar to the additional change you were going to apply?

In my testing this PR alone was still a big improvement, because even if the file timestamp changes, Xcode eventually bypasses recompiling etc because the file hashes match (or something). After this second change (linked below) the file timestamps are also no longer updated so it speeds up the Xcode compilation even more.

https://github.com/RufusMall/smithy-swift/pull/2/changes


Nice, I was surprised that Xcode was running the plugin multiple times... interesting that the non-Xcode Swift toolchain doesn't have this bug...

Thanks for looking into this!

jbelkins:
Made the following changes:
- Slight refactoring of Node object rendering
- Added a space after the comma separating object elements in rendered Node objects
- Removed test modifications to JSON serialization tests, which are different rendering code from Node
- Added a test for ordered rendering of Node objects

RufusMall:
> > @RufusMall @trickart I have looked into this issue in detail. Here's my findings:
> > While it is intended that the generated files should be "deterministic", i.e. should be the same content byte-for-byte every time they are generated for the same inputs, this is not the actual cause of the constant rebuilding that you are observing.
> > In fact, it should not matter for incremental builds at all if the codegen output is different, because the build system should not even be attempting to regenerate the affected files in the first place. And it's not the changed content that triggers rebuild, it's the changed file timestamps, which would happen whether or not the generated code is different.
> > The real cause is apparently that the Xcode build system has a bug which causes the `SmithyCodegenCLI` tool to be rebuilt on every build, even when its inputs do not change. See this ticket, which describes the problem and notes that other Swift packages have experienced the same: [swiftlang/swift-build#305](https://github.com/swiftlang/swift-build/issues/305)
> > I have confirmed that frequent rebuilds do not occur on the non-Xcode Swift toolchain, bolstering the belief that this particular swift-build issue is the cause. (I can't propose a fix for it either, because the cause is apparently in closed-source Xcode source code.)
> > So these are the actions I plan to take this week:
> > 
> > * I will merge this proposed PR after a couple mods, because it does correct the issue of deterministic code generation.
> > * I will modify the build plugin to mitigate the effect of the swift-build bug cited above.  I will probably do something similar to [Support caching generated outputs to workaround incremental build bugs liamnichols/xcstrings-tool#129](https://github.com/liamnichols/xcstrings-tool/pull/129) which addresses this issue but in a different package that utilizes a similar plugin.
> > 
> > (If you're an amplify-swift customer, I'll work with that team to get the fix published in an upcoming amplify-swift release once it's fixed here.)
> 
> Interesting, my next change was going to be this [PR](https://github.com/RufusMall/smithy-swift/pull/2/changes) which further speeds things up due to the timestamp of the generated files no longer being updated. Is this similar to the additional change you were going to apply?
> 
> In my testing this PR alone was still a big improvement, because even if the file timestamp changes, Xcode eventually bypasses recompiling etc because the file hashes match (or something). After this second change (linked below) the file timestamps are also no longer updated so it speeds up the Xcode compilation even more.
> 
> https://github.com/RufusMall/smithy-swift/pull/2/changes
> 
> Nice, I was surprised that Xcode was running the plugin multiple times... interesting that the non-Xcode Swift toolchain doesn't have this bug...
> 
> Thanks for looking into this!

@jbelkins I can merge this additional work on the timestamp fix into this PR if you want, as they are kind of related.

jbelkins:
@RufusMall I intend to merge this once CI passes, thank you for the contribution.  You don't need to contribute a timestamp fix, I'll prepare & merge that next.

Not rewriting files to preserve timestamps should help to mitigate the long build times until we can get an Xcode fix for https://github.com/swiftlang/swift-build/issues/305.  I'll advocate for that swift-build issue to be fixed, though it will likely take a while.

@RufusMall and @trickart Thank you both for your research into this issue, I'll update you once there is a fixed version of amplify-swift available.

jbelkins (approved):

jbelkins:
@RufusMall please open this as a PR against origin/main and I will merge it as well:
https://github.com/RufusMall/smithy-swift/pull/2

jbelkins:
@RufusMall @trickart amplify-swift has made the following release, which includes our fixes for incremental builds:
https://github.com/aws-amplify/amplify-swift/releases/tag/2.60.1

Please update to that version and retest to see if incremental builds are any more snappy.

RufusMall:
> @RufusMall @trickart amplify-swift has made the following release, which includes our fixes for incremental builds: https://github.com/aws-amplify/amplify-swift/releases/tag/2.60.1
> 
> Please update to that version and retest to see if incremental builds are any more snappy.

@jbelkins I did a quick test of incremental builds on one of our main apps after the upgrade.

Macbook Pro with M5 Pro
Before:
- No code changes: 44 seconds
- App delegate changed: 40 seconds
- WebView changed: 50 seconds

After:
- No code changes: 9 seconds
- App delegate changed: 19 seconds
- WebView changed: 22 seconds

This is going to save us a huge amount of time :P. Thanks for your help!

I would be curious to know what improvements @trickart sees!

trickart:
We've also confirmed the improvement on our side. Thank you!

No code changes

- Before: 90 seconds
- After: 35 seconds

PR #1157, opened 2026-09-03
perf: URLEncodingUtils pure Swift implementation

jbelkins:
## Description of changes
Increases performance on URL encoding by replacing Foundation methods with a pure-Swift implementation of percent-escaping.
- The currently used methods bridge to `NSString` which is an Objective-C class.  Calling Objective-C code from Swift incurs a performance penalty, both in the bridging and in that `NSString` is less performant than Swift `String`.
- A UTF-8 byte-level method is used to escape the characters in the string, directly in Swift.  Escaping is same as the previous implementation, and is RFC-compliant.  https://www.rfc-editor.org/info/rfc3986/
- Tests are added for escaping many possible combinations of characters.  There were no tests (at least not unit tests) on `URLEncodingUtils` previously.

## Scope


averyk22 (approved):
LGTM!

PR #1158, opened 2026-09-03
perf: HTTP `Headers` performance improvements & tests

jbelkins:
## Description of changes
Two performance improvements in the `Headers` type, plus tests.
- Removes internal locking from the `Headers` type.  Locking causes additional overhead when accessing headers, damaging performance.  The locking on `Headers` is unnecessary because:
  - `Headers` is a Swift value type, with value semantics, so it is copied across concurrency zones, not carried as a reference.
  - There is one place where `Headers` may be modified from multiple threads because it is a property of a reference-type `HTTPResponse` (in the CRT HTTP client), but that access is already made exclusive by a lock on `HTTPResponse`, rendering the lock on `Headers` superfluous.
- The case-insensitive compare code has been improved by comparing strings in-place, avoiding the extra allocations from calling `.downcased()` on each string before comparison.
- Tests have been added for existing behavior of `Headers`.

This change increases performance measurably on any protocol / request where many headers are used.  As much as 3x performance increase has been seen on REST-type requests with numerous headers.

## Scope

### DIFF
--- a/Sources/SmithyHTTPAPI/Headers.swift
+++ b/Sources/SmithyHTTPAPI/Headers.swift
@@ -5,28 +5,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-import class Foundation.NSRecursiveLock
+public struct Headers: Sendable {
 
-public struct Headers: @unchecked Sendable {
-    private let lock = NSRecursiveLock()
-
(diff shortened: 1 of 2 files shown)


averyk22 (approved):
Lgtm!

PR #1166, opened 2026-09-18
feat: Raise Swift, Xcode, and platform minimums

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/discussions/2140

## Description of changes

_Companion to aws-sdk-swift PR https://github.com/awslabs/aws-sdk-swift/pull/2222_

Makes the following changes to Swift, Xcode and platform support:
* Minimum supported Swift version will be changed from Swift 5.9 to **Swift 6.0** on all Apple platforms and on Linux.
* Minimum supported Xcode version will be **Xcode 16.0**.
* Minimum supported Apple platforms will change as follows:
    * Minimum supported macOS will change from macOS 12 to **macOS 11**.
    * Minimum supported iOS will change from iOS 13 to **iOS 15**.
    * Minimum supported tvOS will change from tvOS 13 to **tvOS 15**.
    * Minimum supported watchOS will change from watchOS 6 to **watchOS 8**.
    * Minimum supported visionOS will remain **visionOS 1.0**.

## Scope


smithy-swift-codegen/src/test/kotlin/software/amazon/smithy/swift/codegen/manifestanddocs/PackageManifestGeneratorTests.kt
```diff
@@ -32,7 +32,10 @@ class PackageManifestGeneratorTests {
         assertNotNull(packageManifest)
         val expected = """
     platforms: [
-        .macOS(.v12), .iOS(.v13), .tvOS(.v13), .watchOS(.v6)
+        .macOS(.v11),
```
Madrigal: is this a downgrade of macOS platform?

Madrigal: I see this is explained on the PR description

> Minimum supported macOS will change from macOS 12 to macOS 11.

it'd be great if you can elaborate on why this is changing to a lower version

jbelkins: Yes, and that is intended.  We previously took a dependency on OpenTelemetry that required macOS 12 minimum.  We split that dependency out to a [separate module](https://github.com/smithy-lang/smithy-swift-opentelemetry), but we didn't restore the macOS floor in `smithy-swift` to 11, until now.

Madrigal (approved):

